PR #4, opened 2025-10-31
Validator ci setup

rmeissner:
- Add github workflow to run biome checks and tests

### DIFF
--- a/.github/workflows/validator.yml
+++ b/.github/workflows/validator.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      - uses: actions/setup-node@v4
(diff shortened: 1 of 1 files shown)


validator/package.json
```diff
@@ -4,7 +4,8 @@
 	"scripts": {
 		"build": "rimraf dist && tsc",
 		"dev": "tsx src/validator.ts",
-		"check": "biome check"
+		"check": "biome check",
+		"test": "echo \"Error: no test specified\" && exit 0"
```
rmeissner: I use `exit 0` so that the test is not failing the workflow runs right now (which would block it from making runs required).

nlordell (approved):

PR #13, opened 2025-11-07
Implement Keygen flow in validator

rmeissner:
- Add FrostClient which handle the keygen logic
- Add FROST math functions
  - Note: these need to be refactored for better code structure later on
- Add event watchers to connect events to frost client
  - Note: this was done to completely separate the FROST logic from contract logic
- Add FrostCoordinator (and OnchainCoordinator) that is used by the FrostClient to submit data
- Add schemas for event parsing 
  - Note: by default all event params by viem seem to be optional, therefore using zod allows typesafety
- Add end to end tests:
  - "unit test" which does not require the blockchain to run
  - "integration test" which connects to the local anvil blockchain

Note: 
There are quite some TODOs which will be handled in following PRs

### DIFF
--- a/validator/src/consensus/client.test.ts
+++ b/validator/src/consensus/client.test.ts
@@ -155,7 +155,7 @@ describe("client", () => {
 
 	it.skip("integration keygen flow", { timeout: 30000 }, async () => {
 		// Make sure to first start the Anvil testnode (run `anvil` in the root)
-		// and run the deployment script: forge script contracts/script/Deploy.s.sol:DeployScript --rpc-url http://127.0.0.1:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --broadcast
+		// and run the deployment script: forge script DeployScript --rpc-url http://127.0.0.1:8545 --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --broadcast
 		// Private key from Anvil testnet
 		const accounts = [
 			privateKeyToAccount(
--- a/validator/src/types/schemas.ts
+++ b/validator/src/types/schemas.ts
@@ -24,19 +24,19 @@ export const validatorConfigSchema = z.object({
 });
 
 export const frostPointSchema = z.object({
-	x: z.bigint().positive(),
-	y: z.bigint().positive(),
+	x: z.bigint().nonnegative(),
+	y: z.bigint().nonnegative(),
 });
 
 export const frostCommitmentSchema = z.object({
 	c: z.array(frostPointSchema),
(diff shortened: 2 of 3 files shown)


validator/src/types/schemas.ts
```diff
@@ -13,7 +13,47 @@ export const checkedAddressSchema = z
...
+	.transform((val) => val as Hex);
+
 export const validatorConfigSchema = z.object({
 	RPC_URL: z.url(),
 	CONSENSUS_CORE_ADDRESS: checkedAddressSchema,
 });
+
+export const frostPointSchema = z.object({
+	x: z.bigint().positive(),
+	y: z.bigint().positive(),
```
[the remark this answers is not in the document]
rmeissner: The contract type of the event is uint256, which by definition is positive

validator/src/consensus/client.ts
```diff
@@ -1,126 +1,237 @@
...
+): [bigint, number] => {
+	const arrayIndex = participants.findIndex(
+		(p) => p.address === validatorAddress,
+	);
+	if (arrayIndex < 0)
+		throw Error("Cannot find participant in participants array!");
+	const participantIndex = participants.at(arrayIndex)?.index;
+	if (participantIndex === undefined)
+		throw Error("Cannot determine FROST index of participant!");
+	return [participantIndex, arrayIndex];
```
nlordell: Not a comment on your code specifically, but I find "index" to be an overloaded term which causes confusion (does it start at 0? Is it a participant index that starts at 1? etc.)

I would propose that we change this to `Identifier` (this is what the [reference FROST implementation calls it](https://docs.rs/frost-core/latest/frost_core/struct.Identifier.html), which would give us clearer names for participants "start from 1" indexes/identifiers, from array "start from 0" indexes. I would propose renaming this in the contract as well.

Obviously, not for this PR, just something to help with code clarity going forward.

validator/src/consensus/client.test.ts
```diff
@@ -93,41 +114,138 @@ describe("client", () => {
...
+				"0x5de4111afa1a4b94908f83103eb1f1706367c2e68ca870fc3fb9a804cdab365a",
+			),
+			privateKeyToAccount(
+				"0x7c852118294e51e653712a81e05800f419141751be58f605c371e15141b007a6",
+			),
+		];
+		const participants: Participant[] = accounts.map((a, i) => {
+			return { index: BigInt(i + 1), address: a.address };
+		});
+		const coordinatorAddress = "0x601CDd9C9e743AEb40106865775FDB073ee4E889";
```
nlordell: nit: This is pretty fragile (any changes to contracts will require manual intervention for the E2E test. It would be great if we could derive this somehow (off the top of my head - you could load the byte code from the artifact JSON and compute the expected address).

PR #16, opened 2025-11-11
FROST Signing Ceremony

nlordell:
This PR implements the onchain FROST signing ceremony coordination logic. This allows for an established group to sign a message with a minimum threshold of participants.

This follows very closely the RFC-9591 implementation, with a few small, but important, differences:
- Committing to nonces is done before a signing ceremnory, by committing to a Merkle root of a chunk of 1024 nonces. This ensures that nonces are "set in stone" before the message to sign is revealed with is important for preventing Wagner's Birthday Attack (whereby an attacker mines a problematic nonce that reveals the another participants secret share).
- The coordinator implicitely assumes that the participants have some other mechanism for computing the set `S` of signing participants. In the context of Shieldnet, we are implicitly assuming that "all honest participants" are part of the signing ceremony. This allows each validator to independently compute the group commitment from the set of revealed nonces.
- Submitting a share onchain is done with a "commitment root". This is (another... :sob:) Merkle tree of all of the participant's commitment shares and `c * λ` values (which each participant can compute themselves with public information from events that are emitted when the signing nonces are revealed). This ensures that participants can only update the group signature values when they are using commitment shares that match what other participants computed. This allows us to both verify participant shares and compute the group signature onchain, without allowing malicious participants to influence the group signature from an honest majority.

An E2E test was added that goes over the whole signing ceremony and successfully produces a group FROST signature that can be verified onchain :smile:.

This superseded #15

### DIFF
--- a/contracts/src/FROSTCoordinator.sol
+++ b/contracts/src/FROSTCoordinator.sol
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-3.0-only
 pragma solidity ^0.8.30;
 
-import {FROSTCommitmentSet} from "@/lib/FROSTCommitmentSet.sol";
+import {FROSTNonceCommitmentSet} from "@/lib/FROSTNonceCommitmentSet.sol";
 import {FROSTParticipantMap} from "@/lib/FROSTParticipantMap.sol";
 import {FROSTSignatureShares} from "@/lib/FROSTSignatureShares.sol";
 import {Secp256k1} from "@/lib/Secp256k1.sol";
@@ -9,7 +9,7 @@ import {Secp256k1} from "@/lib/Secp256k1.sol";
 /// @title FROST Coordinator
 /// @notice An onchain coordinator for FROST key generation and signing.
 contract FROSTCoordinator {
-    using FROSTCommitmentSet for FROSTCommitmentSet.T;
+    using FROSTNonceCommitmentSet for FROSTNonceCommitmentSet.T;
     using FROSTParticipantMap for FROSTParticipantMap.T;
     using FROSTSignatureShares for FROSTSignatureShares.T;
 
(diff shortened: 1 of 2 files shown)


contracts/src/lib/FROST.sol
```diff
@@ -0,0 +1,222 @@
...
+        uint256 index;
+        Secp256k1.Point d;
+        Secp256k1.Point e;
+    }
+
+    error UnorderedCommitments();
+    error InvalidScalar();
+
+    /// @notice Generate a random nonce from some randomness and a secret key.
+    function nonce(bytes32 random, Secp256k1.Point memory secret) internal view returns (uint256 n) {
```
rmeissner: Are you sure that this method is correctly implemented? 

As far as I understand the `nonce_generate(secret)` function definition, `secret` is not the public key (so a point), but rather the secret key (so a uint256). 

As far as I understand you signing test (https://github.com/safe-research/shieldnet/pull/16/files#diff-99c3b88eb27150979236c3a222c847c6dd5f6e444dd739b42183db3ecfb52362R236) you use g(s_1), which would be the public key, right?

rmeissner: https://datatracker.ietf.org/doc/html/rfc9591#section-6.5 links to https://secg.org/sec1-v2.pdf. On page 12 we see that this points for secp256k1 (as it p is odd, so case 1) to use  Integer-to-Octet-String Conversion, which is

contracts/src/lib/FROST.sol
```diff
@@ -0,0 +1,222 @@
...
+            }
+        }
+    }
+
+    function _hash(bytes memory input, bytes32 dst) private view returns (bytes32 digest) {
+        assembly ("memory-safe") {
+            let dstLen := byte(31, dst)
+            let dstOffset := sub(32, dstLen)
+            let len := mload(input)
+            mstore(input, shr(shl(4, dstOffset), dst))
```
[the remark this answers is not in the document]
nlordell: Yup, definitely a typo. I also noticed this when adding tests in https://github.com/safe-research/shieldnet/pull/17 and fixed it there.

PR #24, opened 2025-11-14
Validator Signing Flow

rmeissner:
## Add signing support

Besides signing support this PR also includes some refactorings. Namely the introduction of a storage layer. As shown in the image below, this is done to clearly separate concerns and allow us to refactor different parts of the validator in parallel.

image

### Changes:
- Extract storage into separate class
  - Use to provide a common interface for storage management
  - Currently only a in-memory storage is available
- Add SigningCoordinator
  - Used to post signature related information to the coordinator
- Add SigningClient
  - Used to handle nonce commitments and signature share creation
- Added tests
  - Basic tests for some primitives (i.e. hashing, nonce creation and encoding)
  - E2E test for signing, based with fixed input generated by Solidity tests 
  
### Notes:
- This PR does not yet link the validator logic to the smart contracts, this will happen in a follow up PR

### DIFF
--- a/validator/src/consensus/signing/client.ts
+++ b/validator/src/consensus/signing/client.ts
@@ -16,7 +16,7 @@ import {
 	nonceCommitmentsWithProof,
 	type PublicNonceCommitments,
 } from "./nonces.js";
-import { createSignatureShare, lagrangeChallange } from "./shares.js";
+import { createSignatureShare, lagrangeChallenge } from "./shares.js";
 import { verifySignatureShare } from "./verify.js";
 
 export class SigningClient {
@@ -145,7 +145,7 @@ export class SigningClient {
 				throw Error(`Missing nonce commitments for ${signerId}`);
 			const r = groupCommitmentShares[index];
 			const coeff = lagrangeCoefficient(signers, signerId);
-			const cl = lagrangeChallange(coeff, challenge);
+			const cl = lagrangeChallenge(coeff, challenge);
 			const node = keccak256(
 				encodePacked(
 					["uint256", "uint256", "uint256", "uint256"],
--- a/validator/src/consensus/signing/client.test.ts
+++ b/validator/src/consensus/signing/client.test.ts
@@ -172,7 +172,7 @@ describe("signing", () => {
 					_signingParticipantsHash: Hex,
 					groupCommitementShare: FrostPoint,
 					signatureShare: bigint,
-					_lagrangeChallange: bigint,
+					_lagrangeChallenge: bigint,
 					_signingParticipantsProof: Hex[],
 				): Promise<Hex> => {
 					signatureShareEvents.push({
--- a/validator/src/consensus/signing/shares.ts
+++ b/validator/src/consensus/signing/shares.ts
@@ -1,22 +1,22 @@
 import { addmod, mulmod } from "../../frost/math.js";
 import type { SecretNonceCommitments } from "./nonces.js";
 
-export const lagrangeChallange = (
-	langrangeCoefficient: bigint,
+export const lagrangeChallenge = (
+	lagrangeCoefficient: bigint,
 	challenge: bigint,
-): bigint => mulmod(challenge, langrangeCoefficient);
+): bigint => mulmod(challenge, lagrangeCoefficient);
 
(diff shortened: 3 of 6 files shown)


validator/src/consensus/keyGen/client.ts
```diff
@@ -0,0 +1,221 @@
...
+		);
+	}
+
+	async handleKeygenCommitment(
+		groupId: GroupId,
+		senderId: ParticipantId,
+		peerCommitments: readonly FrostPoint[],
+		pok: ProofOfKnowledge,
+	) {
+		const participantIndex = this.#storage.participantId(groupId);
```
nlordell: consistency nit: 
```suggestion
		const participantId = this.#storage.participantId(groupId);
```

validator/src/consensus/keyGen/client.ts
```diff
@@ -0,0 +1,221 @@
...
+			const secretShares = this.#storage.secretSharesMap(groupId);
+			const signingShare = createSigningShare(secretShares);
+			verifyKey(verificationShare, signingShare);
+			this.#storage.registerSigningShare(groupId, signingShare);
+			this.#storage.clearKeyGen(groupId);
+			const participantId = this.#storage.participantId(groupId);
+			console.info(`Final signing key for ${participantId} calculated`);
+		}
+	}
+}
```
nlordell: General comment, the code is clear and super easy to follow given my FROST knowledge 👏.

validator/src/consensus/signing/client.ts
```diff
@@ -0,0 +1,197 @@
...
+		this.#storage.linkNonceTree(groupId, chunk, nonceCommitmentsHash);
+	}
+
+	async handleSignatureRequest(
+		groupId: GroupId,
+		signatureId: SignatureId,
+		message: Hex,
+		sequence: bigint,
+	) {
+		// TODO: check if we really want to sign the message
```
nlordell: What, no blind signing? I think we should invent a protocol so you can blind sign this `message` with confidence. Maybe called "Schildnetz"?

validator/src/frost/hashes.ts
```diff
@@ -0,0 +1,55 @@
+import {
+	type H2COpts,
+	hash_to_field,
```
nlordell: So happy there is a library for this - it is annoying to implement.

nlordell (approved):
Looks great - only tiny comments. Code was very clear and easy to follow/review.

validator/src/consensus/signing/client.ts
```diff
@@ -0,0 +1,197 @@
...
+
+		const participants = this.#storage.participants(groupId);
+		const signers = participants.map((p) => p.id).sort();
```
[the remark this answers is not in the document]
rmeissner: I actually took this from your code, but yes it should be fair to assume that the same sorting as for the participants is applied.

validator/src/consensus/signing/client.ts
```diff
@@ -0,0 +1,197 @@
...
+		await this.#coordinator.publishSignatureShare(
+			signatureId,
+			signingParticipantsHash,
+			signerPart.r,
+			signatureShare,
+			signerPart.cl,
+			signingParticipantsProof,
+		);
+	}
+}
```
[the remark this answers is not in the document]
rmeissner: yeah, I would assume that out of order messages might cause quite some hick ups ... lets see, something to test :P. We will have to do a "break this system" session :D

validator/src/frost/hashes.ts
```diff
@@ -0,0 +1,55 @@
...
+} from "@noble/curves/abstract/hash-to-curve.js";
+import { secp256k1 } from "@noble/curves/secp256k1.js";
+import { concatBytes, numberToBytesBE } from "@noble/curves/utils.js";
+import { sha256 } from "@noble/hashes/sha2.js";
+import { stringToBytes } from "viem";
+import type { FrostPoint } from "./types.js";
+
+const N = secp256k1.Point.CURVE().n;
+const DOMAIN = "FROST-secp256k1-SHA256-v1";
```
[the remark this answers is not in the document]
rmeissner: Will be updated in #25

PR #27, opened 2025-11-17
Add workflow for integration tests

rmeissner:
- Add integration checks script
  - Pull coordinator address from Deploy script output
  - Skip test if anvil is not running or no deployment information can be found
- Add workflow to run integration checks script

### DIFF
--- a/scripts/run_integration_test.sh
+++ b/scripts/run_integration_test.sh
@@ -20,7 +20,8 @@ sleep 2
 echo "Deploying contracts..."
 forge script DeployScript \
     --rpc-url $ANVIL_RPC_URL \
-    --private-key 0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d \
+    --unlocked \
+    --sender 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 \
     --broadcast
 
 # --- 4. Run Client Integration Tests ---
--- a/validator/src/consensus/integration.test.ts
+++ b/validator/src/consensus/integration.test.ts
@@ -1,9 +1,9 @@
 import { randomBytes } from "node:crypto";
 import fs from "node:fs";
 import path from "node:path";
-import { bytesToNumberBE } from "@noble/curves/utils.js";
 import {
-	Address,
+	type Address,
+	bytesToHex,
 	createPublicClient,
 	createWalletClient,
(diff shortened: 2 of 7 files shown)


scripts/run_integration_test.sh
```diff
@@ -0,0 +1,34 @@
...
+# --- 2. Setup Cleanup ---
+trap 'echo "Stopping Anvil (PID $ANVIL_PID)..." && kill $ANVIL_PID' EXIT
+sleep 2
+
+# --- 3. Deploy Contracts ---
+# Use anvil default account
+echo "Deploying contracts..."
+forge script DeployScript \
+    --rpc-url $ANVIL_RPC_URL \
+    --private-key 0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d \
```
nlordell: I wonder if you can just use "account 0", so you don't need the private key.

scripts/run_integration_test.sh
```diff
@@ -0,0 +1,34 @@
...
+set -e
+
+# --- Configuration ---
+CLIENT_PACKAGE_NAME="client" # <-- Replace with your client package name
+CONTRACTS_OUTPUT_FILE=".deploy-output.json"
+ANVIL_RPC_URL="http://127.0.0.1:8545"
+
+# --- 1. Start Anvil in the background ---
+echo "Starting Anvil..."
+anvil > /dev/null &
```
nlordell: nit: might be nice to log to file, so we can `cat` it in the `trap`.

validator/src/consensus/integration.test.ts
```diff
@@ -29,10 +32,40 @@ import type { Participant } from "./types.js";
...
+		try {
+			await readClient.getBlockNumber();
+		} catch {
+			// Anvil not running
+			skip();
+		}
+		const deploymentInfo = JSON.parse(
+			fs.readFileSync(deploymentInfoFile, "utf-8"),
+		);
+		const coordinatorAddress = deploymentInfo.returns["0"].value as Hex;
```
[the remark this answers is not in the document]
rmeissner: yes :see_no_evil:

scripts/run_integration_test.sh
rmeissner: I will log them into `anvil_logs.txt` (also added that to gitignore)

PR #36, opened 2025-11-20
Transaction Attestations

nlordell:
This PR adds support for transaction attestations to the consensus contract. Additionally, it provides convienient methods to get the attestations by meta-transaction and message (to pass the "Etherscan" test of being able to easily get signatures for a transaction).

It might be a good idea to add a similar convenience method for getting signatures for epoch rollovers, and potentially even "building" a complete signature (that combines a meta-transaction attestation with all the "catchup" signatures to form an epoch chain, in other words the exact signature bytes that should be passed to a potential guard).

### DIFF
--- a/contracts/src/Consensus.sol
+++ b/contracts/src/Consensus.sol
@@ -30,7 +30,9 @@ contract Consensus is IFROSTCoordinatorCallback {
         uint64 indexed activeEpoch, uint64 indexed proposedEpoch, uint64 rolloverBlock, Secp256k1.Point groupKey
     );
     event EpochRolledOver(uint64 indexed newActiveEpoch);
-    event TransactionProposed(bytes32 indexed message, uint64 epoch, MetaTransaction.T transaction);
+    event TransactionProposed(
+        bytes32 indexed message, bytes32 indexed transactionHash, uint64 epoch, MetaTransaction.T transaction
+    );
     event TransactionAttested(bytes32 indexed message);
 
     error InvalidRollover();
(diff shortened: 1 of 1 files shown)


contracts/src/Consensus.sol
```diff
@@ -3,163 +3,183 @@ pragma solidity ^0.8.30;
...
+        emit EpochStaged(epochs.active, proposedEpoch, rolloverBlock, groupKey);
     }
 
-    function onKeyGenCompleted(FROSTGroupId.T group, bytes calldata context) external {
-        (uint64 proposedEpoch, uint64 rolloverAt) = abi.decode(context, (uint64, uint64));
-        proposeEpoch(proposedEpoch, rolloverAt, group);
+    function proposeTransaction(MetaTransaction.T memory transaction) external returns (bytes32 message) {
+        Epochs memory epochs = _processRollover();
+        message = domainSeparator().transactionProposal(epochs.active, transaction.hash());
+        emit TransactionProposed(message, epochs.active, transaction);
```
rmeissner: Why not just emit the safeTxHash instead of the whole tx. If we index that then it should be easy to find the info for a specific safeTx.

rmeissner (approved):

contracts/src/Consensus.sol
nlordell: `message` is indexed and is kind of the "safe transaction hash" for proposed transactions.

Do you mean additionally emit and index the `transaction.hash()` value?

nlordell: Added - as a side note, we should probably go through all the events at the end and decide "what fields we need and which ones need to be indexed" once we are up and running. I'm 100% OK with erring on the side of caution for now, but I suspect that we are logging some superfluous data that we can remove. Same goes for storage.

PR #39, opened 2025-11-21
Implement State Machine

rmeissner:
- Implemented a state machine to handle the interaction with the protocol
- Renamed "Coordinator" to "Protocol" to represent interactions with the protocol
- Adjusted the integration test to use a state machine

### DIFF
--- a/validator/src/service/machine.ts
+++ b/validator/src/service/machine.ts
@@ -204,7 +204,7 @@ export class ShieldnetStateMachine {
 			(block === this.#lastProcessedBlock && index <= this.#lastProcessedIndex)
 		) {
 			throw Error(
-				`Invalid block number (${block}) and index ${index} (currentyl at block ${this.#lastProcessedBlock} and index ${this.#lastProcessedIndex})`,
+				`Invalid block number (${block}) and index ${index} (currently at block ${this.#lastProcessedBlock} and index ${this.#lastProcessedIndex})`,
 			);
 		}
 		this.#lastProcessedIndex = index;
--- a/validator/src/consensus/protocol.ts
+++ b/validator/src/consensus/protocol.ts
@@ -38,7 +38,7 @@ export class OnchainProtocol implements ShieldnetProtocol {
 		return this.#consensus;
 	}
 	coordinator(): Address {
-		return this.#consensus;
+		return this.#coordinator;
 	}
 	async triggerKeygenAndCommit(
 		participants: Hex,


contracts/script/Deploy.s.sol
```diff
@@ -2,17 +2,22 @@
...
 
 contract DeployScript is Script {
-    function run() public returns (FROSTCoordinator) {
+    function run() public returns (FROSTCoordinator, Consensus) {
         vm.startBroadcast();
 
         FROSTCoordinator coordinator = new FROSTCoordinator{salt: bytes32(0)}();
+        console.log("FROSTCoordinator deployed to:", address(coordinator));
+
+        Consensus consensus = new Consensus{salt: bytes32(0)}(address(coordinator), FROSTGroupId.T.wrap(0xfa3efbc5dd215abad66149ad58d7869f90cf6d2c1bc472be0000000000000000));
```
nlordell: I imagine that this should, at some point, be read from `vm.getEnv`.

rmeissner: yes, totally

validator/src/consensus/signing/client.test.ts
```diff
@@ -124,18 +130,18 @@ describe("signing", () => {
 	it("e2e signing flow", async () => {
 		const nonceCommitmentsEvents: {
 			groupId: GroupId;
-			index: bigint;
+			signerId: bigint;
```
nlordell: Just so I am sure, `signerId` is a `FROST.Identifier` correct?

rmeissner: yes, and it should actually be of type ParticipantId here (because that is my type for FROST.Identifier)

validator/src/service/machine.ts
```diff
@@ -0,0 +1,499 @@
...
+		if (
+			block < this.#lastProcessedBlock ||
+			(block === this.#lastProcessedBlock && index <= this.#lastProcessedIndex)
+		) {
+			throw Error(
+				`Invalid block number (${block}) and index ${index} (currentyl at block ${this.#lastProcessedBlock} and index ${this.#lastProcessedIndex})`,
+			);
+		}
+		this.#lastProcessedIndex = index;
+		await this.progressToBlock(block);
```
nlordell: Theoretically, the block is processed _before_ the event index. I don't think it matters for this class though.

validator/src/service/machine.ts
```diff
@@ -0,0 +1,499 @@
...
+						groupId: event.gid,
+						nextEpoch,
+					};
+				}
+				return;
+			}
+			case "KeyGenSecretShared": {
+				// A participant has submitted secret share for new group
+				// Ignore if not in "collecting_shares" state
+				if (this.#keyGenState.id !== "collecting_shares") return;
```
nlordell: This should be an error if `this.#keyGenState.groupId !== event.gid` IMO.

validator/src/service/machine.ts
```diff
@@ -0,0 +1,499 @@
...
+	}
+
+	async processBlockEvent(
+		block: bigint,
+		index: number,
+		eventName: string,
+		eventArgs: unknown,
+	) {
+		if (
+			block < this.#lastProcessedBlock ||
```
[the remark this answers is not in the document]
rmeissner: I have it in both places right now, and here is the additional check for event ordering.

validator/src/service/machine.ts
rmeissner: currently we listen to all events, and the coordinator contract can be used by others to do keygen too, so I would just ignore it right now

PR #41, opened 2025-11-25
Initial service implementation

rmeissner:
- Setup state machine in service
  - Use same setup as for integration tests for now
- Adjust validator config scheme to include all require params
  - Account key
  - Consensus and Coordinator address
  - Chain id
  - Participants
- Sort events when they are returned by viem


validator/src/validator.ts
```diff
@@ -1,13 +1,19 @@
...
 import { z } from "zod";
+import type { Participant } from "./consensus/storage/types.js";
 import { createValidatorService } from "./service/service.js";
-import type { ConsensusConfig } from "./types/interfaces.js";
+import type { ProtocolConfig } from "./types/interfaces.js";
 import { validatorConfigSchema } from "./types/schemas.js";
 
 dotenv.config();
 
+const BLOCKTIME_IN_SECONDS = 5n; // value assumed for gnosis chain
```
nlordell: I would create a `chainConfig.ts` that contains the chain ID and the block information - otherwise this might get accidentally out of sync once we use different networks.

Not for this PR, but a future improvement.

PR #42, opened 2025-11-25
Validator Specification

nlordell:
This PR is an attempt to formalize the validator state machine in an unabiguous way, so that we can agree on exactly how the validator should function.

This is PR is in draft, as I don't know whether or not we want to commit it in its current form (especially since, it may become outdated very quickly since we don't run any tests against it).

It is still useful to share - as we can use it as a starting point for discussin _how_ we expect validators to work.

### DIFF
--- a/spec/validator.ml
+++ b/spec/validator.ml
@@ -1,0 +1,868 @@
+module IntMap = Map.Make (Int)
+module StringMap = Map.Make (String)
+
+module Address = struct
+  type t = unit (* place holder *)
+
+  let equal (_ : t) (_ : t) : bool = failwith "place holder"
+end
+
+module Secp256k1 = struct
+  type secret_key = unit (* place holder *)
+  type public_key = unit (* place holder *)
+  type scalar = unit (* place holder *)
+  type point = unit (* place holder *)
+
+  let infinity : point = () (* place holder *)
+  let random_scalar () : scalar = failwith "place holder"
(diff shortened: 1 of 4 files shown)


spec/validator.ml
```diff
@@ -0,0 +1,868 @@
...
+  let validate_transaction _ =
+    (* TBD. *)
+    true
+
+  let transaction_proposed state message transaction_hash epoch transaction =
+    if epoch == state.active_epoch.id && validate_transaction transaction then
+      let signing_ceremonies' =
+        StringMap.add message
+          {
+            status = Waiting_for_sign_request { responsible = account_address };
```
rmeissner: Make responsible None

spec/validator.ml
```diff
@@ -0,0 +1,868 @@
...
+      group_commitment : Secp256k1.point;
+      shares : Secp256k1.scalar IntMap.t;
+    }
+  | Waiting_for_consensus_attestation of {
+      id : signature_id;
+      responsible : Address.t;
+    }
+
+type packet =
+  | Epoch_rollover of { epoch : int; rollover : int; group_id : group_id }
```
rmeissner: this is not how the packet is defined onchain, as there is no "offchain" way to link from a groupKey to a groupId

nlordell: Good point

rmeissner (approved):

PR #43, opened 2025-11-26
[Chore] Biome and TS Configuration Fixes

nlordell:
Just a small chore PR with some unrelated changes to the storage work on the build system:

- Fix the Biome configuration to correctly extend the workspace root; previously it was ignoring the workspace root configuration.
- Removed defaults from Biome configuration (all fields that were removed were the default, and setting them had no effect).
- Create a separate TSConfig file for outputing the build to `dist/`; this allows to _only_ output files that are needed to run the validator and not include tests in the output as well.
- Remove unused TSConfiguration fields (we don't use JSX, nor do we use decorators, and excluding `node_modules` does nothing, since we don't include it in the first place).

As a follow up, I will propose bumping up the line length to 120; but I don't want to do it here and causes a bunch of merge conflicts with the state machine changes @rmeissner is working on.


validator/package.json
```diff
@@ -2,9 +2,9 @@
 	"name": "validator",
 	"type": "module",
 	"scripts": {
-		"build": "rimraf dist && tsc",
+		"build": "rimraf dist && tsc --project tsconfig.dist.json",
 		"dev": "tsx src/validator.ts",
-		"check": "biome check",
+		"check": "biome check && tsc",
```
rmeissner: why do we build on check?

rmeissner (approved):

PR #46, opened 2025-11-28
Validator Sqlite KeyGen Storage

nlordell:
This PR adds the KeyGen storage methods to the Sqlite storage engine.

Note that we also changed the valious data bits to use `BLOB` columns instead of `TEXT` as hex for more efficient storage.

### DIFF
--- a/validator/src/consensus/storage/sqlite.ts
+++ b/validator/src/consensus/storage/sqlite.ts
@@ -1,14 +1,22 @@
 import Sqlite3, { type Database } from "better-sqlite3";
-import type { Address, Hex } from "viem";
+import { type Address, concat, type Hex } from "viem";
 import { z } from "zod";
-import { pointFromHex, scalarFromHex, scalarToHex } from "../../frost/math.js";
+import {
+	pointFromBytes,
+	scalarFromBytes,
+	scalarToBytes,
+} from "../../frost/math.js";
 import type {
 	FrostPoint,
 	GroupId,
 	ParticipantId,
 	SignatureId,
 } from "../../frost/types.js";
-import { checkedAddressSchema, hexDataSchema } from "../../types/schemas.js";
+import {
+	checkedAddressSchema,
+	chunked,
+	hexDataSchema,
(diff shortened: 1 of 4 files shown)


validator/src/consensus/storage/sqlite.ts
```diff
@@ -232,56 +311,154 @@ export class SqliteStorage
...
+	missingSecretShares(groupId: GroupId): ParticipantId[] {
+		return this.#db
+			.prepare(`
+				SELECT p.id FROM group_participants AS p
+				LEFT JOIN group_secret_shares AS s ON s.group_id = p.group_id AND s.address = ? AND s.from_participant = p.id
+				WHERE p.group_id = ? AND s.secret_share IS NULL
+			`)
+			.pluck(true)
+			.all(this.#account, groupId)
+			.map((row) => dbIntegerSchema.parse(row));
```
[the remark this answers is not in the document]
nlordell: I guess this is a question for you @rmeissner - do you put in your "own secret share" (I.e. the evaluation of the polynomial with `coefficients` for your own participant ID) in storage, or only store the secret shares of others?

validator/src/consensus/storage/sqlite.ts
```diff
@@ -47,17 +63,29 @@ export class SqliteStorage
...
 				PRIMARY KEY(group_id, id),
+				FOREIGN KEY(group_id) REFERENCES groups(id) ON DELETE CASCADE,
+				UNIQUE(group_id,address)
+			);
+
+			CREATE TABLE IF NOT EXISTS group_secret_shares(
+				group_id TEXT NOT NULL,
+				address TEXT NOT NULL,
+				from_participant INTEGER NOT NULL,
+				secret_share BLOB NOT NULL,
```
rmeissner: did you have a specific reason why you used blobs for secret_share (and some other bytes based data), but not for group_id and address (so some other bytes based data)?

rmeissner: Makes sense

PR #49, opened 2025-12-01
State Machine Refactoring

rmeissner:
- Move every state transition into a separate function
- Add local state classes that track pending changes to the state
- Start adding tests for state transitions

### DIFF
--- a/validator/src/machine/consensus/epochStaged.ts
+++ b/validator/src/machine/consensus/epochStaged.ts
@@ -13,7 +13,7 @@ export const handleEpochStaged = async (
 			`Not expecting epoch staging during ${machineStates.rollover.id}!`,
 		);
 	}
-	// Check that state for signature id is "collect_signing_shares"
+	// Check that signing state is waiting for attestation
 	const status = machineStates.signing[machineStates.rollover.message];
 	if (status?.id !== "waiting_for_attestation") return {};
 
--- a/validator/src/machine/consensus/transactionAttested.ts
+++ b/validator/src/machine/consensus/transactionAttested.ts
@@ -8,7 +8,7 @@ export const handleTransactionAttested = async (
 	// The transaction attestation was submitted
 	// Parse event from raw data
 	const event = transactionAttestedEventSchema.parse(eventArgs);
-	// Check that state for signature id is "collect_signing_shares"
+	// Check that signing state is waiting for attestation
 	const status = machineStates.signing[event.message];
 	if (status?.id !== "waiting_for_attestation") return {};
 
--- a/validator/src/machine/keygen/committed.ts
+++ b/validator/src/machine/keygen/committed.ts
@@ -35,35 +35,35 @@ export const handleKeyGenCommitted = async (
 			mu: event.commitment.mu,
 		},
 	);
-	// If all participants have committed update state to "collecting_shares"
-	if (event.committed) {
-		const { verificationShare, shares } = keyGenClient.createSecretShares(
-			event.gid,
-		);
-		const callbackContext =
-			consensusState.genesisGroupId === event.gid
-				? undefined
-				: encodePacked(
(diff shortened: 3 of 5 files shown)


validator/src/consensus/signing/client.ts
```diff
@@ -55,13 +55,22 @@ export class SigningClient {
...
+		signers: readonly ParticipantId[],
 	): {
 		nonceCommitments: PublicNonceCommitments;
 		nonceProof: Hex[];
 	} {
-		// TODO: pass in signers to allow to remove unhonest
-		const participants = this.#storage.participants(groupId);
-		const signers = participants.map((p) => p.id).sort();
+		if (signers.length < this.#storage.threshold(groupId)) {
+			throw new Error("Not enough signers to start signing process");
```
nlordell: So, general comment (and doesn't have to be addressed here), but I don't think this is an error (i.e. an unexpected state). I think we need to handle this more gracefully and drop the signature request as opposed to failing the state transition. Especially if we want to recover for network outages.

This is in contrast to, for example, the "invalid signer id provided" error which should only happen because of a logic error (and therefore throwing an error makes sense).

validator/src/machine/consensus/epochStaged.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export const handleEpochStaged = async (
+	machineStates: MachineStates,
+	eventArgs: unknown,
+): Promise<StateDiff> => {
+	// An epoch was staged
+	const event = epochStagedEventSchema.parse(eventArgs);
+	// Ignore if not in "request_rollover_data" state
+	if (machineStates.rollover.id !== "sign_rollover") {
+		throw new Error(
```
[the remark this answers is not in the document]
rmeissner: True, happy to change it to an "ignore" instead of an "Error"

validator/src/machine/signing/sign.ts
```diff
@@ -0,0 +1,137 @@
...
+	MachineStates,
+	StateDiff,
+} from "../types.js";
+
+const NONCE_THRESHOLD = 100n;
+
+export const handleSign = async (
+	machineConfig: MachineConfig,
+	verificationEngine: VerificationEngine,
+	signingClient: SigningClient,
```
nlordell: Again, not for this PR, but one thing that caught my attention is that the only thing that is "side-effecty" right now are the clients. This can make a state transition valid or invalid depending on the internal state of the client (which is the only thing holding back making state transitions 100% replayable). Just some food for thought (as the "side-effecty-ness" is reasonable in this context IMO, and a valid optimization).

nlordell (approved):
Overall, GREAT change to make the state transitions easier to reason about. My high level comments (don't have to be addressed here) are:
1. I think some of the error states shouldn't be errors if we want to more gracefully handle network outages and restarts
2. I think the "side-effecty-ness" of the clients don't fit super well with the "pure" state transition functions; but that is a much larger refactor (so please don't include in this PR) and the jury's still out on whether or not its a good idea
3. This is more of a nit: observability is super important - can we just setup logging in the project as a whole instead of passing a logger? (again not for this PR).

validator/src/consensus/signing/client.ts
nlordell: Yes, this should be handled in the state transition function and not the client.

validator/src/machine/signing/sign.ts
rmeissner: I do think it is fair to refactor this and would start this after we setup the internal testenet (Note: I will create a project for this)

rmeissner: https://www.notion.so/safefoundation/Shieldnet-Stateless-key-gen-and-signing-client-2bda8a34f3b8809491c0faaa1ddfe1fc?source=copy_link

PR #57, opened 2025-12-03
`FROSTFinalization` Contract changes

remedcu:
## TLDR

This PR introduces a complaint system within the `FROSTCoordinator` for cases when a malicious validator intentionally shares the wrong secret share for a particular validator. In this case, the honest validator won't be able to compute the private key share, so they will need to raise this in the coordinator contract.

We also updated to use an explicit status for the group to know the current stage of the keyGen process.

Validator changes were also introduced: instead of the 2-round system, it now uses a 3-round system.

## LLM Description
This pull request introduces a comprehensive complaint and confirmation mechanism to the FROST key generation protocol, improving its robustness and security. The changes add new group status tracking, participant complaint/confirmation logic, and corresponding events, as well as update the validator protocol to support confirmation actions. These enhancements ensure that key generation can be finalized only after all complaints are resolved, and allow detection of compromised groups.

**Smart contract changes:**

_Group status and complaint handling:_

* Added a `GroupStatus` enum to `FROSTCoordinator.sol` and updated `GroupParameters` to include a `status` field for more granular tracking of the key generation process.
* Implemented complaint and confirmation mechanisms: participants can submit complaints against others, respond to complaints, and confirm their participation only when all complaints are resolved. The group can be marked as compromised if complaints reach the threshold. [[1]](diffhunk://#diff-9ed38bdaa749aaa2ed89de9e81d2c701723bff2581cc6bef42c839c3e343732dL183-R251) [[2]](diffhunk://#diff-eba253e288acd1184648bd60e56e45831f8d3898757481ce35e2d09433f66335R16-R44) [[3]](diffhunk://#diff-eba253e288acd1184648bd60e56e45831f8d3898757481ce35e2d09433f66335R85-R114)
* Added new events (`KeyGenConfirmed`, `KeyGenComplained`, `KeyGenComplaintResponded`) and errors to support complaint and confirmation flows. [[1]](diffhunk://#diff-9ed38bdaa749aaa2ed89de9e81d2c701723bff2581cc6bef42c839c3e343732dR85-R89) [[2]](diffhunk://#diff-9ed38bdaa749aaa2ed89de9e81d2c701723bff2581cc6bef42c839c3e343732dL90-R106)

_Key generation flow updates:_

* Refactored key generation logic to use the new `status` field instead of the previous `sequence`-based state tracking, ensuring proper transitions between committing, sharing, confirming, and finalized states. [[1]](diffhunk://#diff-9ed38bdaa749aaa2ed89de9e81d2c701723bff2581cc6bef42c839c3e343732dL112-R129) [[2]](diffhunk://#diff-9ed38bdaa749aaa2ed89de9e81d2c701723bff2581cc6bef42c839c3e343732dL133-R146) [[3]](diffhunk://#diff-9ed38bdaa749aaa2ed89de9e81d2c701723bff2581cc6bef42c839c3e343732dL173-R188) [[4]](diffhunk://#diff-9ed38bdaa749aaa2ed89de9e81d2c701723bff2581cc6bef42c839c3e343732dL215-R270)
* Updated participant map to track complaint status, confirmation, and accusation counts, with new methods for complaint submission, response, confirmation, and participation checks. [[1]](diffhunk://#diff-eba253e288acd1184648bd60e56e45831f8d3898757481ce35e2d09433f66335R16-R44) [[2]](diffhunk://#diff-eba253e288acd1184648bd60e56e45831f8d3898757481ce35e2d09433f66335R85-R114) [[3]](diffhunk://#diff-eba253e288acd1184648bd60e56e45831f8d3898757481ce35e2d09433f66335R126-R131)

**Validator protocol changes:**

_Protocol action support:_

* Added `ConfirmKeyGen` action type and protocol handler, allowing validators to confirm key generation on-chain after complaints are resolved. [[1]](diffhunk://#diff-b7b262fe3349b48c703c8895de01bfd1c8941198e8c4410e7afeb9180da782efR5) [[2]](diffhunk://#diff-b7b262fe3349b48c703c8895de01bfd1c8941198e8c4410e7afeb9180da782efR78-R79) [[3]](diffhunk://#diff-b7b262fe3349b48c703c8895de01bfd1c8941198e8c4410e7afeb9180da782efR98-R99) [[4]](diffhunk://#diff-4a1dfcacb819af2ad989a1e95bff18914a1a185169806527e1170bec56b78bbbR7) [[5]](diffhunk://#diff-f3ee0ab85cc6c03789af38abf6fa2077f17d76cb11e152e98ff99197fcaf2fd0R73-R81)
* Refactored publishing secret shares and confirmation methods to support callback and non-callback variants, aligning with new contract methods.

_Event schema updates:_

* Added schema for the new `KeyGenConfirmed` event for better event parsing and validation.

_Test and documentation updates:_

* Updated tests to cover the confirmation step in key generation.
* Minor documentation typo fix ("recive" → "receive").

### DIFF
--- a/contracts/src/FROSTCoordinator.sol
+++ b/contracts/src/FROSTCoordinator.sol
@@ -103,7 +103,6 @@ contract FROSTCoordinator {
     error InvalidGroupCommitment();
     error GroupNotInitialized();
     error GroupNotReady();
-    error InvalidAccused();
     error InvalidSecretShare();
     error InvalidMessage();
     error NotSigning();
@@ -227,10 +226,7 @@ contract FROSTCoordinator {
             GroupNotReady()
         );
         FROST.Identifier plaintiff = group.participants.identifierOf(msg.sender);
-        require(!FROST.eq(plaintiff, accused), InvalidAccused());
-        require(group.participants.isParticipating(accused), InvalidAccused());
-        group.participants.complain(plaintiff, accused);
-        if (group.participants.getAccusationCount(accused) >= (group.parameters.threshold)) {
+        if (group.participants.complain(plaintiff, accused) >= group.parameters.threshold) {
             group.parameters.status = GroupStatus.COMPROMISED;
         }
         emit KeyGenComplained(gid, plaintiff, accused);
@@ -267,7 +263,7 @@ contract FROSTCoordinator {
         Group storage group = $groups[gid];
         GroupParameters memory parameters = group.parameters;
         require(parameters.count > 0, GroupNotInitialized());
-        require(parameters.pending == 0, GroupNotReady());
+        require(parameters.status == GroupStatus.FINALIZED, GroupNotReady());
         uint64 sequence = parameters.sequence++;
         sid = FROSTSignatureId.create(gid, sequence);
         Signature storage signature = $signatures[sid];
--- a/contracts/src/libraries/FROSTParticipantMap.sol
+++ b/contracts/src/libraries/FROSTParticipantMap.sol
@@ -39,6 +39,7 @@ library FROSTParticipantMap {
     error NotParticipating();
     error AlreadySet();
     error ComplaintAlreadySubmitted();
+    error InvalidAccused();
     error ParticipantAlreadyFinalized();
     error ComplaintNotSubmitted();
     error UnrespondedComplaintExists();
@@ -83,12 +84,17 @@ library FROSTParticipantMap {
     }
 
     /// @notice Submits a complaint from a plaintiff against an accused.
-    function complain(T storage self, FROST.Identifier plaintiff, FROST.Identifier accused) internal {
+    function complain(T storage self, FROST.Identifier plaintiff, FROST.Identifier accused)
+        internal
+        returns (uint64 totalAccusations)
+    {
         require(self.states[plaintiff].statuses[accused] == ComplaintStatus.NONE, ComplaintAlreadySubmitted());
         require(!self.states[plaintiff].confirmed, ParticipantAlreadyFinalized());
+        require(isParticipating(self, accused), InvalidAccused());
+        require(!FROST.identifierEq(plaintiff, accused), InvalidAccused());
         self.states[plaintiff].statuses[accused] = ComplaintStatus.SUBMITTED;
         self.states[plaintiff].complaints++;
-        self.states[accused].accusations++;
+        totalAccusations = ++self.states[accused].accusations;
     }
 
     /// @notice Responds to a complaint from a plaintiff against an accused.
@@ -105,11 +111,6 @@ library FROSTParticipantMap {
         require(self.states[participant].complaints == 0, UnrespondedComplaintExists());
         require(!self.states[participant].confirmed, ParticipantAlreadyFinalized());
         self.states[participant].confirmed = true;
-    }
-
-    /// @notice Gets the count of complaints received by an accused.
-    function getAccusationCount(T storage self, FROST.Identifier accused) internal view returns (uint64) {
-        return self.states[accused].accusations;
     }
 
     /// @notice Gets the participant's identifier.
--- a/contracts/src/libraries/FROST.sol
+++ b/contracts/src/libraries/FROST.sol
@@ -277,7 +277,7 @@ library FROST {
         }
     }
 
-    function eq(Identifier a, Identifier b) internal pure returns (bool) {
+    function identifierEq(Identifier a, Identifier b) internal pure returns (bool) {
         return Identifier.unwrap(a) == Identifier.unwrap(b);
     }
 }


contracts/src/libraries/FROSTFinalization.sol
```diff
@@ -0,0 +1,98 @@
...
+
+    struct ComplaintSubmittedCount {
+        uint64 complaintRaised;
+        bool finalized;
+    }
+
+    struct T {
+        mapping(FROST.Identifier plaintiff => mapping(FROST.Identifier accused => ComplaintStatus)) statuses;
+        mapping(FROST.Identifier plaintiff => ComplaintSubmittedCount) submittedCounts;
+        mapping(FROST.Identifier accused => uint64) receivedCounts;
```
rmeissner: This is currently not used, what is the purpose of this mapping?

contracts/src/FROSTCoordinator.sol
```diff
@@ -194,6 +210,54 @@ contract FROSTCoordinator {
...
+    }
+
+    /// @notice Resolve a complaint from a plaintiff against the sender by
+    ///         providing the secret share publicly.
+    function keyGenResolveComplain(FROSTGroupId.T gid, FROST.Identifier plaintiff, uint256 secretShare) external {
+        Group storage group = $groups[gid];
+        require(group.parameters.status == GroupStatus.CONFIRMING, GroupNotCommitted());
+        FROST.Identifier accused = group.participants.identifierOf(msg.sender);
+        group.finalizations.resolve(plaintiff, accused);
+        emit KeyGenComplaintResolved(gid, plaintiff, accused, secretShare);
```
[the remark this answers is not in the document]
rmeissner: I have to agree with Copilot here, it is misleading that we call this state "resolved" as this is not guaranteed. While I don't think we need to check the correctness of the secretShare (every validator can do this offchain), for me the state name sets a wrong expectation. Maybe "responded" would be a better name.

contracts/src/FROSTCoordinator.sol
```diff
@@ -180,20 +197,55 @@ contract FROSTCoordinator {
...
         if (completed) {
             callback.target.onKeyGenCompleted(gid, callback.context);
         }
     }
 
+    /// @notice Submit a complaint from the sender against another participant
+    ///         during key generation.
+    function keyGenComplain(FROSTGroupId.T gid, FROST.Identifier accused) external {
+        Group storage group = $groups[gid];
+        require(group.parameters.status == GroupStatus.CONFIRMING, GroupNotCommitted());
```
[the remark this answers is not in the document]
nlordell: Also, I would potentially rename `GroupNotCommitted` to `GroupNotReady`, as a finalized group can also not be complained (and it is a little more broad error to indicate that the group is not in the correct state).

contracts/src/FROSTCoordinator.sol
```diff
@@ -194,6 +210,54 @@ contract FROSTCoordinator {
...
+
+    /// @notice Submit a complaint from the sender against another participant
+    ///         during key generation.
+    function keyGenComplain(FROSTGroupId.T gid, FROST.Identifier accused) external {
+        Group storage group = $groups[gid];
+        require(group.parameters.status == GroupStatus.CONFIRMING, GroupNotCommitted());
+        // Check if the accused identifier is valid
+        require(group.participants.getKey(accused).x != 0, InvalidAccused());
+        FROST.Identifier plaintiff = group.participants.identifierOf(msg.sender);
+        group.finalizations.submit(plaintiff, accused);
```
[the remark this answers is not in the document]
nlordell: This is correct.

contracts/src/FROSTCoordinator.sol
```diff
@@ -180,20 +197,55 @@ contract FROSTCoordinator {
...
 
+    /// @notice Submit a complaint from the sender against another participant
+    ///         during key generation.
+    function keyGenComplain(FROSTGroupId.T gid, FROST.Identifier accused) external {
+        Group storage group = $groups[gid];
+        require(group.parameters.status == GroupStatus.CONFIRMING, GroupNotCommitted());
+        // Check if the accused identifier is valid
+        require(group.participants.getKey(accused).x != 0, InvalidAccused());
+        FROST.Identifier plaintiff = group.participants.identifierOf(msg.sender);
+        group.finalizations.submit(plaintiff, accused);
```
nlordell:
```suggestion
        group.finalizations.complain(plaintiff, accused);
```

contracts/src/FROSTCoordinator.sol
```diff
@@ -24,13 +26,23 @@ contract FROSTCoordinator {
...
         Secp256k1.Point key;
+        FROSTFinalization.T finalizations;
+    }
+
+    enum GroupStatus {
+        UNINITIALIZED,
+        COMMITTING,
+        SHARING,
+        CONFIRMING,
+        FINALIZED
```
nlordell: I think we need an additional `COMPROMISED` state for when there are `threshold` complaints to the same participant.

contracts/src/FROSTCoordinator.sol
```diff
@@ -180,20 +197,55 @@ contract FROSTCoordinator {
...
-        Callback calldata callback
-    ) public returns (bool completed) {
-        completed = keyGenSecretShare(gid, share);
+    /// @notice Finalize the Key Generation ceremony for the sender if no unresolved complaints exist.
+    function keyGenFinalize(FROSTGroupId.T gid) public returns (bool completed) {
+        Group storage group = $groups[gid];
+        GroupParameters memory parameters = group.parameters;
+        require(parameters.status == GroupStatus.CONFIRMING, GroupNotCommitted());
+        FROST.Identifier identifier = group.participants.identifierOf(msg.sender);
+        group.finalizations.finalizeValidator(identifier);
```
nlordell:
```suggestion
        group.finalizations.confirm(identifier);
```

contracts/src/FROSTCoordinator.sol
```diff
@@ -180,20 +197,55 @@ contract FROSTCoordinator {
...
+    ///         during key generation.
+    function keyGenComplain(FROSTGroupId.T gid, FROST.Identifier accused) external {
+        Group storage group = $groups[gid];
+        require(group.parameters.status == GroupStatus.CONFIRMING, GroupNotCommitted());
+        // Check if the accused identifier is valid
+        require(group.participants.getKey(accused).x != 0, InvalidAccused());
+        FROST.Identifier plaintiff = group.participants.identifierOf(msg.sender);
+        group.finalizations.submit(plaintiff, accused);
+        emit KeyGenComplained(gid, plaintiff, accused);
+    }
```
nlordell: The group needs to enter a "COMPROMISED" state in case of `threshold` complaints to a participant.

rmeissner (approved):
To Note: this is still missing the complain flow in the validator, we should track this somewhere (either as a GitHub issue or Notion project)

contracts/src/libraries/FROSTParticipantMap.sol
```diff
@@ -64,6 +82,36 @@ library FROSTParticipantMap {
...
+    function confirm(T storage self, FROST.Identifier participant) internal {
+        require(self.states[participant].complaints == 0, UnrespondedComplaintExists());
+        require(!self.states[participant].confirmed, ParticipantAlreadyFinalized());
+        self.states[participant].confirmed = true;
+    }
+
+    /// @notice Gets the count of complaints received by an accused.
+    function getAccusationCount(T storage self, FROST.Identifier accused) internal view returns (uint64) {
+        return self.states[accused].accusations;
+    }
```
nlordell: This should be returned with the `complain` to avoid additional storage reads.

validator/src/machine/keygen/confirmed.ts
```diff
@@ -0,0 +1,82 @@
...
+	keyGenClient: KeyGenClient,
+	signingClient: SigningClient,
+	consensusState: ConsensusState,
+	machineStates: MachineStates,
+	eventArgs: unknown,
+	logger?: (msg: unknown) => void,
+): Promise<StateDiff> => {
+	// A participant has confirmed their participation in the key gen ceremony
+	// Ignore if not in "collecting_confirmations" state
+	if (machineStates.rollover.id !== "collecting_confirmations") {
```
nlordell: Just adding a note here, but I don't think this is correct: if we have 3 participants where 1 and 2 honestly share secret shares and 3 dishonestly shares with 2; then 1 and 3 can enter the `collecting_confirmations` state, but 2 cannot because it is missing a share resulting from the complaint system. It needs to additionally wait for a `KeyGenComplaintResponse` with the missing share before it can construct its signing share. This means that we need to potentially allow confirmations while waiting for pending complaints.

I would add a note here, as it is something to consider for when we implement complaints in the validator (so not for this PR).

PR #59, opened 2025-12-04
Persist Transition and Action Queue

rmeissner:
- Extract log parsing from state machine
- Define proper state transitions types (block clock and events)
- Add schemas for transition and action persistation
- Introduce SQLite backed Queue

### DIFF
--- a/validator/src/machine/transitions/onchain.ts
+++ b/validator/src/machine/transitions/onchain.ts
@@ -1,7 +1,9 @@
 import {
+	epochProposedEventSchema,
 	epochStagedEventSchema,
 	keyGenCommittedEventSchema,
 	keyGenConfirmedEventSchema,
+	keyGenEventSchema,
 	keyGenSecretSharedEventSchema,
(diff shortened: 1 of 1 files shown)


nlordell (approved):

validator/src/consensus/protocol/base.ts
```diff
@@ -58,9 +56,12 @@ export abstract class BaseProtocol implements ShieldnetProtocol {
 		const executionDelay = action.retryCount > 0 ? ERROR_RETRY_DELAY : 0;
 		setTimeout(() => {
 			this.performAction(action)
+				.then(() => {
+					// If action was successfully executed, remove it from queue
+					this.#actionQueue.pop();
+				})
 				.catch(() => {
 					action.retryCount++;
```
remedcu: Doesn't this only update in memory and not write to the DB?

rmeissner: hmm yeah, will need to fix this

PR #65, opened 2025-12-09
Staking NatSpec (Opinionated)

remedcu:
This PR contains the opinionated NatSpec "style" for the Staking contract.

If this style is good, I will use the same style on all the other contracts as well, as a first step towards standardization.

### DIFF
--- a/contracts/src/Staking.sol
+++ b/contracts/src/Staking.sol
@@ -14,7 +14,7 @@ contract Staking is Ownable {
     // STRUCTS
     // ============================================================
 
-    /*
+    /**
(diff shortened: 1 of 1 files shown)


contracts/src/Staking.sol
```diff
@@ -15,33 +15,43 @@ contract Staking is Ownable {
     // ============================================================
 
     /*
```
nlordell: Note that this isn't valid NatSpec - it _MUST_ be either `/** */` or `///`. I personally prefer the latter (its much easier to properly indent and you don't need to manage closing tags), but we should fix all the NatSpec in the contract.

PR #68, opened 2025-12-10
Finalization Flow update

remedcu:
The previous implementation had two issues identified during review:

- Incorrect transition trigger: The state machine transitioned to sign_rollover after receiving our own confirmation, rather than waiting for all participants to confirm. This meant we marked the group as "setup" prematurely.

- Empty message placeholder: Because we transitioned early, we had to set an empty message ("0x") and rely on the EpochProposed event to populate it. This was a workaround that deviated from the spec.

This PR addresses that by tracking confirmation and transitioning only after all participants have confirmed.


validator/src/machine/consensus/epochProposed.ts
```diff
@@ -2,25 +2,34 @@ import type { ShieldnetProtocol } from "../../consensus/protocol/types.js";
 import type { VerificationEngine } from "../../consensus/verify/engine.js";
 import type { EpochRolloverPacket } from "../../consensus/verify/rollover/schemas.js";
 import type { EpochProposedEvent } from "../transitions/types.js";
-import type { MachineConfig, MachineStates, StateDiff } from "../types.js";
+import type { MachineStates, StateDiff } from "../types.js";
 
 export const handleEpochProposed = async (
```
rmeissner: I would remove this method completely and the message is already verified (https://github.com/safe-research/shieldnet/pull/68/files#diff-7c2e07da90f8ad73f0cfdac7076dceb43f8da263edcd1b9bcd4e6b2db0eda712R110)

validator/src/machine/keygen/confirmed.ts
```diff
@@ -25,56 +31,105 @@ export const handleKeyGenConfirmed = async (
...
+					lastParticipant: event.identifier,
+				},
+			};
+		}
+		// All confirmed for genesis group - start preprocessing and return to waiting state
+		logger?.("Genesis group all confirmations received, starting preprocessing");
+		const consensus: ConsensusDiff = {
+			groupPendingNonces: [groupId, true],
+		};
+		const nonceTreeRoot = signingClient.generateNonceTree(groupId);
```
rmeissner: As we do this in the epoch staged event I would not do this here, it feels duplicated

validator/src/machine/keygen/confirmed.ts
```diff
@@ -25,56 +31,105 @@ export const handleKeyGenConfirmed = async (
...
 
-	// Check if this is our own confirmation
-	const ourParticipantId = keyGenClient.participantId(groupId);
-	if (event.identifier !== ourParticipantId) {
-		// Not our confirmation, just track
-		logger?.(`Group ${event.gid} confirmation from ${event.identifier} (not ours)`);
+	// Track this confirmation
+	const confirmedParticipants = [...machineStates.rollover.confirmedParticipants, event.identifier];
+	const participants = signingClient.participants(groupId);
+	const allConfirmed = confirmedParticipants.length === participants.length;
```
rmeissner: An array does not prevent duplicated identifiers. In the spec we use a set (which I don't think is necessary for the validator implementation). I would check that for each participant we have a confirmation.

validator/src/consensus/signing/client.ts
```diff
@@ -167,8 +167,12 @@ export class SigningClient {
 		};
 	}
 
-	signers(groupId: GroupId): ParticipantId[] {
-		return this.#storage.signers(groupId);
+	signers(signatureId: SignatureId): ParticipantId[] {
+		return this.#storage.signers(signatureId);
+	}
+
+	participants(groupId: GroupId): ParticipantId[] {
```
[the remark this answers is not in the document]
remedcu: Now this is used in the signing flow as well 😅

validator/src/machine/keygen/confirmed.ts
```diff
@@ -25,56 +31,105 @@ export const handleKeyGenConfirmed = async (
 	}
 	const groupId = event.gid;
 
-	// Check if this is our own confirmation
-	const ourParticipantId = keyGenClient.participantId(groupId);
-	if (event.identifier !== ourParticipantId) {
-		// Not our confirmation, just track
-		logger?.(`Group ${event.gid} confirmation from ${event.identifier} (not ours)`);
+	// Track this confirmation
+	const confirmedParticipants = [...machineStates.rollover.confirmedParticipants, event.identifier];
```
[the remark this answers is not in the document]
remedcu: Updated to `sharesFrom`

validator/src/machine/keygen/confirmed.ts
remedcu: Updated to check that every participant has an item in `sharesFrom`.

PR #73, opened 2025-12-11
Natspec Styling and some basic description

remedcu:
This is part of the NatSpec changes, started with #65 

This is part 2, where we use the same styling introduced in #65, and have used the same in all the contracts in the `src` folder.

- There should be no logical changes.
- Only basic description and information added in NatSpec (more details will follow).
- NatSpec styling should be consistent across all the smart contracts now.

### DIFF
--- a/contracts/src/libraries/FROST.sol
+++ b/contracts/src/libraries/FROST.sol
@@ -6,6 +6,8 @@ import {Secp256k1} from "@/libraries/Secp256k1.sol";
 /**
  * @title FROST
  * @notice Implementation of the FROST(secp256k1, SHA-256) ciphersuite.
+ * @dev FROST Implementation is based on RFC-9591 (https://datatracker.ietf.org/doc/html/rfc9591).
+ *      The dst (Domain Separation Tag) values for `_hashToField` are based on https://datatracker.ietf.org/doc/html/rfc9591#section-6.5
  */
 library FROST {
     using Secp256k1 for Secp256k1.Point;
@@ -22,8 +24,8 @@ library FROST {
     /**
      * @notice Group commitment used in FROST signing.
      * @param identifier The participant identifier associated with this commitment.
-     * @param d The first public nonce commitment point.
-     * @param e The second public nonce commitment point.
+     * @param d The hiding nonce commitment point.
+     * @param e The binding nonce commitment point.
      */
     struct Commitment {
         Identifier identifier;
@@ -33,7 +35,7 @@ library FROST {
 
     /**
      * @notice A FROST group signature.
-     * @param r The group commitment point.
+     * @param r The group commitment represented as a point.
      * @param z The scalar signature component.
      */
     struct Signature {
@@ -43,7 +45,7 @@ library FROST {
 
     /**
      * @notice A FROST signature share for a single participant.
-     * @param r The participant commitment point.
+     * @param r The participant commitment represented as a point.
      * @param z The participant scalar share.
      * @param l The Lagrange coefficient for this participant.
      */
@@ -58,7 +60,7 @@ library FROST {
     // ============================================================
 
     /**
-     * @notice Thrown when an identifier is zero or otherwise invalid.
+     * @notice Thrown when a participant identifier is zero or otherwise invalid.
      */
     error InvalidIdentifier();
 
@@ -100,25 +102,22 @@ library FROST {
 
     /**
      * @notice Generates a random nonce from randomness and a secret key.
-     * @dev Implements the RFC-9591 `nonce_generate` function.
      * @param random The source randomness.
      * @param secret The participant secret key.
      * @return n The derived nonce scalar.
+     * @dev Implements the RFC-9591 `nonce_generate` function (https://datatracker.ietf.org/doc/html/rfc9591#section-4.1).
      */
     function nonce(bytes32 random, uint256 secret) internal view returns (uint256 n) {
-        // The RFC-9591 `nonce_generate` function.
-        // <https://datatracker.ietf.org/doc/html/rfc9591#section-4.1>
-
         return _h3(abi.encodePacked(random, secret));
     }
 
     /**
      * @notice Computes the binding factors for a message and group commitments.
-     * @dev Implements the RFC-9591 `compute_binding_factors` function.
      * @param y The group public key.
      * @param commitments The list of participant commitments.
      * @param message The message being signed.
      * @return rho The binding factors for each participant.
+     * @dev Implements the RFC-9591 `compute_binding_factors` function.
      */
     function bindingFactors(Secp256k1.Point memory y, Commitment[] memory commitments, bytes32 message)
         internal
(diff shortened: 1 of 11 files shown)


contracts/src/libraries/FROST.sol
```diff
@@ -3,54 +3,123 @@ pragma solidity ^0.8.30;
...
+ */
 library FROST {
     using Secp256k1 for Secp256k1.Point;
 
+    // ============================================================
+    // TYPES AND STRUCTS
+    // ============================================================
+
+    /**
+     * @notice Type alias for a FROST participant identifier.
```
rmeissner: When do we call something "identifier" and when do we call it "id"?

contracts/src/libraries/FROST.sol
```diff
@@ -3,54 +3,123 @@ pragma solidity ^0.8.30;
...
+     */
     type Identifier is uint256;
 
+    /**
+     * @notice Group commitment used in FROST signing.
+     * @param identifier The participant identifier associated with this commitment.
+     * @param d The first public nonce commitment point.
```
rmeissner: it is not the first and second, but rather a hidding and binding nonce commitment (see https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-frost-08#section-5.1). 

Ref: https://github.com/safe-research/shieldnet/blob/main/validator/src/consensus/signing/nonces.ts#L11-L20

contracts/src/libraries/FROST.sol
```diff
@@ -3,54 +3,123 @@ pragma solidity ^0.8.30;
...
+     */
     struct Commitment {
         Identifier identifier;
         Secp256k1.Point d;
         Secp256k1.Point e;
     }
 
+    /**
+     * @notice A FROST group signature.
+     * @param r The group commitment point.
```
rmeissner: nit: I would write "The group commitment represented as a point"

contracts/src/libraries/FROST.sol
```diff
@@ -93,14 +168,26 @@ library FROST {
...
         uint256 c = challenge(signature.r, y, message);
         Secp256k1.mulmuladd(signature.z, c, y, signature.r);
     }
 
-    /// @notice Verifies a FROST signature share.
+    /**
+     * @notice Verifies a FROST signature share.
+     * @param group The group public key.
+     * @param r The group commitment point.
+     * @param participant The participant public key.
```
rmeissner: this is called a verification share in FROST, right?

See https://frost.zfnd.org/frost.html#key-generation

contracts/src/libraries/FROST.sol
```diff
@@ -155,30 +258,75 @@ library FROST {
...
+    // PRIVATE FUNCTIONS - HASH DOMAINS
+    // ============================================================
+
+    /**
+     * @notice Hashes input to a field element for binding factors.
+     * @param input The input bytes.
+     * @return result The field element.
+     */
     function _h1(bytes memory input) private view returns (uint256 result) {
         return _hashToField(input, "FROST-secp256k1-SHA256-v1rho\x00\x00\x00\x1c");
```
rmeissner: Is it worth mentioning that this string is specific to our implementation/ cyphersuit?

contracts/src/libraries/FROSTGroupId.sol
```diff
@@ -1,14 +1,38 @@
...
     error InvalidGroupId();
 
-    /// @notice Computes the deterministic group ID for a given configuration.
+    // ============================================================
+    // INTERNAL FUNCTIONS
+    // ============================================================
+
+    /**
+     * @notice Computes the deterministic group ID for a given configuration.
+     * @param participants The participants hash.
```
rmeissner: I would call it the participants (merkle) root.

contracts/src/libraries/FROSTSignatureId.sol
```diff
@@ -3,37 +3,77 @@ pragma solidity ^0.8.30;
...
+    /**
+     * @notice Computes the deterministic signature ID for a given configuration.
+     * @param gid The group ID.
+     * @param seq The sequence number.
+     * @return result The computed signature ID.
+     * @dev We encode `sequence + 1` in the signature ID. This allows us to tell
+     *      whether an ID belongs to a group or a signature by non-zero LSBs.
+     */
     function create(FROSTGroupId.T gid, uint64 seq) internal pure returns (T result) {
         // We encode `sequence + 1` in the signature ID. This allows us to tell
```
rmeissner: This is now duplicated, should we remove it?

contracts/src/libraries/MetaTransaction.sol
```diff
@@ -1,9 +1,25 @@
...
+ */
 library MetaTransaction {
+    // ============================================================
+    // STRUCTS
+    // ============================================================
+
+    /**
+     * @notice Represents a meta-transaction.
+     * @param chainId The chain ID for the transaction.
+     * @param account The account address initiating the transaction.
```
rmeissner: nit: I would write "the account address that is executing the transaction".

For me initiating is slightly different to executing, i.e. the initiating account is the one that sends the transaction to the Safe smart account.

contracts/src/libraries/Secp256k1.sol
```diff
@@ -76,22 +120,26 @@ library Secp256k1 {
...
     }
 
-    /// @notice Verifies that `z⋅G - e⋅P = R`.
+    /**
+     * @notice Verifies that `z⋅G - e⋅P = R`.
+     * @param z The scalar multiplier for the generator point.
+     * @param e The scalar multiplier for point P.
+     * @param p The point P.
+     * @param r The expected result point R (witness).
+     * @dev See <https://ethresear.ch/t/you-can-kinda-abuse-ecrecover-to-do-ecmul-in-secp256k1-today/2384>
```
nlordell: nit: The ordering for the NatSpec is not consistent (sometimes, `@dev` comes before the apramters and return, sometimes after). Can we make it consistent.

I would also _love_ to have a `@custom:reference` or such tag for external references like this.


```suggestion
     * @custom:reference https://ethresear.ch/t/you-can-kinda-abuse-ecrecover-to-do-ecmul-in-secp256k1-today/2384
```

We can use it for other links as well.

nlordell: My personal preference is `@dev` at the bottom - like it is here.

nlordell (approved):
Looks mostly good to me. I didn't review everything, but randomly sampled a good amount of comments.

contracts/src/libraries/FROST.sol
```diff
@@ -3,54 +3,123 @@ pragma solidity ^0.8.30;
...
         uint256 z;
         uint256 l;
     }
 
+    // ============================================================
+    // ERRORS
+    // ============================================================
+
+    /**
+     * @notice Thrown when an identifier is zero or otherwise invalid.
```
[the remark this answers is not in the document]
remedcu: Right now in the FROST code it is only used for a participant identifier.

contracts/src/libraries/FROSTParticipantMap.sol
```diff
@@ -6,58 +6,146 @@ import {MerkleProof} from "@oz/utils/cryptography/MerkleProof.sol";
...
     function init(T storage self, bytes32 root) internal {
         require(root != bytes32(0), InvalidRootHash());
         require(self.root == bytes32(0), AlreadyInitialized());
         self.root = root;
     }
 
-    /// @notice Returns whether or not a group is initialized. A sealed group is
-    ///         not considered initialized.
+    /**
+     * @notice Returns whether or not a group is initialized. A sealed group is
```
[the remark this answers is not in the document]
remedcu: Not sure.

CC @nlordell

contracts/src/libraries/FROST.sol
remedcu: This will be handled in either the next PR or the Glossary PR. Right now, I am not sure.

PR #83, opened 2025-12-15
Simplify Last Participant Handling

nlordell:
As discussed on Slack, the handling of the `last_participant` and how it related to the `responsible` participant was unclear.

This PR simplifies handling by removing the `last_participant` everywhere except for `share` collection. There, the last participant is still needed (and represents the last participant to broadcast a signing share), as it is used to determine the responsible participant for restarting the signing ceremony in case of a timeout.

### DIFF
--- a/spec/validator.ml
+++ b/spec/validator.ml
@@ -77,7 +77,6 @@ type rollover_finalization = {
   shares : FROST.KeyGen.secret_share ParticipantMap.t local;
   complaints : rollover_complaints ParticipantMap.t;
   confirmations : ParticipantSet.t;
-  last_participant : FROST.Identifier.t;
   complaint_deadline : int;
   response_deadline : int;
   confirm_deadline : int;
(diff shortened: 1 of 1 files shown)


spec/validator.ml
```diff
@@ -987,20 +988,48 @@ struct
...
                 (None, [])
               else
-                let state', action, actor =
+                let state', actions' =
                   match signature.status with
+                  | Waiting_for_sign_request _ ->
+                      ( Waiting_for_sign_request { responsible = None },
+                        `Coordinator_sign (signature.epoch.group.id, message)
+                        :: actions )
+                  | Collecting_sign_nonces { id; nonces } ->
```
rmeissner: Note: This case is currently different in the validator service. If this times out then the validator triggers a new signature request with a smaller set signers.

So this is something we should adjust in the Validator at some point

rmeissner (approved):

spec/validator.ml
nlordell: This behaviour didn't change in this PR (it is [like this already](https://github.com/safe-research/shieldnet/blob/3425d11b88f97b51a1cc129e812d0d2919f62aaa/spec/validator.ml#L1018-L1025)). I am personally a fan of this, as signing moves forward more quickly in case nodes are down.

PR #100, opened 2025-12-18
Natspec Stage 3 Changes

remedcu:
TLDR

- This includes the final changes to the NatSpec for detailed information.

Note:
- If any of the other contracts require more information or are still ambiguous, please mention so I can also add extra information. 
- This does not involve any of the Glossary-based changes, which will be a separate PR (mainly standardization/name-change of variable names, etc)
- Technical documentation would also be a separate PR (extra information outside the contract, could include flows, etc.)

### DIFF
--- a/contracts/foundry.toml
+++ b/contracts/foundry.toml
@@ -3,6 +3,7 @@ out = "build/out"
 cache_path = "build/cache"
 broadcast = "build/broadcast"
 
+solc_version = "0.8.30"
 remappings = [
   "@/=src/",
   "@test/=test/",


contracts/src/libraries/FROST.sol
```diff
@@ -262,54 +262,76 @@ library FROST {
     // ============================================================
 
     /**
-     * @notice Hashes input to a field element for binding factors.
-     * @param input The input bytes.
+     * @notice h1: Hashes input to a field element for calculating binding factors (rho).
+     * @param input The input bytes, which include group public key, message hash, and commitments hash.
```
nlordell: The input is just input bytes to the hash function. How the hash function is used is specific to the outer algorithm, and should not be included in the natspec here.


```suggestion
     * @param input The input bytes.
```

contracts/src/libraries/FROST.sol
```diff
@@ -262,54 +262,76 @@ library FROST {
...
      * @return result The field element.
+     * @dev Uses a domain separation tag "FROST-secp256k1-SHA256-v1rho" as required by
+     *      RFC 9591, Section 6.5. Domain separation is critical to prevent an attacker
+     *      from reusing a signature or proof from one context in another. The DST is
+     *      packed into a bytes32 with its length (28 / 0x1c) as the final byte,
+     *      following RFC 9380 recommendations for DST formatting.
```
nlordell: This summary is quite inaccurate IMO:


```suggestion
     * @dev Uses a domain separation tag "FROST-secp256k1-SHA256-v1rho" as required by
     *      RFC 9591, Section 6.5. The DST is
     *      packed into a bytes32 with its length (28 / 0x1c) as the final byte.
```

The "following RFC 9380 recommendations for DST formatting" is not accurate - this is just how we do strings without allocations in Solidity.

Same goes for other hash functions.

contracts/src/libraries/FROST.sol
nlordell: In particular this is how the last byte is used.

contracts/src/libraries/FROSTSignatureShares.sol
```diff
@@ -15,17 +15,22 @@ library FROSTSignatureShares {
...
 
     /**
-     * @notice Aggregated signature data for a commitment root.
-     * @param participants Mapping from participant identifier to their leaf hash.
-     * @param signature The accumulated group signature.
+     * @notice Aggregated signature data for a specific group of signing participants.
+     * @param participants A mapping to track which participants have already submitted their
+     *                     shares for this ceremony, preventing replay or duplicate submissions.
+     *                     The value is the participant's leaf hash in the Merkle tree.
+     * @param signature The accumulated group signature, which is built up as each participant
```
nlordell: Idem on NatSpec, `@param` and `struct`s.

contracts/src/libraries/Secp256k1.sol
```diff
@@ -39,17 +39,25 @@ library Secp256k1 {
     // ============================================================
 
     /**
-     * @notice The b coefficient of the secp256k1 curve equation y² = x³ + b.
+     * @notice The 'b' coefficient of the secp256k1 curve equation y² = x³ + ax + b.
+     * @dev For secp256k1, the 'a' coefficient is 0, and 'b' is 7. This is a defining
+     *      parameter of the curve, as specified in standards like "SEC 2: Recommended
+     *      Elliptic Curve Domain Parameters".
```
nlordell: TIL!

contracts/src/libraries/Secp256k1.sol
```diff
@@ -39,17 +39,25 @@ library Secp256k1 {
...
      */
     uint256 internal constant B = 7;
 
     /**
-     * @notice The prime field modulus for secp256k1.
+     * @notice The prime field modulus for secp256k1, defining the finite field F_p.
+     * @dev All arithmetic operations on the curve's coordinates are performed modulo this
+     *      prime number. Its value is 2^256 - 2^32 - 977 as specified in standards.
```
nlordell:
```suggestion
     *      prime number. Its value is 2^256 - 2^32 - 977 as specified in SEC 2.
```

contracts/src/libraries/Secp256k1.sol
```diff
@@ -39,17 +39,25 @@ library Secp256k1 {
...
      */
     uint256 internal constant P = 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f;
 
     /**
-     * @notice The order of the secp256k1 curve.
+     * @notice The order of the generator point of the secp256k1 curve.
+     * @dev This is a large prime number that specifies the number of distinct points
+     *      that can be generated by repeatedly adding the generator point to itself.
+     *      All scalar multiplications are performed modulo N.
```
nlordell:
```suggestion
     * @dev All scalar arithmetic are performed modulo N.
```

Nit: I don't think we need to explain the cryptography behind `secp256k1` in the comments. This opens us up to bug bounty submissions of the kind "your cryptography explanation is wrong" etc.

contracts/src/libraries/FROSTParticipantMap.sol
```diff
@@ -33,11 +33,16 @@ library FROSTParticipantMap {
     // ============================================================
 
     /**
-     * @notice State tracking for a single participant.
-     * @param confirmed Whether the participant has confirmed key generation.
-     * @param complaints The number of unresolved complaints submitted by this participant.
-     * @param accusations The number of accusations against this participant.
-     * @param statuses Mapping of complaint statuses against other participants.
+     * @notice State tracking for a single participant during DKG.
+     * @param confirmed Whether the participant has successfully confirmed their participation
```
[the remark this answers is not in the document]
remedcu: You are right, only `@notice, @dev & @custom` works for NatSpec: https://docs.soliditylang.org/en/latest/natspec-format.html#tags

Will update.

contracts/src/libraries/FROSTParticipantMap.sol
```diff
@@ -177,11 +182,16 @@ library FROSTParticipantMap {
...
      * @param self The storage struct.
      * @param plaintiff The plaintiff's FROST identifier.
      * @param accused The accused's FROST identifier.
-     * @return totalAccusations The total number of accusations against the accused.
+     * @return totalAccusations The total number of unresolved accusations against the accused.
+     * @dev This function is a key part of the DKG's security. If a participant
+     *      detects that another participant is misbehaving (e.g., by providing an
+     *      invalid secret share), they can file a public complaint on-chain.
+     *      A participant cannot file the same complaint twice or complain after
+     *      they have already confirmed their own DKG completion.
```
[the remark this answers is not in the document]
remedcu: Assuming you are referring to the group being `COMPROMISED`, should we specify that in `FROSTCoordinator:keyGenComplain(...)` as a `@dev` comment? As we don't have a `COMPROMISED` state in `FROSTParticipantMap.sol`.

Something like:
```
@dev Only a threshold number of complaints can be raised against an accused, after that, the group will be marked as compromised.
```

contracts/src/FROSTCoordinator.sol
```diff
@@ -502,9 +512,13 @@ contract FROSTCoordinator {
      * @param gid The group ID.
      * @param commitment The nonce commitment Merkle root.
      * @return chunk The chunk index used for this commitment.
-     * @dev The commitment is a Merkle root to 256 nonces that get revealed as
-     *      part of the signing process. This allows signing requests to reveal
-     *      the message immediately while still preventing Wagner's Birthday Attacks.
+     * @dev This function implements the first step of a two-round signing protocol.
+     *      Participants pre-commit to a large set of nonces (256) by submitting
```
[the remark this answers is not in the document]
remedcu: Funnily, AI suggested `eg 1024`, but I changed it to 256 based on the previous comment 😅

contracts/src/libraries/FROSTSignatureShares.sol
```diff
@@ -15,17 +15,22 @@ library FROSTSignatureShares {
     // ============================================================
 
     /**
-     * @notice The main storage struct for tracking signature shares.
-     * @param aggregates Mapping from commitment root to aggregate data.
+     * @notice The main storage struct for tracking aggregated signature shares.
+     * @param aggregates Mapping from the Merkle root of a signing participant set to their aggregate signature data.
```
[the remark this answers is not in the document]
remedcu: Forge has a default line length of 120 characters (`line_length` in `fmt`). But for some reason, it isn't enforced on comments, so I have to adjust them manually this time to < 120 characters. Don't like this as we cannot use the formatter for this, but don't have a good solution either.

PR #110, opened 2025-12-23
Staking External Documentation

remedcu:
This PR adds external documentation (in addition to the inline/Natspec comments in the contract) for users, teams, and auditors to understand the Staking contract. The documentation is written in such a way that even non-devs can understand the process.

### DIFF
--- a/docs/staking.md
+++ b/docs/staking.md
@@ -34,7 +34,7 @@ The contract is non-upgradeable and relies on `Ownable` for admin control. It do
 - `WithdrawalQueue`
   - **Purpose**: Tracks head/tail of a withdrawal list for a staker/validator pair.
   - **Invariants**:
-    - `head == 0 && tail == 0` indicates empty queue.
+    - `head == 0 && tail == 0` indicates empty queue. (`head != 0 <=> tail != 0`)
     - `head` and `tail` are withdrawal IDs; `0` is reserved as null.
 
 - `ConfigProposal`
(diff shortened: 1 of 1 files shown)


docs/staking.md
```diff
@@ -0,0 +1,258 @@
...
+    - `previous` / `next`: Linked list pointers (0 is null).
+  - **Invariants / edge cases**:
+    - Queue is expected to be sorted by `claimableAt` ascending.
+    - Linked list pointers must remain consistent; incorrect manual insertion reverts.
+    - `claimableAt` is stored as `uint128`; truncation only possible at extreme timestamps.
+
+- `WithdrawalQueue`
+  - **Purpose**: Tracks head/tail of a withdrawal list for a staker/validator pair.
+  - **Invariants**:
+    - `head == 0 && tail == 0` indicates empty queue.
```
nlordell: Does that mean that `head != 0 <=> tail != 0`? If so, I would add that as an invariant here as well.

docs/staking.md
```diff
@@ -0,0 +1,258 @@
...
+  - **Purpose**: Increase stake for a validator.
+  - **Invariants**: Validator must be registered; amount must be non-zero.
+  - **Security**: State updates precede `safeTransferFrom` (reverts roll back state).
+
+- `initiateWithdrawal(validator, amount)`
+  - **Purpose**: Create a withdrawal and insert into a sorted queue automatically.
+  - **Edge case**: Traverses from tail; long queues may be gas-intensive.
+  - **Ordering**: Ensures queue remains sorted by `claimableAt`.
+
+- `initiateWithdrawalAtPosition(validator, amount, previousId)`
```
nlordell: Might we worth expanding on the purpose here (it was because we wanted to maintain a sorted list, and to deal with edge cases around changes to the delay). AFAIU, `initiateWithdrawal` should be used 99.99% of the time.

nlordell (approved):
LGTM, just some minor comments.

PR #111, opened 2025-12-29
Certora Setup

remedcu:
This PR just adds the initial Certora setup with a `conf` file and a dummy rule.

This does not include the CI, which will be added as a follow-up later.


certora/conf/Staking.conf
```diff
@@ -0,0 +1,14 @@
...
+    "msg": "Shieldnet Staking",
+    "verify": "StakingHarness:certora/specs/Staking.spec",
+    "files": [
+        "certora/harnesses/StakingHarness.sol"
+    ],
+    "packages": [
+        "@oz=contracts/lib/openzeppelin-contracts/contracts/"
+    ],
+    "loop_iter": "3",
+    "optimistic_loop": true,
```
nlordell: I would avoid adding these prover options until we actually need them.

nlordell (approved):

PR #114, opened 2026-01-05
Validator Checks v1

rmeissner:
- Only allow following delegate calls
  - multisend call only (1.3.0, 1.4.1, 1.5.0)
  - Safe migration (1.4.1, 1.5.0)
  - SignMsg Lib (all versions)
- Limit modules, guards and fallback handlers that can be set
  - Only fallback handlers deployed by Safe team
  - No modules are currently supported
  - Only zero address guards are supported (to disable the guards)
- Inner multisend calls are checked with same rules

### DIFF
--- a/validator/src/consensus/verify/safeTx/checks/basic.test.ts
+++ b/validator/src/consensus/verify/safeTx/checks/basic.test.ts
@@ -221,7 +221,7 @@ describe("basic checks", () => {
 		});
 	});
 
-	describe("buildSupportedSignatureCheck", () => {
+	describe("buildSelectorChecks", () => {
 		it("should throw for data shorter than a selector", async () => {
 			const selectors: Record<string, TransactionCheck> = {};
 			const tx: MetaTransaction = {
--- a/validator/src/consensus/verify/safeTx/checks/config/fallback.test.ts
+++ b/validator/src/consensus/verify/safeTx/checks/config/fallback.test.ts
@@ -20,7 +20,7 @@ describe("fallback", () => {
 			expect(() => check["0xf08a0323"](tx)).toThrow();
 		});
 
-		it("should throw for unknown guard", async () => {
+		it("should throw for unknown fallback handler", async () => {
 			const tx: MetaTransaction = {
 				to: "0xF01888f0677547Ec07cd16c8680e699c96588E6B",
 				value: 0n,
@@ -37,7 +37,7 @@ describe("fallback", () => {
 			);
 		});
 
-		it("allow zero address as module guard (to reset)", async () => {
+		it("allow zero address as fallback handler (to reset)", async () => {
 			const tx: MetaTransaction = {
 				to: "0xF01888f0677547Ec07cd16c8680e699c96588E6B",
 				value: 0n,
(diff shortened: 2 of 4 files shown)


validator/src/consensus/verify/safeTx/checks/config/base.ts
```diff
@@ -0,0 +1,51 @@
...
+	check(tx: MetaTransaction): void {
+		// Config changes are transactions to the account itself
+		if (tx.account !== tx.to) return;
+		if (tx.operation !== 0) throw new Error("Delegatecall not allowed");
+		if (tx.value !== 0n) throw new Error(`Expected no value got ${tx.value}`);
+		// Don't handle transactions with too short data
+		if (size(tx.data) < 4) return;
+		const extractedSelector = tx.data.slice(0, 10);
+		// Don't handle unknown selector
+		if (extractedSelector !== this.selector) return;
```
nlordell: I thing generally speaking, fall-through cases have to come first:


```suggestion
		if (tx.account !== tx.to) return;
		// Don't handle transactions with too short data
		if (size(tx.data) < 4) return;
		const extractedSelector = tx.data.slice(0, 10);
		// Don't handle unknown selector
		if (extractedSelector !== this.selector) return;
		
		if (tx.operation !== 0) throw new Error("Delegatecall not allowed");
		if (tx.value !== 0n) throw new Error(`Expected no value got ${tx.value}`);
```

While for the checks on the Safe itself, it doesn't matter - generally it feels icky to abort for checks that aren't yours. If there would be a case where you would `DELEGATECALL` yourself, then we would have a logic bug.

rmeissner: > generally it feels icky to abort for checks that aren't yours

For me the operation and value check are part of config change checks. Or we add a separate check that is "no self delegatecalls" as a general.

validator/src/consensus/verify/safeTx/checks/config/base.ts
```diff
@@ -0,0 +1,51 @@
...
+		const parsedData = decodeFunctionData<ParseAbiItem<S>[]>({ abi: [this.abi], data: tx.data });
+		this.handleParsedData(parsedData.args);
+	}
+
+	abstract handleParsedData(data: ContractFunctionArgs<ParseAbiItem<S>[], AbiStateMutability>): void;
+}
+
+export function createConfigCheck<S extends string>(
```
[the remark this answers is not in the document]
nlordell: You can always define interfaces as:

```
interface TransactinoCheck {
    (tx: MetaTransaction): void
}
```

I just find functions easier to compose with much less ceremony.

> I would do this in a follow up PR if ok

selbstverständlich, I approved because I think it is fine as is - I definitely don't want to block PRs on style discussions (just wanted to share my preference).

validator/src/consensus/verify/safeTx/checks/config/base.ts
rmeissner: > with something that would allow self delegate calls

That sounds like a very theoretical case, where I can't think of a practical use case right now :sweat_smile: 

>  we should explicitly allow things, instead of allow things by default.

We can use the "allowed selectors" checks to enforce this.

PR #115, opened 2026-01-06
Remove `USER=node` From Dockerfile

nlordell:
User can be specified when starting the container with `--user` if something else is needed. We are running the containers as unprivileged users already (with Quadlets) and the default non-root user makes mounting volumes annoying, and doesn't add much security for us.


nlordell:
For some additional reading, there is an article for RedHat explaining "why use users with rootless podman": https://www.redhat.com/en/blog/rootless-podman-makes-sense. Many of the arguments don't apply to our setup:

> Doing so means that the escaped process could read the secrets in your home directory, like ~/.ssh and ~/.gpg, or other information that you would prefer the container not have access to.

As the unprivileged user starting the Quadlet is already super restricted.

PR #116, opened 2026-01-07
Improve transaction submission

rmeissner:
- Allow to set fee params via env vars
- Enforce nonce manager

### DIFF
--- a/validator/src/validator.ts
+++ b/validator/src/validator.ts
@@ -1,6 +1,7 @@
 import dotenv from "dotenv";
 import type { ChainFees } from "viem";
-import { privateKeyToAccount } from "viem/accounts";
+import { createNonceManager, privateKeyToAccount } from "viem/accounts";
+import { jsonRpc } from "viem/nonce";
 import { z } from "zod";
 import { createValidatorService } from "./service/service.js";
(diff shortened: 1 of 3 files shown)


validator/src/validator.ts
```diff
@@ -32,7 +34,16 @@ const config: ProtocolConfig = {
 	blocksPerEpoch: validatorConfig.BLOCKS_PER_EPOCH,
 };
 
-const account = privateKeyToAccount(validatorConfig.PRIVATE_KEY);
+const fees: ChainFees = {
+	// Use a higher default multiplier to ensure transaction inclusion
+	baseFeeMultiplier: validatorConfig.BASE_FEE_MULTIPLIER ?? 2,
+	// Allow to set higher default priority fee to ensure transaction inclusion
+	maxPriorityFeePerGas: validatorConfig.PRIORITY_FEE_PER_GAS,
```
nlordell: in my experience, this should be dynamic based on `eth_getFeeHistory` with a cap. That is a more involved change for another time though.

rmeissner: yes, I wrote in the initiative that we accept this for now. I you don't set it, then it will use the default mechanism of viem. it would be more a temporary overwrite if necessary.

PR #119, opened 2026-01-08
Initial structure for public docs + gh page deployment

tschubotz:
- Adds the initial public documentation that we'll need for Shieldnet.
- Uses docusaurus.
- Basic placeholder structure added - to be iterated upon.
- Uses placeholder images/logos/icons and general design.
- Still contains TODOs. Minimum tasks to clear before release [are tracked here](https://www.notion.so/safefoundation/Shieldnet-Public-docs-wiki-2c6a8a34f3b88034b41bf5b941168bff?source=copy_link#2c6a8a34f3b8802995a5e5849a863a79). 
- Adds GH action to deploy to GH pages


docs/package.json
```diff
@@ -0,0 +1,49 @@
...
+		"start": "docusaurus start",
+		"build": "docusaurus build",
+		"swizzle": "docusaurus swizzle",
+		"deploy": "docusaurus deploy",
+		"clear": "docusaurus clear",
+		"serve": "docusaurus serve",
+		"write-translations": "docusaurus write-translations",
+		"write-heading-ids": "docusaurus write-heading-ids",
+		"typecheck": "tsc",
+		"check": "biome check && tsc",
```
rmeissner: you need to provide a "test" script (and soon also "coverage"). We could just do an `exit 0`

docs/package.json
```diff
@@ -0,0 +1,49 @@
...
+		"fix": "biome check --write"
+	},
+	"dependencies": {
+		"@docusaurus/core": "3.9.2",
+		"@docusaurus/preset-classic": "3.9.2",
+		"@mdx-js/react": "^3.0.0",
+		"clsx": "^2.0.0",
+		"prism-react-renderer": "^2.3.0",
+		"react": "^19.0.0",
```
rmeissner: nit: In the explorer we use `19.1.0` as a react version. Probably makes sense that we align these versions

docs/README.md
```diff
@@ -0,0 +1,41 @@
+# Website
+
+This website is built using [Docusaurus](https://docusaurus.io/), a modern static website generator.
+
+## Installation
+
+```bash
+yarn
```
rmeissner: nit: we are using npm

.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,50 @@
+name: Deploy Docs to GitHub Pages
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - "docs/**"
+      - ".github/workflows/deploy-pages.yml"
+  workflow_dispatch:
```
rmeissner: Potentially it would make sense to limit the dispatch to "main", to do this you will have to add a condition to `build` job: `if: github.ref == 'refs/heads/main'`

rmeissner: Alternatively we limit the environment in the github repo settings, actually this is cleaner and more stable

rmeissner: just checked this is already the case, so nothing to do, this workflow will only work on `main`

docs/package.json
```diff
@@ -0,0 +1,51 @@
+{
+	"name": "shieldnet-docs",
+	"version": "0.0.0",
+	"private": true,
+	"scripts": {
+		"docusaurus": "docusaurus",
+		"start": "docusaurus start",
+		"build": "docusaurus build",
+		"swizzle": "docusaurus swizzle",
+		"deploy": "docusaurus deploy",
```
rmeissner: nit: potentially we should remove this, as we deploy via github actions

docs/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Build
+
+```bash
+yarn build
+```
+
+This command generates static content into the `build` directory and can be served using any static contents hosting service.
+
+## Deployment
```
rmeissner: nit: would make sense to update this to our process

docs/package.json
tschubotz: removed.

PR #128, opened 2026-01-09
Add MacOS devcontainer & update devcontainer node version

tschubotz:
- Update image to node v24
- The current devcontainer doesn't work with MacOS
- Adding a workable one for MacOS that can be selected.
- Linux stays default.
- Added comments to linux devcontainer.
- Added info that the linux devcontainer is specific to Fedora and might need an update.

### DIFF
--- a/.devcontainer/macos/devcontainer.json
+++ b/.devcontainer/macos/devcontainer.json
@@ -4,7 +4,7 @@
 
 {
   "name": "Shieldnet (MacOS)",
-  "image": "mcr.microsoft.com/devcontainers/typescript-node:dev-24-bookworm"
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:dev-24-bookworm",
 
   // User that VS Code connects as *inside* the container.
   "remoteUser": "node"


.devcontainer/macos/devcontainer.json
```diff
@@ -0,0 +1,18 @@
+// Separate devcontainer file for MacOS.
+// The one in ../devcontainer.json is specific to Linux/Fedora and some args don't work.
+// Devcontainers don't support `extends` yet, hence we need to keep them in sync manually.
+
+{
+  "name": "Shieldnet (MacOS)",
+	"image": "mcr.microsoft.com/devcontainers/typescript-node:1-22-bookworm",
```
rmeissner: Maybe it makes sense to update both to `typescript-node:24-bookworm` with this PR (or we do it in a separate one)

.devcontainer/macos/devcontainer.json
```diff
@@ -0,0 +1,18 @@
...
+
+	"mounts": [
+  ],
+
+  "remoteEnv": {
+      "SSH_AUTH_SOCK": "/tmp/ssh-agent.sock"
```
rmeissner: Does this work without the mount? I mount this so that I can use my host ssh agent in the dev container.

.devcontainer/macos/devcontainer.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+
+  "remoteUser": "node",
+	"updateRemoteUserUID": true
```
rmeissner: the indent here seems off

rmeissner (approved):

.devcontainer/macos/devcontainer.json
tschubotz: will do

.devcontainer/macos/devcontainer.json
tschubotz: ya, seems I wasn't think when copying here. will remove

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
nlordell: According to this thread, it looks like it isn't entirely solved: https://github.com/orgs/devcontainers/discussions/60

PR #129, opened 2026-01-09
Docs: Add introduction page.

tschubotz:
- This PR adds the intro page to the docs including links for more details on respective other docs pages.
- Also adds placeholder pages for integrators

image

image

image

image

### DIFF
--- a/docs/docs/introduction.md
+++ b/docs/docs/introduction.md
@@ -75,7 +75,7 @@ Shieldnet v1 focuses on:
 
 - **Safe transactions on EVM chains**
 - **Permissioned validator set**
-- **Pre-defined security policies**
+- **Pre-defined security security checks**
 - **Onchain enforcement via Safe guards**
 
 This is a deliberate starting point. Transaction security is complex, and Shieldnet prioritizes correctness, reliability, and measurable guarantees in its initial deployment.


docs/docs/introduction.md
```diff
@@ -1,8 +1,85 @@
...
 
-TODO
+Shieldnet is a protocol for **onchain transaction security enforcement**.
 
-- What Shieldnet is, what problem it solves, and who it is for.
-- Value prop
-- Explicit non-goals?
-- Needs direct links such as "Want to learn about the protocol - click here", "Want to stake? - here"
\ No newline at end of file
+It acts as a **last line of defense** against malicious or high-risk transactions by ensuring that transactions are **validated before execution**, rather than merely displaying non-binding warnings.
```
nlordell:
```suggestion
It acts as a **first line of defense** against malicious or high-risk transactions by ensuring that transactions are **validated before execution**, rather than merely displaying non-binding warnings.
```

tschubotz: I'd argue that the first line of defense would be tx checks that just have a UI indicator of whether or not a tx is safe. However even if the UI is compromised and a user signed a tx, we argue that Shieldnet could still be a safety net once onchain enforcement of the check is enabled. Thereby it acts as _last_ line of defense rather than the first line.

What did you have in mind, what was your mental model?

docs/docs/introduction.md
```diff
@@ -1,8 +1,85 @@
...
+Even simple onchain actions (e.g. approvals, signer changes) can be embedded in deceptive transaction flows.
+
+## Goals and value proposition
+
+Shieldnet is designed to:
+
+- **Prevent malicious transactions from executing**, not just warn about them
+- Make transaction security **accessible to all users**, not only advanced or institutional ones
+- Provide **additive security** that complements existing wallets and protocols
+- Create **public, accountable security guarantees**, enforced onchain
```
ennisjo: "Create **public, accountable security guarantees**, enforced onchain"

Is this really a goal? I think this could be rephrased or split into another goal. You could rephrase to say we want to enforce things onchain, as its the most resilient/robust way. Or another goal is to make the checks/guarantees onchain, so checkers are publicly accountable/auditable for their reviews.

ennisjo (changes requested):
See line 30, and whether this goal needs to be revised.

PR #138, opened 2026-01-15
Docs: Rename Shieldnet -> Safenet

tschubotz:
- Rename Shieldnet to Safenet in all public docs following the branding decision.
- urls and mentions of the gh repo aren't touched since they remain shieldnet for now.


nlordell (approved):
LGTM - we should probably rename the repository and the `package.json` package names as well.

PR #146, opened 2026-01-19
Use block watcher to trigger pending transaction checks

rmeissner:
Assuming that there is a strict relation between nonce and action we can rely on the transaction count to check for execution. This has the advantage that we only need to do one rpc call to check all pending transactions.

Changes:
- Compare nonce of pending txs against current transaction count and mark as pending of nonce is smaller
- Add possibility to disable polling
- Use `block_new` event to trigger pending check


rmeissner:
I added it to https://github.com/safe-research/shieldnet/issues/86, most likely will fix it later this week.

PR #152, opened 2026-01-20
Use block number for pending txs check 

rmeissner:
- Add migration sql that needs to be applied to exisiting databases before using this version
- When submitting txs use `null` as a submission block
- Block number has to be provided to the check function
  - Will set the submission block for pending transactions
  - Will retry transactions based on elapsed blocks


nlordell (approved):
LGTM!

PR #154, opened 2026-01-21
More Robust Share Index Computation

nlordell:
This PR fixes an issue in the share index computation where it assumes that the FROST identifiers are sequential starting from 1. While this is true of the spec, in the current implementation the FROST identifiers (i.e. participant IDs in the code) are stable across different groups and are determined by the index in the total validator set (and not just the participating validators in a FROST group, which may be smaller in case a participant is removed from the set for an epoch - in case it is offline for example).

The fix is very simple: compute the share index from the actual participants list that we read from the database. This makes it so the client has no assumptions on the "structure" of the participant ID but relies on the data that its storage implementation provides. This is verified by modifying the unit test so that the FROST identifier is just the address of the participant, and modifying the integration test to stop the middle participant (creating a "hole" in the identifiers and making them no longer sequential).

The added bonus for this is that, for a given default participant set, the FROST identifier (i.e. participant ID) is stable across epochs. This is nice for debugging as identifier `N` will always be the same validator.

### DIFF
--- a/spec/eth.ml
+++ b/spec/eth.ml
@@ -7,7 +7,8 @@ end
 
 module Abi = struct
   type t =
-    [ `Uint32 of int
+    [ `Uint16 of int
+    | `Uint32 of int
     | `Uint64 of int
     | `Uint256 of int
     | `Address of Address.t
--- a/spec/validator.ml
+++ b/spec/validator.ml
@@ -172,27 +172,33 @@ struct
 
   let participant_identifier participants address =
     (* We explicitely define the participant identifier to be its index in the
-       sorted list of active participants starting at 1. Note that OCaml `Set`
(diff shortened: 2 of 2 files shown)


validator/src/consensus/integration.test.ts
```diff
@@ -194,8 +194,8 @@ describe("integration", () => {
 			onLogs: () => {
 				// Only react to first completed keygen
 				unsubscribe();
-				testLogger.notice("Stop client with index 2, keygen will timeout");
-				clients[2].service.stop();
+				testLogger.notice("Stop client with index 1, keygen will timeout");
```
rmeissner: did you check if the test failed now without your adjustments to the client?

rmeissner (approved):

PR #159, opened 2026-01-27
Improve tx resubmission

rmeissner:
- Allow to configure blocks before resubmit via `BLOCKS_BEFORE_RESUBMIT`

### DIFF
--- a/validator/src/consensus/protocol/onchain.ts
+++ b/validator/src/consensus/protocol/onchain.ts
@@ -48,7 +48,6 @@ export interface TransactionStorage {
 }
 
 export class GasFeeEstimator {
-	#currentBlockNumber = 0n;
 	#cachedPrices: Promise<FeeValues> | null = null;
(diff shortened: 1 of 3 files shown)


validator/src/consensus/protocol/onchain.ts
```diff
@@ -31,20 +33,51 @@ import type {
...
+	constructor(client: PublicClient) {
+		this.#client = client;
+	}
+
+	invalidate(blockNumber: bigint) {
+		if (blockNumber > this.#currentBlockNumber) {
+			this.#cachedPrices = null;
+			this.#currentBlockNumber = blockNumber;
+		}
+	}
```
nlordell: I would actually just `invalidate` on every block, and not track the current block number at all. In particular, reorgs change fee parameters (as they can affect how full the block is and the current base fee).


```suggestion
	invalidate() {
		this.#cachedPrices = null;
	}
```

PR #160, opened 2026-01-30
Enforce minimum group size

rmeissner:
So far we only enforce the signing threshold and the minimum group size required by FROST (2). We have additional requirement related to the security of our network which require a minimum group size based on the default participant set to avoid the "Shrinking Quorum Attack".

Details are outlined in the code

### DIFF
--- a/validator/src/machine/keygen/genesis.test.ts
+++ b/validator/src/machine/keygen/genesis.test.ts
@@ -41,6 +41,10 @@ const MACHINE_CONFIG: MachineConfig = {
 			id: 3n,
 			address: entryPoint08Address,
 		},
+		{
+			id: 4n,
+			address: entryPoint08Address,
+		},
 	],
 	genesisSalt: zeroHash,
 	keyGenTimeout: 25n,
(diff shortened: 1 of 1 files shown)


validator/src/machine/keygen/genesis.test.ts
```diff
@@ -55,7 +55,7 @@ const EVENT: KeyGenEvent = {
 	gid: "0x06cb03baac74421225341827941e88d9547e5459c4b3715c0000000000000000",
 	participants: "0xc41a668cab547c9e2cfd2975f4b6130877c8500c21820daffba14b24c40c742e",
 	count: 3,
-	threshold: 2,
+	threshold: 3,
```
nlordell: Shouldn't this change be reverted? Threshold is half again.

nlordell (approved):

validator/src/machine/keygen/genesis.test.ts
rmeissner: actually count should be 4 for the proper example

PR #169, opened 2026-02-09
Rework Consensus Events

nlordell:
This PR rewords the consensus events so they are more useful to how we expect users to interact with the contracts. Namely:

1. We now emit transaction proposal events with `chainId` and `safe` so you can scan all proposed Safe transactions for a particular account or all accounts with the same address.
2. We include attestations (i.e. FROST signatures) with some events, making it easier for clients to build guard transactions. In particular, `EpochStaged` events now also include the signature data, in order to allow clients to build a "signing chain" in order to update the group key in guard contracts. Similarly, `TransactionAttested` events include the attestation data.
3. We store _all_ attestations in storage, and provide view functions to read them. This allows accounting (for example, when computing validator rewards) to query all signature IDs for all attestations in order to derive participation values.
4. We add the selection `root` to the signature events to make it easier to determine participation for a completed attestation

### DIFF
--- a/contracts/src/FROSTCoordinator.sol
+++ b/contracts/src/FROSTCoordinator.sol
@@ -97,7 +97,7 @@ contract FROSTCoordinator {
     /**
      * @notice Secret share data for key generation.
      * @custom:param y The participant public key share.
-     * @custom:param f The polynomial coefficients encrypted for participants.
+     * @custom:param f The polynomial evaluations encrypted for participants.
      */
     struct KeyGenSecretShare {
         Secp256k1.Point y;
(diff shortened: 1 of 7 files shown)


validator/src/consensus/integration.test.ts
```diff
@@ -142,6 +147,8 @@ describe("integration", () => {
 			const watcherConfig: WatcherConfig = {
 				maxReorgDepth: 1,
 				blockTimeOverride: blockTime,
+				blockPropagationDelay: Math.floor(blockTime / 5),
+				blockRetryDelays: [Math.floor(blockTime / 10), Math.floor(blockTime / 20), Math.floor(blockTime / 20)],
```
nlordell: Makes the test more robust. I had a case where the first epoch rotation was aborted because we got to block 20 before staging was complete.

rmeissner (approved):

PR #174, opened 2026-02-11
Replace waiting_for_rollover with waiting_for_genesis and epoch_staged

rmeissner:
Motivation:

- `waiting_for_rollover` was a state that should never exist without `stagedEpoch` > 0. To better represent this align the current validator with the states of the spec
- The specs don't consider the genesis state (`waiting_for_rollover` and `stagedEpoch` == 0), therefore a dedicated state `waiting_for_genesis` is introduced

Notable changes:
- Replace `waiting_for_rollover` with `waiting_for_genesis` and `epoch_staged`
- Remove `stagedEpoch` from consensus state
- Remove `abortKeyGen` method, as this is not handled with the rollover check. 
- The rollover check is performed at the beginning of each transition, as it should happen first with every new block 

Notes on genesis key gen:
- While in genesis key gen (`nextEpoch` == 0) the key gen is **never** aborted
- It is still possible that timeouts and wrong information during genesis key gen reduce the participant set
- This also changes the group id, making the setup of the initial group impossible (even if a threshold of participants is active)!
- **TL;DR: For gensis key gen it must be ensured that all validators participate**

### DIFF
--- a/validator/src/machine/consensus/rollover.ts
+++ b/validator/src/machine/consensus/rollover.ts
@@ -20,12 +20,16 @@ export const checkEpochRollover = (
 		return {};
 	}
 
+	if (currentState.id !== "epoch_staged" && currentState.nextEpoch === 0n) {
+		// Rollover should not happen while in genesis keygen.
+		return {};
+	}
+
 	// This check applies to all states
 	// When staged or skipped then keygen should be started for next epoch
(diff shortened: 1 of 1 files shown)


nlordell (approved):
Very nice change! Simplifies the state machine logic!

PR #192, opened 2026-02-18
Add global backoff for action submission

rmeissner:
Currently actions where aggressively retried every second without a backoff, and new actions immediately triggered a new retry. To avoid kicking the service when it is already on the ground a global backoff for the action submission is introduced.

Notable changes:
- Introduce global backoff that increases on each retry by 1s with a maximum of 5s.
- Don't submit actions when backing off

### DIFF
--- a/validator/src/consensus/protocol/base.test.ts
+++ b/validator/src/consensus/protocol/base.test.ts
@@ -165,11 +165,14 @@ describe("BaseProtocol", () => {
 		// Called 3 times: 2 times in the test, 1 time in the implementation
 		expect(peekSpy).toBeCalledTimes(3);
 		// Check if retry was scheduled via setTimeout
-		for (let i = 1; i <= 7; i++) {
+		for (const [times, delay] of [
+			1000, 2000, 3000, 4000, 5000,
+			// Does not increase past maximum!
+			5000, 5000,
+		].map((delay, i) => [i + 1, delay])) {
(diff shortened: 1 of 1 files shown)


validator/src/consensus/protocol/base.test.ts
```diff
@@ -97,6 +97,187 @@ describe("BaseProtocol", () => {
...
+		expect(pushSpy).toBeCalledTimes(1);
+		expect(pushSpy).toBeCalledWith(actionWithTimeout);
+		// Called 3 times: 2 times in the test, 1 time in the implementation
+		expect(peekSpy).toBeCalledTimes(3);
+		// Check if retry was scheduled via setTimeout
+		for (let i = 1; i <= 7; i++) {
+			await vi.waitFor(() => {
+				expect(timeoutSpy).toBeCalledTimes(i);
+			});
+			const delay = Math.min(i * 1000, 5000);
```
nlordell: nit: This duplicates the calculation of the function under test, and could add more testing value IMO:

```suggestion
		for (const [times, delay] of [
		   1000,
		   2000,
		   3000,
		   4000,
		   5000,
		   // Does not increase past maximum!
		   5000,
		   5000,
		].map((delay, i) => [i + 1, delay] as const)) {
			await vi.waitFor(() => {
				expect(timeoutSpy).toBeCalledTimes(times);
			});
```

PR #208, opened 2026-02-24
Expose Additional FROST Coordinator Info

nlordell:
This PR exposes additional information from the FROST coordinator contract in order to provide a richer explorer experience. In particular, it exposes:

1. The group parameters (count, threshold, paraticipants root hash)
2. The number of signatures the group has start (internally, this is tracked as the `sequence`, but the API was explicitely chosen to not call it that as part of the view function in order to not give the false sense that you can query what the "expected sequence" is, and have clients make false assumptions that break under race conditions)
3. The participant addresses once they are revealed as part of the group commitment
4. Prevent reading the group key before the group is ready. This makes it explicit that groups "under construction" should not be used. As a bonus side-effect, it prevents `Consensus` to stage an epoch with a group that is not ready.

Note that we used to have a `GroupParameters` struct, but it was renamed to `GroupState`, as it keeps group internal state such as pending counters and the current status. I kept "parameters" to mean the group parameterization (i.e. count, threshold, participants which are specified at `keyGen` time) to avoid confusion (in fact, we already had `InvalidGroupParameters` error to indicate an issue with group parameterization and not internal `GroupState`).

### DIFF
--- a/validator/src/__tests__/data/protocol.ts
+++ b/validator/src/__tests__/data/protocol.ts
@@ -336,11 +336,12 @@ export const TEST_EVENTS: [ProtocolLog | null, StateTransition][] = [
 		{
 			blockNumber: 111n,
 			logIndex: 0,
-			// KeyGenCommitted(bytes32 indexed gid, uint256 identifier, ((uint256 x, uint256 y) q, (uint256 x, uint256 y)[] c, (uint256 x, uint256 y) r, uint256 mu) commitment, bool committed)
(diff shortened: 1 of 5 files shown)


contracts/test/FROSTCoordinator.t.sol
```diff
@@ -224,8 +212,6 @@ contract FROSTCoordinatorTest is Test {
                 s[identifier] = addmod(s[identifier], f[identifier][l], Secp256k1.N);
             }
         }
-        a = new uint256[][](0);
-        f = new uint256[][](0);
```
nlordell: This was symbolic to show that you could delete the coefficients. Not needed for the test, and removed so that we can do the final assertion about the group key.

PR #211, opened 2026-02-25
Introduce Validator Handbook

nlordell:
This PR adds a short validator handbook that should help give a sense of "what you need to run a validator". This can be shared to the teams we are currently reaching out to so they have a better understanding of the current costs associated with running one.

### DIFF
--- a/docs/validator-handbook.md
+++ b/docs/validator-handbook.md
@@ -32,14 +32,20 @@ To run a validator, you need a reliable Ethereum RPC node that can accommodate a
 
 The validator node writes JSON-formatted logs to standard output. It also exposes Prometheus metrics on `:3555` by default, which can be scraped over HTTP.
 
-### Private Keys
+### Secrets
 
-#### `secp256k1` Keys
+#### `secp256k1` Validator Key
 
-Each validator must be provisioned with a `secp256k1` private key. This key is used to authenticate the validator onchain for participation in Safenet Beta. It must be funded with sufficient gas for the EVM transactions required for onchain consensus-related communication. The exact amount varies by chain, but you can expect the account to consume roughly 1.000.000.000 gas per day under peak load.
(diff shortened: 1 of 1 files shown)


docs/validator-handbook.md
```diff
@@ -0,0 +1,88 @@
...
+
+#### Logging and Metrics
+
+The validator node writes JSON-formatted logs to standard output. It also exposes Prometheus metrics on `:3555` by default, which can be scraped over HTTP.
+
+### Private Keys
+
+#### `secp256k1` Keys
+
+Each validator must be provisioned with a `secp256k1` private key. This key is used to authenticate the validator onchain for participation in Safenet. It must be funded with sufficient gas for the EVM transactions required for onchain consensus-related communication. The exact amount varies by chain, but you can expect the account to consume roughly 1.000.000.000 gas per day under peak load.
```
rmeissner: I would separate the gas consumption into a separate section (`####`) and maybe have some napkin math related to Gnosis Chain (inkl links where to get the latest gas prices). 

Why I would do this, is because I think that the validators in the best case just would like a number (i.e. fund it with X$ a month, that should be sufficient for now)

docs/validator-handbook.md
```diff
@@ -0,0 +1,88 @@
+# Safenet Validator Handbook
+
+This document provides a brief guide to operating a Safenet validator.
```
tschubotz:
```suggestion
This document provides a brief guide to operating a Safenet Beta validator.
```

tschubotz (approved):
I think this doc is great and exactly what the devops folks of prospective validators want and need. 

I've made only stylistic suggestions. It needs to be really clear that this is a Beta. Both for the validators, in case stuff goes wrong as well as any 3rd party reviewing any docs to judge how production ready everything is.

PR #214, opened 2026-02-25
Instrument RPC Requests

nlordell:
This PR just adds some basic instrumentation to our Ethereum RPC requests. This allows us to better track what RPC methods are being called, in what quantity, and which ones are failing. This allows us to rely on internal tracking instead of RPC providers giving us a nice dashboard.

### DIFF
--- a/validator/src/utils/metrics/transport.ts
+++ b/validator/src/utils/metrics/transport.ts
@@ -1,14 +1,14 @@
 import type { Transport } from "viem";
 import type { Metrics } from "./index.js";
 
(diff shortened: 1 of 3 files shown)


validator/src/utils/metrics/transport.ts
```diff
@@ -0,0 +1,21 @@
+import type { Transport } from "viem";
+import type { Metrics } from "./index.js";
+
+export const withMetrics = <T extends Transport>(transport: T, metrics: Metrics): T => {
+	return ((options) => {
```
rmeissner: Is this the recommended way for intercepting request with viem?

PR #222, opened 2026-02-27
Allow to skip genesis

rmeissner:
So far it was not possible to skip genesis key gen. This mean if it was missed the validator got stucked in waiting for genesis. This PR makes it possible to skip genesis and join the next possible key gen. This is achieved by moving into an "epoch_skipped" state as soon as the first block is received.

### DIFF
--- a/validator/src/machine/consensus/rollover.ts
+++ b/validator/src/machine/consensus/rollover.ts
@@ -22,7 +22,7 @@ export const checkEpochRollover = (
 
 	if (currentState.id === "skip_genesis") {
 		// The state of the current key gen (for the next epoch) is unknown.
-		// To avoid unexpected behvior, skip the current key gen and wait for the next.
+		// To avoid unexpected behavior, skip the current key gen and wait for the next.
 		return {
 			rollover: { id: "epoch_skipped", nextEpoch: currentEpoch + 1n },
 		};


validator/src/types/schemas.ts
```diff
@@ -52,6 +52,8 @@ const genesisSaltSchema = z.preprocess((val) => {
 const BLOCKTIME_IN_SECONDS = 5n; // value assumed for gnosis chain
 const BLOCKS_PER_EPOCH = (24n * 60n * 60n) / BLOCKTIME_IN_SECONDS; // ~ blocks for 1 day
 
+const strictBoolSchema = z.enum(["true", "false", "0", "1"]).transform((v) => v === "true" || v === "1");
```
rmeissner: the default `z.coerce.boolean()` is equivalent to `Boolean(value)` which for `"false"` returns `true` ... therefore this strict bool schema was introduced

validator/src/machine/consensus/rollover.ts
```diff
@@ -21,9 +21,8 @@ export const checkEpochRollover = (
 	}
 
 	if (currentState.id === "skip_genesis") {
-		// If genesis was skipped then prepare to participate in the next keygen
-		// Current key gen is for the next epoch, which is skipped.
-		// Next key gen is for the epoch after
+		// The state of the current key gen (for the next epoch) is unknown.
+		// To avoid unexpected behvior, skip the current key gen and wait for the next.
```
remedcu:
```suggestion
		// To avoid unexpected behavior, skip the current key gen and wait for the next.
```

PR #224, opened 2026-02-27
Add CoW Extensible Fallback Handler

nlordell:
We used it ;).

The address of the contract can be verified from the CoW repository: https://github.com/cowprotocol/composable-cow?tab=readme-ov-file#deployed-contracts

---

Joking aside some other handlers/modules/adapters we should consider adding as safe:

- 4337 adapter
- 7579 adapter
- Candide recovery module
- allowance module v1

If we agree, I would add them here as well.

> [!TIP]
> An _adapter_ is a contract that integrates with a Safe by being configured as both a fallback handler and a module.


nlordell:
> For reference to reviewers, this is where I checked

Thanks, that was missing from the PR description :see_no_evil: - added it now.

rmeissner:
Contracts that are part of the Safe bug bounty program should be no-brainers (4337 and allowance module). Recovery is also part of the safe-modules repo.

One note tho, 4337 and 7579 both allow arbitrary delegate calls. Currently we don't have a way to attest module transactions, so not sure how we want to handle this.

TL;DR: I would for now only add recovery and allowance.

rmeissner (approved):

PR #230, opened 2026-03-03
Validator `env` example update

remedcu:
This PR adds more reasons for not just the what, but also the why, for each variable, along with the possible and default values for each environment variable used by validators.

P.S. Minor cleanup of `.gitignore` also added.

### DIFF
--- a/validator/.env.sample
+++ b/validator/.env.sample
@@ -17,7 +17,6 @@
 #   "silent" — No logs will be produced.
 #
 # Default: "notice"
-# More info: https://github.com/winstonjs/winston
 LOG_LEVEL=
 
 # Metrics Port [Optional]
@@ -72,11 +71,10 @@ STORAGE_BACKUP=
 # HTTP (`http://`, `https://`) and WebSocket (`wss://`) are supported.
 # Transport is auto-detected: URLs starting with "wss" use WebSocket, others use HTTP.
 #
-# Choose a reliable, low-latency provider — the validator's responsiveness depends
-# entirely on this connection.
+# Choose a reliable provider — the validator's responsiveness depends entirely on
+# this connection.
 #
 # Values:  Any valid URL (e.g. "https://rpc.gnosis.io", "wss://rpc.gnosis.io")
-# Default: None (required — validator will fail to start without this).
 RPC_URL=
 
 # Private Key [Required]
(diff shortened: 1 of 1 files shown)


validator/.env.sample
```diff
@@ -1,46 +1,315 @@
...
+#   "error"  — An operation failed and functionality was impacted.
+#   "warn"   — Something unexpected happened, but the validator can recover.
+#   "notice" — Normal but significant events (e.g. epoch transitions, key gen).
+#   "info"   — General operational information.
+#   "debug"  — Detailed internal state, useful for diagnosing issues.
+#   "silly"  — Extremely detailed, typically only for development.
+#   "silent" — No logs will be produced.
+#
+# Default: "notice"
+# More info: https://github.com/winstonjs/winston
```
nlordell: I think our levels are a bit custom, so I wouldn't point to this.

validator/.env.sample
```diff
@@ -1,46 +1,315 @@
...
+
+# RPC URL [Required]
+#
+# URL of the JSON-RPC node to connect to. Used for all on-chain interactions:
+# reading blocks, querying logs, estimating fees, and submitting transactions.
+#
+# HTTP (`http://`, `https://`) and WebSocket (`wss://`) are supported.
+# Transport is auto-detected: URLs starting with "wss" use WebSocket, others use HTTP.
+#
+# Choose a reliable, low-latency provider — the validator's responsiveness depends
```
nlordell: Latency isn't a problem IMO, even a high latency validator will let you process blocks within the 5-12s block time.

💯 on the reliability part though.

validator/.env.sample
```diff
@@ -1,46 +1,315 @@
...
+# Values:  Any positive integer
+# Default: 1 — resubmits on every new block if unconfirmed.
+BLOCKS_BEFORE_RESUBMIT=
+
+# Base Fee Multiplier [Optional]
+#
+# Multiplier for the base fee when estimating maxFeePerGas:
+#   maxFeePerGas = baseFee × baseFeeMultiplier + maxPriorityFeePerGas
+#
+# Higher = willing to pay more during fee spikes = better inclusion reliability.
```
nlordell: AFAIU, this is _slightly_ inaccurate, you will always pay the same base fee regardless of this value. It just accelerates how fast it increases your gas price on resubmissions. Higher values speed up inclusion (fewer iterations to get to the market price for gas in case of a spike) with the downside that you need more idle gas funds in the validator account.

PR #307, opened 2026-03-17
Block Indexing Delay

nlordell:
It turns out that some nodes have a delay between seeing a block and indexing things like receipts and logs. During this time, `eth_getLogs` will return incorrect data (`[]` despite there being logs).

This PR introduces a work around is to have a configurable additional delay for waiting for the RPC to finish indexing a block. This is a fairly un-invasive but heuristic approach to the problem. Unfortunately, it doesn't guarantee a solution, but wanted to suggest it as a pragmatic approach to the issues we are currently seeing with our RPC provider.


nlordell:
Closing in favour of #309

PR #338, opened 2026-04-07
Additional Staking Configuration Context

nlordell:
This PR adds some additional context on how to correctly configure and provide stake for Safenet Beta.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -57,6 +57,6 @@ See validator [.env.sample](../validator/.env.sample) for reference and addition
 
 #### Staking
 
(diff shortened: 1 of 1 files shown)


ennisjo (approved):

rmeissner (approved):

tschubotz (approved):
Generally looks good. 

Gemini found 2 typos - would be good to fix them.

I would add a note on the importance/implications of the staker address eg.

> Validators earn a fixed commission on delegated stake. Commission will only be earned if the `STAKER_ADDRESS` is set AND the minimum stake has been put up by the `STAKER_ADDRESS`. [More infos here](https://docs.safefoundation.org/safenet/staking/rewards).

PR #344, opened 2026-04-21
Enable hash-based routing in Plausible Analytics

tschubotz:
## Summary
Enable hash-based routing support in the Plausible Analytics integration to correctly track pageviews when navigating within the explorer's single-page application.

## Changes
- Updated `Analytics.tsx` to pass `hashBasedRouting: true` to the Plausible `init()` call
- Updated corresponding test cases in `Analytics.test.tsx` to verify the new configuration option is passed

## Implementation Details
The explorer uses hash-based routing for client-side navigation. Without this setting, Plausible only records a single pageview on initial page load and misses subsequent in-app navigations triggered by hash changes. By enabling `hashBasedRouting: true`, the Plausible tracker will fire a pageview event on every `hashchange` event, ensuring all navigations are properly tracked.

This is a minimal, single-phase change that leverages the existing `hashBasedRouting` flag exposed by the `@plausible-analytics/tracker` package. No new dependencies, environment variables, or components are required.

https://claude.ai/code/session_01BVCsL2RPNajTNe2HPfxXQw


nlordell (approved):

features/2026_04_21_plausible_hash_based_routing.md
```diff
@@ -0,0 +1,43 @@
+# Feature Proposal: Plausible Hash-Based Routing
```
rmeissner: We should remove this file in the same PR as it was fully implemented. 

Note for the future: For single PR changes (or smaller changes) we don't require the feature doc. It is a nice to have if it is accessible withing the commit history if we want to find something, but actually the PR description is a better place for the most important decisions (as it will be converted to a commit message).

rmeissner (approved):

features/2026_04_21_plausible_hash_based_routing.md
tschubotz: ok will remove

PR #348, opened 2026-04-22
Safenet Guard

remedcu:
This PR introduces the Safenet Guard contract, which uses the FROST signature scheme to verify attestations produced on the Consensus-deployed chain. This is the key ingredient for the onchain enforcement of the Safenet checks.

### Some opinionated choices and questions about the Safenet Guard:

1. For now, module-based transaction attestation is not supported. Could be supported in the next version of the guard.
2. Historic epoch information history is stored rather than just a few epoch information to support one-step attestation verification and execution of the Safe Tx (some transactions might be attested in one epoch, but take a longer time to get all the signatures from the owner, making the epoch rollover multiple times during that time). This could result in old malicious epoch participants attesting to transactions even after getting removed from the current epoch. In the current beta version, we have a permissioned set of validators, so expecting to avoid this kind of behaviour. A possible solution is to limit the attestation to the last n epochs (based on either a count or absolute epoch numbers).


contracts/src/SafenetGuard.sol
```diff
@@ -0,0 +1,698 @@
...
+     *      attestation exists, call `cancelModuleAttestation` first to clear it.
+     *      Module transactions must be hashed with all gas and nonce fields set to zero, matching
+     *      the convention used by `checkModuleTransaction` and `cancelModuleAttestation`.
+     * @param safeTxHash EIP-712 hash of the Safe transaction. For module transactions this must
+     *                   be computed with zeroed gas and nonce fields.
+     * @param epoch      Epoch in which the attestation was produced. Must be either the current
+     *                   or the previous epoch; any other value reverts with `InvalidEpoch`.
+     * @param signature  FROST threshold signature over the transaction proposal message.
+     */
+    function submitAttestation(bytes32 safeTxHash, uint64 epoch, FROST.Signature calldata signature) external {
```
rmeissner: why did you decide to attest with a method call instead by extending the signatures bytes?

If it is related to module support, I would have left used modules for now

contracts/src/SafenetGuard.sol
```diff
@@ -0,0 +1,698 @@
...
+     */
+    EpochState private _currentEpoch;
+
+    /**
+     * @notice The epoch immediately preceding `_currentEpoch`.
+     * @dev Only valid when `_hasPreviousEpoch` is true (after the first `updateEpoch` call).
+     *      Retaining the previous epoch allows attestations produced during a signing ceremony
+     *      that straddles a rollover boundary to be submitted after the epoch has advanced.
+     */
+    EpochState private _previousEpoch;
```
rmeissner: Not sure what the advantages of "current + previous" epoch are over a mapping of epoch to groupkey (besides minimal gas savings on update). 

We have multisig transactions that are pending for days or weeks, so I would not assume that our current epoch time (or 2x the epoch time) is enough to get a transaction executed. 

Also this kind of invalidates attestations (or makes the guard not accept them) which I find unexpected.

contracts/src/SafenetGuard.sol
remedcu: Two things:
- This was the only way I could think of for the modules (so your assumption was right)
- This way, I think Safenet (Research Team) can handle the attestation submission, and Safe Wallet can work as usual without any change (less dependency).

contracts/src/SafenetGuard.sol
remedcu: I don't think we would need more than the current and previous epoch for our use case, since each epoch is 24 hours.

> We have multisig transactions that are pending for days or weeks, so I would not assume that our current epoch time (or 2x the epoch time) is enough to get a transaction executed.

Yes, but when the transaction is initially proposed, I expect the attestation to be checked by the validators and then submitted to the guard on a particular chain (which might be in the current or previous epoch). During execution, we only check if an attestation is present for that particular safeTxHash (which was already verified during the `submitAttestation(...)` step.

rmeissner: our epoch is currently 2 hours (see https://github.com/safe-research/safenet/blob/main/docs/configuration.md#network-configuration)

rmeissner: > I expect the attestation to be checked by the validators and then submitted to the guard on a particular chain

This is interesting, as I would have expected the attestation to be part of the signature (so not an extra call)

contracts/src/SafenetGuard.sol
rmeissner: For me neither one is a requirement. For me not having an extra call (i.e. having no state beyond the epochs) would have been preferable.

As this is a prototype for technical  evaluation, not sure if it is important which one we chose (so we probably don't need to change your approach).

contracts/src/SafenetGuard.sol
remedcu: > This is interesting, as I would have expected the attestation to be part of the signature (so not an extra call)

For a normal transaction, yes, we could do that. For module transactions, this isn't available right now.

If we did in that way, then we would require more epochs than just the current and previous.

rmeissner: As said for me supporting modules is not a requirement right now (as we did not explicitly consider them with the current rules either).

With the approach that attestations are submitted to the target chain once available for me has (at least) 2 downsides:
- You potentially push attestations for transactions that are never executed
- You require another service for each chain that monitors attestations and submits them

remedcu: Expanded to have a history of epoch information.

contracts/src/SafenetGuard.sol
remedcu: Removed module support for now.

contracts/src/SafenetGuard.sol
```diff
@@ -0,0 +1,698 @@
...
+                safeTxGas: 0,
+                baseGas: 0,
+                gasPrice: 0,
+                gasToken: address(0),
+                refundReceiver: address(0),
+                nonce: 0
+            })
+        );
+        require(_attestations[safeTxHash] != bytes32(0), NoModuleAttestationPending());
+        delete _attestations[safeTxHash];
```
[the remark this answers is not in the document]
remedcu: Removed module support for now.

nlordell (commented):
From a high level, this is what I had in mind for the guard and it generally looks good to me. I talked with @rmeissner yesterday in the office about a couple of things:

- modules: _I agree we shouldn't support it at the moment_ as we don't have a working replay protection mechanism with them (so - an attestation can be reused to execute the same module transaction multiple times, or if you don't allow that, then you potentially brick modules that don't have a way of salting their transactions). That being said, it is important to make sure to think about this in the context of AA (I created a project and assigned you to it - its currently in the backlog but we can discuss it in the next weekly planning)
- pre-approving transactions: @rmeissner doesn't like this pattern but I do 😅. We agreed that, it _may_ be that we go back to that pattern depending on whether or not we want to support modules. My argument is that I want to have a single path for attestations instead of 2 (one with signatures, one without).

My general comment on the PR itself is that I would have loved for this to be done in pieces instead of being dropped into one bit change with everything implemented.

nlordell (approved):

contracts/src/guard/README.md
```diff
@@ -0,0 +1,42 @@
...
+- Inline attestation requires no separate on-chain registration for the happy path.
+- Replay protection is inherited from the Safe nonce — no additional registry.
+- `updateEpoch` is permissionless, improving liveness during validator set changes.
+- Fully self-contained: no cross-chain calls at execution time.
+- Escape hatch provides a liveness guarantee if Safenet is unavailable.
+
+### Cons
+
+- Non-standard `signatures` encoding: wallets and dApps must be Safenet-aware to append the attestation trailer; standard Safe UIs will not work without modification.
+- All historic epoch keys remain valid for new transactions. A compromised past key can sign attestations for future transactions (though not replay past ones, due to nonce binding).
```
rmeissner, line 40: > compromised past key

This references a compromised FROST group key, right?

contracts/src/guard/SafenetGuardA.sol
```diff
@@ -0,0 +1,455 @@
...
+     * @dev The security of the escape hatch depends on this value being large enough for Safe
+     *      owners to detect and cancel a mistakenly registered hash before it becomes executable.
+     */
+    uint256 private immutable _ALLOW_TX_DELAY;
+
+    /**
+     * @notice The currently active epoch number.
+     * @dev Set at construction and advanced on every `updateEpoch` call.
+     */
+    uint64 private _activeEpoch;
```
rmeissner, line 64: shouldn't private mutable fields start with `$`?

contracts/src/guard/README.md
remedcu: Yes, in cases where we have some validators who have turned malicious or whose validator keys have been compromised. In this case, the attacker only needs to obtain the required threshold of validator keys from previous validators (if they are no longer participating) or from validators who have rotated their keys (due to suspicion that the keys have been leaked/attacked).

contracts/src/guard/SafenetGuardA.sol
remedcu: Yes, it should have. Will change in the next PR.

PR #350, opened 2026-04-24
Add oracle result event handling to state machine

rmeissner:
## Summary
This PR adds support for handling oracle result events in the validator state machine. When an oracle approves or rejects a transaction proposal, the state machine now transitions appropriately—either dropping the state on rejection or moving to the nonce commitment collection phase on approval.

## Key Changes
- **New event type**: Added `OracleResultEvent` to represent oracle approval/rejection decisions with fields for requestId, proposer, result, and approval status
- **Event handling**: Implemented `handleOracleResult` function that:
  - Validates the signing state exists and is in `waiting_for_oracle` phase
  - Drops the state if the oracle rejects the transaction
  - Transitions to `collect_nonce_commitments` state if approved, creating nonce commitments via the signing client
- **Schema updates**: Added validation schemas for the new `OracleResult` event and integrated it into the state transition schema
- **Event parsing**: Updated `logToTransition` to parse `OracleResult` events from on-chain logs
- **State enrichment**: Enhanced `waiting_for_oracle` signing state to include `gid`, `signatureId`, and `sequence` fields needed for nonce commitment creation
- **Configuration**: Added `allowedOracles` and `oracleTimeout` configuration parameters throughout the system
- **Event watching**: Updated the transition watcher to monitor oracle contract events

## Implementation Details
- The oracle result handler creates nonce commitments using the group ID, signature ID, sequence number, and list of signers from the waiting state
- The deadline for the next phase is calculated as the current block plus the configured signing timeout
- A `sign_reveal_nonce_commitments` action is generated to trigger the signing process
- The consensus state is updated to map the signature ID to the request ID for tracking

https://claude.ai/code/session_01LTR5hFdQ1BqgwXYqStAAgs


validator/src/machine/consensus/oracleResult.ts
```diff
@@ -0,0 +1,49 @@
...
+	if (event.oracle !== status.oracle) {
+		logger?.debug?.(`Oracle mismatch for result ${event.requestId}: expected ${status.oracle}, got ${event.oracle}`);
+		return {};
+	}
+	if (!event.approved) {
+		logger?.info?.("Oracle rejected transaction, dropping state", { requestId: event.requestId });
+		return { signing: [event.requestId] };
+	}
+	logger?.info?.("Oracle approved transaction, participating in signing", { requestId: event.requestId });
+	const { nonceCommitments, nonceProof } = signingClient.createNonceCommitments(
```
rmeissner: the following logic is shared with non-oracle handlesign method and could be extracted into a method.

validator/src/machine/transitions/watcher.ts
```diff
@@ -124,7 +124,7 @@ export class OnchainTransitionWatcher {
 			...this.#watcherConfig,
 			lastIndexedBlock,
 			blockTime,
-			address: [this.#config.consensus, this.#config.coordinator],
+			address: [this.#config.consensus, this.#config.coordinator, ...(this.#config.allowedOracles ?? [])],
```
rmeissner, line 130: maybe worth documenting that allowedOracles should never emit consensus or coordinator related events (as we currently don't check the address that is emitting the events for these)

rmeissner (commented):
Please also add an integration test for this (or adjust the existing one to also propose a oracle packet)

validator/src/consensus/integration.test.ts
```diff
@@ -540,4 +550,133 @@ describe("integration", () => {
...
+		await testClient.revert({ id: snapshotId });
+		snapshotId = await testClient.snapshot();
+		const deployHash = await testClient.deployContract({ abi: [], bytecode: oracleBytecode });
+		await testClient.mine({ blocks: 1 });
+		const deployReceipt = await testClient.getTransactionReceipt({ hash: deployHash });
+		const oracleAddress = deployReceipt.contractAddress as Address;
+		testLogger.notice(`Deployed AlwaysApproveOracle at ${oracleAddress}`);
+
+		// Reset snapshot so setup() sees the oracle deployment
+		snapshotId = await testClient.snapshot();
```
rmeissner: By doing this, it is not reverting to the original snapshot afterwards (as it will not use the snapshot with oracle deployd. 

Better would be to deploy the oracle as part of the integration test script.

rmeissner (commented):
Please make sure that the new integration test is run when triggering `npm run test:integration`

scripts/run_integration_test.sh
```diff
@@ -32,8 +32,29 @@ npm run -w contracts cmd:deploy -- \
     --sender 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 \
     --broadcast
 
-# --- 4. Run Client Integration Tests ---
+# --- 4. Deploy Oracle for Integration Testing (optional) ---
+# If deployment fails the oracle integration test is skipped, but other tests still run.
+echo "Deploying AlwaysApproveOracle..."
```
rmeissner: Why not add this deployment ot the contract deploy script so that the oracle is deployed as part of `npm run -w contracts cmd:deploy`?

validator/src/machine/consensus/oracleResult.ts
```diff
@@ -0,0 +1,37 @@
...
+		logger?.debug?.(`No waiting_for_oracle state for oracle result ${event.requestId}`);
+		return {};
+	}
+	if (event.oracle !== status.oracle) {
+		logger?.debug?.(`Oracle mismatch for result ${event.requestId}: expected ${status.oracle}, got ${event.oracle}`);
+		return {};
+	}
+	if (!event.approved) {
+		logger?.info?.("Oracle rejected transaction, dropping state", { requestId: event.requestId });
+		return { signing: [event.requestId] };
```
rmeissner: an explicit `undefined` to indicate the removal of the signing request would be better

validator/src/machine/signing/nonces.ts
rmeissner: maybe the callback context building should be extracted into a method, so that the branches don't have to be nested and are rather a flat `if -> return` setup.

nlordell (approved):

PR #352, opened 2026-04-29
Safenet Guard B

remedcu:
This is the second type of Guard for the Safenet Guard Tech Evaluation, which has some key changes from `SafenetGuardA`:
- Only 2 epochs stored
- Module Transaction supported
- Attestation Submission and Verification are different steps

We will include the test once the evaluation part is complete. The NatSpec is kept to a minimum for the contract.|


contracts/test/libraries/ConsensusMessages.t.sol
```diff
@@ -15,9 +15,9 @@ contract ConsensusMessagesTest is Test {
                 1,
                 0xbaddad42,
                 Secp256k1.Point({
-                    x: 0x8318535b54105d4a7aae60c08fc45f9687181b4fdfc625bd1a753fa7397fed75,
-                    y: 0x3547f11ca8696646f2f3acb08e31016afac23e630c5d11f59f61fef57b0d2aa5
-                })
+                x: 0x8318535b54105d4a7aae60c08fc45f9687181b4fdfc625bd1a753fa7397fed75,
+                y: 0x3547f11ca8696646f2f3acb08e31016afac23e630c5d11f59f61fef57b0d2aa5
+            })
```
nlordell: change request: please revert the unnecessary formatting. We can pin to a known good foundry version (last release was 1.5.1) instead of dealing with intermittent CI issues.

My rationale here is: this formatting is clearly wrong and a bug in `nightly` Foundry. I would even suggest opening an issue there so they can fix it.

contracts/foundry.toml
```diff
@@ -31,3 +32,6 @@ fs_permissions = [
   # script to write the deployment transaction details in JSON format for the Tx Builder Safe App.
   { access = "read-write", path = "build/staking-deployment.json" },
 ]
+
+[lint]
+exclude_lints = ["block-timestamp"]
```
nlordell: nit: I personally prefer adding lint exceptions on a line-by-line basis. It makes the fact that this would generate a linting error if not for an exception more obvious. Generally, the lint makes sense (you need to be careful around `block.timestamp`).

contracts/src/guard/SafenetGuardB.sol
```diff
@@ -0,0 +1,497 @@
...
+                to: to,
+                value: value,
+                data: data,
+                operation: SafeTransaction.Operation(uint8(operation)),
+                safeTxGas: 0,
+                baseGas: 0,
+                gasPrice: 0,
+                gasToken: address(0),
+                refundReceiver: address(0),
+                nonce: nonce
```
nlordell, line 485: change request: This collides with actual Safe transactions, we can do better IMO:

```
nonce: (uint256(uint160(module)) << 96) | uint256(nonce)
```

We would then change `nonce` to be `uint64` - a reasonable upper bound, and have it keyed only on the module address.

This will also not reasonable collide with actual Safe transaction nonces (since its value will be >2**96).

contracts/src/guard/SafenetGuardB.sol
```diff
@@ -0,0 +1,497 @@
...
+        address to,
+        uint256 value,
+        bytes memory data,
+        Enum.Operation operation,
+        address module
+    ) external override returns (bytes32 moduleTxHash) {
+        if (_isAutoAllowed(to, value, data, operation)) return bytes32(0);
+
+        // forge-lint: disable-next-line(asm-keccak256)
+        bytes32 moduleNonceKey = keccak256(abi.encode(msg.sender, module, to, value, data, operation));
```
nlordell: question: I don't understand the rationale behind this key. Why not just `module` address?

nlordell (changes requested):
I like the solution for module transactions, requested some changes on the actual implementation of key-ing and choice of nonces.

contracts/test/libraries/ConsensusMessages.t.sol
remedcu: Removed 👍🏾

contracts/foundry.toml
remedcu: Removed 👍🏾 

I also prefer the line-by-line basis; the only reason not to go with that was the Staking contract changes, which were audited. But as the foundry is pinned now, this does not need to be resolved.

contracts/src/guard/SafenetGuardB.sol
remedcu: Initially, I thought of just a module address, but the current implementation allows sequential transactions of a single tx param type and also allows non-sequential transactions of different tx param types.

So, let's say we have mainly 2 transactions from a module:
- USDC Transfer (frequent)
- DAI Transfer (less frequent)

With the module approach, we need these transactions to be sequential (not the end of the world, but I thought it could have been better)
With the per-transaction param approach, USDC transfers can happen one after the other as required, and DAI transfers can independently perform sequential ones as well.

These are mainly useful for modules without sequence management.

nlordell: Thanks for the clarification. I would definitely include this somewhere, as it was not immediately obvious to me. I'm sorry if you had already written this down somewhere and I missed it, in which case I would request a comment somewhere closer to the code so this non-obvious choice is explained.

nlordell:
```suggestion
        bytes32 moduleNonceKey = keccak256(abi.encode(msg.sender, module, to, value, keccak256(data), operation));
```

Small optimization to prevent a large memory copy.

remedcu: https://github.com/safe-research/safenet/pull/352/changes#diff-10608616c309997efad298a29747df01140128592645615eb73ca5d357f9a9ffR68

It was in the README. I will add it to the contract as well, in a relevant place.

contracts/src/guard/SafenetGuardB.sol
remedcu: Updated 👍🏾

contracts/src/guard/SafenetGuardB.sol
remedcu: Updated 👍🏾

contracts/src/guard/SafenetGuardB.sol
```diff
@@ -0,0 +1,515 @@
...
+     * @notice True once the first `updateEpoch` call has populated `$previousEpoch`.
+     * @dev Guards `previousEpoch()` and `_resolveGroupKey` against treating the zero-initialised
+     *      `$previousEpoch` slot as a valid epoch entry before any rollover has occurred.
+     *      Without this flag, `_resolveGroupKey(0)` would match the zero-initialised
+     *      `$previousEpoch.epoch` and return a zero group key. FROST verification against a zero
+     *      key is trivially forgeable, so an attacker could register a fake attestation for any
+     *      transaction without a real signing ceremony.
+     */
+    // forge-lint: disable-next-line(mixed-case-variable)
+    bool private $hasPreviousEpoch;
```
nlordell: question: Instead of this, why not consider the zero group key invalid for signature verification? We can even enforce it on the `FROSTCoordinator` level (where we taint group ceremonies where this happens), and an honest participant can ensure that this never happens by generating a truly random `c[0]`. I actually think rejecting 0 group keys to be a positive change in our core contracts!

contracts/src/guard/SafenetGuardB.sol
```diff
@@ -0,0 +1,515 @@
...
+     *      The key intentionally includes the full tx-param tuple rather than just `(safe, module)`.
+     *      A per-module counter would serialise all operations from a module onto a single sequence:
+     *      if two different tx-param combinations are attested at consecutive nonces (e.g. TxA at
+     *      nonce=0, TxB at nonce=1) but TxB is executed first, the guard reconstructs the wrong hash
+     *      and rejects it — TxB is blocked until TxA executes, creating a deadlock. With per-tx-param
+     *      counters each combination has an independent track and operations may execute in any order.
+     *      The counter is packed with the module address as `(uint256(uint160(module)) << 96) | counter`
+     *      and used as the `nonce` field in the module tx hash via `_packedModuleNonce`. The packing
+     *      ensures two different modules on the same Safe cannot collide at the same counter value.
+     *      The counter increments on both successful execution and cancellation.
```
nlordell, line 61: This is a cool learning that wasn't immediately obvious to me. Nice thing to record on Notion as well!

contracts/src/guard/SafenetGuardB.sol
nlordell: Wait - we check group keys are non-zero below, so isn't `_resolveGroupKey(0) == zero` equivalent to this flag?

contracts/src/guard/SafenetGuardB.sol
```diff
@@ -0,0 +1,515 @@
...
+    /**
+     * @dev For module transactions use `submitModuleAttestation`, which reads and validates the
+     *      current module nonce internally.
+     *      Only one pending attestation per `safeTxHash` is permitted at a time.
+     */
+    function submitAttestation(bytes32 safeTxHash, uint64 epoch, FROST.Signature calldata signature) external {
+        require($attestations[safeTxHash] == bytes32(0), AttestationAlreadySubmitted());
+        // forge-lint: disable-next-line(asm-keccak256)
+        bytes32 sigId = keccak256(abi.encode(signature.r.x, signature.r.y, signature.z));
+        Secp256k1.Point memory groupKey = _resolveGroupKey(epoch);
```
nlordell, line 177: question: why not?
```suggestion
        Secp256k1.Point memory groupKey = _resolveGroupKey(epoch);
        Secp256k1.requireNonZero(groupKey);
```

contracts/src/guard/SafenetGuardB.sol
```diff
@@ -0,0 +1,515 @@
...
+        address to,
+        uint256 value,
+        bytes calldata data,
+        Enum.Operation operation,
+        uint64 epoch,
+        FROST.Signature calldata signature
+    ) external {
+        // forge-lint: disable-next-line(asm-keccak256)
+        bytes32 moduleNonceKey = keccak256(abi.encode(safe, module, to, value, keccak256(data), operation));
+        uint64 nonce = $moduleNonces[moduleNonceKey];
```
nlordell, line 207: nit: not incrementing the nonce on use seems weird to me. I think I understand _why_ (because verifying or cancelling the transaction requires the nonce as well), but it feels like a code smell.

This is, however, a technical evaluation, so not blocking the PR. Definitely a learning for me is that getting the current Safenet model to work with module transactions is much hackier than I had expected. I would go as far to say that there it is not obvious that there are no soundness issues with module transactions + Safenet.

nlordell (approved):
Generally LGTM.

Biggest learning for me is that: "module transactions is a huge hack right now, and it is not obvious without formal verification that it is sound". Maybe something to record on Notion as part of the findings :).

contracts/src/guard/SafenetGuardB.sol
remedcu: That is correct, updated 👍🏾

contracts/src/guard/SafenetGuardB.sol
remedcu: Simplified the `_resolveGroupKey(...)` and added this check 👍🏾

PR #354, opened 2026-04-30
Safenet Cosigner

remedcu:
This is a different approach to Safenet Guard. Technically, it's not a guard, but a cosigner.

This also shares some features of the guard, like:
- Only 2 epochs stored
- Verification of the hash based on the FROST signature

Here, the escape hatch must be created differently from the other two guards, since the transaction must go through `execTransaction(...)`. The escape hatch is developed, where any Arbitrary transaction can be executed with `max(threshold - 1, 1)` owner signatures. The threshold count is set this way for safety, where the setup is 1 of n (which ideally should not be configured that way with Safenet Cosigner). In this way, the threshold owner schedules the removal of `SafenetCosigner` (similar to how the Guard removal is scheduled using `allowTransaction`) and then executes the removal once the timestamp passes. The first tx signs with the `safeTxHash`, and the second tx executes the tx that results in the `safeTxHash` that was allowed in the first. There is a time delay built in between these two transactions.

Initially, the approach was to allow any owner to call `removeOwner(...)` with hardcoded values to remove the owner (`address(this)`) and reduce the threshold count by 1, or leave it unchanged. The aim was to allow the removal of the `SafenetCosigner` from the owner list if Safenet is down. But based on our recent discussion, we understand that we should not grant any owner unilateral authority to initiate this process, even though it is only allowed for the removal of `SafenetCosigner`.

Later, it was updated so that the `threshold-1` owners can sign an arbitrary transaction once and wait for the timestamp to be reached before allowing it in `SafenetCosigner`. Once the timestamp has passed, the same signature used by the owners can be used for execution (since `SafenetCosigner.isValidSignature(...)` will now be allowed). With Guards, the threshold owners had to sign twice (once to queue the tx and once to execute it), but with Cosigner in the current design, the threshold owner only needs to sign once to remove Cosigner. As the intention is the same (to remove the Cosigner) and it is already approved by the threshold count of owners, the need for another threshold count of owners to sign for execution seems redundant. If the intention changes (i.e., decided not to remove the owner), the threshold count of owners could execute a dummy transaction anyway to increase the nonce, thereby invalidating the cosigner removal transaction. But the shortcomings were noted here: https://github.com/safe-research/safenet/pull/354#discussion_r3218757818

More details are added in the README.

### DIFF
--- a/contracts/src/cosigner/README.md
+++ b/contracts/src/cosigner/README.md
@@ -1,0 +1,52 @@
+# Safenet Cosigner
+
+This directory contains an EIP-1271 contract signer implementation for Safenet. Unlike the guard variants, which intercept every transaction as a Safe Guard module, the cosigner participates as a Safe owner and gates transactions through Safe's standard EIP-1271 signature verification flow.
+
+---
+
+## SafenetCosigner
+
+**File:** `SafenetCosigner.sol`
+
+SafenetCosigner is an EIP-1271 contract signer that gates Safe transactions behind FROST threshold-signature attestation produced by the Safenet validator network. It is deployed once and added as an owner of a Safe alongside the human key holders.
+
+### How it works
+
+**EIP-1271 integration.** Safe calls `isValidSignature(safeTxHash, contractSignature)` on any owner that is a contract. The cosigner implements this interface and returns `0x1626ba7e` when it approves the transaction. The FROST attestation travels as dynamic contract signature data inside Safe's existing `signatures` bytes — no custom trailer encoding is required.
+
+**Signature construction.** Safe requires `signatures` entries sorted ascending by owner address. For the cosigner entry, include a contract signature slot (`v = 0`) and append the FROST attestation as dynamic data:
+
+```
+Static slot (65 bytes):
+  r (bytes32) = address(cosigner) left-padded with zeros
+  s (bytes32) = byte offset of the dynamic data = (total static entries) * 65
+  v (uint8)   = 0x00
+
+Dynamic data (appended after all static entries):
+  uint256                             : byte length of the encoded attestation
+  abi.encode(uint64, FROST.Signature) : epoch and FROST signature
+```
+
+To use the escape hatch instead, set the dynamic data to an empty byte sequence (length = 0, no following bytes).
+
+**Replay protection.** The Safe nonce is embedded in `safeTxHash`, making every attested message unique. No separate spent-signature registry is needed.
+
+**Epoch state.** Only the two most recent epochs are retained: `$currentEpoch` and `$previousEpoch`. On each `updateEpoch` call the current epoch shifts to previous and the new epoch becomes current. The sliding window allows in-flight transactions to complete across epoch boundaries while preventing unbounded storage growth. `updateEpoch` is permissionless — any party holding the FROST-signed rollover message can advance the epoch.
+
+**Escape hatch.** Owners can register a specific transaction for time-delayed execution via `allowTransaction`. The call must go through the Safe's own `execTransaction`, so the Safe's owner-signature threshold is required to register an allowance — including the cosigner's approval if the human owners alone cannot reach it. After a configurable delay, passing empty bytes as the dynamic contract signature data causes the cosigner to approve via the matured allowance. Allowances are not deleted on use because `isValidSignature` is `view`; replay is prevented by Safe's nonce advancing after execution.
+
+### Pros
+
+- Uses Safe's standard EIP-1271 contract-signature mechanism — no custom trailer encoding required.
+- Replay protection is inherited from the Safe nonce — no additional registry.
+- Only two epoch keys are retained, bounding storage growth regardless of how many rollovers occur.
+- `updateEpoch` is permissionless, improving liveness during validator set changes.
+- Fully self-contained: no cross-chain calls at execution time.
+- Escape hatch provides a liveness guarantee if Safenet is unavailable.
+
+### Cons
+
+- Only the two most recent epoch keys are valid. An in-flight transaction attested under an older epoch will be rejected once two subsequent rollovers have occurred.
+- Registering an escape-hatch allowance requires the cosigner's approval at registration time, so it must be done proactively while Safenet is still available.
+- Cancelling an allowance goes through `execTransaction` and requires the Safe's owner-signature threshold. If the remaining human owners can reach that threshold without the cosigner, cancellation is still possible even when Safenet is unavailable.
+- Escape-hatch UX requires owners to proactively register a transaction hash and wait out the full delay, which may be operationally burdensome in time-sensitive situations.
--- a/contracts/src/cosigner/README.md
+++ b/contracts/src/cosigner/README.md
@@ -47,7 +47,7 @@ The registered hash is nonce-bound to the Safe's nonce at registration time. If 
 - `updateEpoch` is permissionless, improving liveness during validator set changes.
 - Fully self-contained: no cross-chain calls at execution time.
 - Escape hatch can be registered by any single Safe owner without Safenet availability, providing a liveness guarantee if Safenet is unavailable.
-- Relay-compatible escape hatch registration via EIP-712 (`allowEscapeHatchWithSig`), supporting both EOA and contract owners.
+- Relay-compatible escape hatch registration via `allowEscapeHatchWithSig`, using Safe's own `checkNSignatures` for signature verification — supports ECDSA, EIP-1271 contract signatures, and pre-approved hashes without any custom signature logic.
 
 ### Cons
 
--- a/contracts/src/cosigner/README.md
+++ b/contracts/src/cosigner/README.md
@@ -33,7 +33,11 @@ To use the escape hatch instead, set the dynamic data to an empty byte sequence 
 
 **Epoch state.** Only the two most recent epochs are retained: `$currentEpoch` and `$previousEpoch`. On each `updateEpoch` call the current epoch shifts to previous and the new epoch becomes current. The sliding window allows in-flight transactions to complete across epoch boundaries while preventing unbounded storage growth. `updateEpoch` is permissionless — any party holding the FROST-signed rollover message can advance the epoch.
 
-**Escape hatch.** Owners can register a specific transaction for time-delayed execution via `allowTransaction`. The call must go through the Safe's own `execTransaction`, so the Safe's owner-signature threshold is required to register an allowance — including the cosigner's approval if the human owners alone cannot reach it. After a configurable delay, passing empty bytes as the dynamic contract signature data causes the cosigner to approve via the matured allowance. Allowances are not deleted on use because `isValidSignature` is `view`; replay is prevented by Safe's nonce advancing after execution.
+**Escape hatch.** Any single Safe owner can register a time-delayed `removeOwner` call that removes the cosigner from the Safe by calling `allowEscapeHatch` directly, or `allowEscapeHatchWithSig` when submitting via a relay service. Registration does not require a Safenet attestation, making it available even when the Safenet network is unavailable. Both functions accept an `EscapeHatchRequest` struct that captures the target Safe address (`safe`), the `removeOwner` parameters (`prevOwner`, `threshold`), and the Safe transaction gas parameters (`safeTxGas`, `baseGas`, `gasPrice`, `gasToken`, `refundReceiver`).
+
+After the configured delay, the Safe owners execute the pre-registered `removeOwner` transaction by passing empty bytes as the cosigner's dynamic signature data; the cosigner approves it via the empty-signature path in `isValidSignature`. Registrations are not deleted on use because `isValidSignature` is `view`; replay is prevented by Safe's nonce advancing after execution.
+
+The registered hash is nonce-bound to the Safe's nonce at registration time. If other transactions advance the nonce before execution, the registration becomes stale and must be repeated. To invalidate a pending registration, `threshold` owners can execute a dummy transaction to advance the Safe nonce.
 
 ### Pros
 
@@ -42,11 +46,12 @@ To use the escape hatch instead, set the dynamic data to an empty byte sequence 
 - Only two epoch keys are retained, bounding storage growth regardless of how many rollovers occur.
 - `updateEpoch` is permissionless, improving liveness during validator set changes.
 - Fully self-contained: no cross-chain calls at execution time.
-- Escape hatch provides a liveness guarantee if Safenet is unavailable.
+- Escape hatch can be registered by any single Safe owner without Safenet availability, providing a liveness guarantee if Safenet is unavailable.
+- Relay-compatible escape hatch registration via EIP-712 (`allowEscapeHatchWithSig`), supporting both EOA and contract owners.
 
 ### Cons
 
 - Only the two most recent epoch keys are valid. An in-flight transaction attested under an older epoch will be rejected once two subsequent rollovers have occurred.
-- Registering an escape-hatch allowance requires the cosigner's approval at registration time, so it must be done proactively while Safenet is still available.
-- Cancelling an allowance goes through `execTransaction` and requires the Safe's owner-signature threshold. If the remaining human owners can reach that threshold without the cosigner, cancellation is still possible even when Safenet is unavailable.
-- Escape-hatch UX requires owners to proactively register a transaction hash and wait out the full delay, which may be operationally burdensome in time-sensitive situations.
+- Escape-hatch UX requires owners to proactively register the `removeOwner` hash and wait out the full delay, which may be operationally burdensome in time-sensitive situations.
+- The registered hash is nonce-bound; if the Safe nonce advances before execution (e.g. due to another transaction), the registration must be repeated. Though, this is by design.
+- Invalidating a pending escape-hatch registration requires `threshold` owners to execute a dummy transaction without the cosigner's approval, which is only possible if the human owners can independently reach the Safe threshold.
--- a/contracts/src/cosigner/README.md
+++ b/contracts/src/cosigner/README.md
@@ -33,7 +33,21 @@ To use the pre-approved transaction path instead, set the dynamic data to an emp
 
 **Epoch state.** Only the two most recent epochs are retained: `$currentEpoch` and `$previousEpoch`. On each `updateEpoch` call the current epoch shifts to previous and the new epoch becomes current. The sliding window allows in-flight transactions to complete across epoch boundaries while preventing unbounded storage growth. `updateEpoch` is permissionless — any party holding the FROST-signed rollover message can advance the epoch.
 
-**Pre-approved transactions.** Safe owners can register any Safe transaction for time-delayed execution by calling `allowTransaction`. Registration requires signatures from `max(threshold - 1, 1)` Safe owners over `SafeTransaction.hash` — the same hash Safe owners sign for normal `execTransaction` — and does not require a Safenet attestation, making it available even when Safenet is unavailable. `allowTransaction` accepts a `SafeTransaction.T` struct covering all Safe transaction parameters (`safe`, `to`, `value`, `data`, `operation`, `safeTxGas`, `baseGas`, `gasPrice`, `gasToken`, `refundReceiver`, `nonce`) along with a `chainId` used to domain-separate the hash.
+**Pre-approved transactions.** Safe owners can register any Safe transaction for time-delayed execution by calling `allowTransaction(safe, safeTxHash, signatures)`. The caller supplies the Safe address, the precomputed `safeTxHash`, and `max(threshold - 1, 1)` owner signatures over the EIP-712 typed message `AllowTransaction(bytes32 safeTxHash)` under the cosigner's own domain. No Safenet attestation is required, making this available even when Safenet is unavailable.
+
+The EIP-712 domain used for `allowTransaction` signatures is:
+```
+EIP712Domain(uint256 chainId, address verifyingContract)
+  chainId           = deployment chain id
+  verifyingContract = address(cosigner)
+```
+
+The struct type is:
+```
+AllowTransaction(bytes32 safeTxHash)
+```
+
+Owners compute `safeTxHash` off-chain using `SafeTransaction.hash(safeTx)` — the same hash they sign for normal `execTransaction` — and sign the `AllowTransaction` EIP-712 message over it. The cosigner's `allowTransactionDomainSeparator()` view function returns the domain separator for off-chain message construction.
 
 After the configured delay, the Safe owners execute the pre-registered transaction by passing empty bytes as the cosigner's dynamic signature data; the cosigner approves it via the empty-signature path in `isValidSignature`. Registrations are not deleted on use because `isValidSignature` is `view`; replay is prevented by Safe's nonce advancing after execution.
 
@@ -47,7 +61,7 @@ The registered hash is nonce-bound to the Safe's nonce at registration time. If 
 - `updateEpoch` is permissionless, improving liveness during validator set changes.
 - Fully self-contained: no cross-chain calls at execution time.
 - Pre-approved transactions require no Safenet attestation, providing a liveness guarantee if Safenet is unavailable.
-- `allowTransaction` uses `SafeTransaction.hash` — the same hash owners sign for normal execution — so no additional signing infrastructure is needed on the client side.
+- `allowTransaction` signatures use a dedicated EIP-712 typed message (`AllowTransaction`), domain-separated by the cosigner contract address — preventing cross-contract replay.
 - Registration requires `max(threshold - 1, 1)` owner signatures, matching the number of human signatures needed to execute the transaction, preventing unilateral registration by a single owner.
 - Any Safe transaction can be pre-approved, not just cosigner removal — `removeOwner` is the typical escape hatch but the mechanism is general.
 
--- a/contracts/src/cosigner/README.md
+++ b/contracts/src/cosigner/README.md
@@ -33,7 +33,7 @@ To use the escape hatch instead, set the dynamic data to an empty byte sequence 
 
 **Epoch state.** Only the two most recent epochs are retained: `$currentEpoch` and `$previousEpoch`. On each `updateEpoch` call the current epoch shifts to previous and the new epoch becomes current. The sliding window allows in-flight transactions to complete across epoch boundaries while preventing unbounded storage growth. `updateEpoch` is permissionless — any party holding the FROST-signed rollover message can advance the epoch.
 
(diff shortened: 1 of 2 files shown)


contracts/src/cosigner/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+**Epoch state.** Only the two most recent epochs are retained: `$currentEpoch` and `$previousEpoch`. On each `updateEpoch` call the current epoch shifts to previous and the new epoch becomes current. The sliding window allows in-flight transactions to complete across epoch boundaries while preventing unbounded storage growth. `updateEpoch` is permissionless — any party holding the FROST-signed rollover message can advance the epoch.
+
+**Escape hatch.** Owners can register a specific transaction for time-delayed execution via `allowTransaction`. The call must go through the Safe's own `execTransaction`, so the Safe's owner-signature threshold is required to register an allowance — including the cosigner's approval if the human owners alone cannot reach it. After a configurable delay, passing empty bytes as the dynamic contract signature data causes the cosigner to approve via the matured allowance. Allowances are not deleted on use because `isValidSignature` is `view`; replay is prevented by Safe's nonce advancing after execution.
```
rmeissner: > including the cosigner's approval if the human owners alone cannot reach it

Why was this decision taken? In the scenario where Safenet is not available anymore then this escape hatch would potentially not work anymore, right?

In what scenarios do we expect the escape hatch to be used?

remedcu: This is more of a niche case where Safenet is available, but does not attest to a particular transaction due to its own risk assessment (validators don't think it is the right transaction), but validators should, in all cases, attest to the escape hatch for time-delayed transactions, which then should allow that particular transaction to be executed. This is not useful in the case where threshold human owner signature is available.

Maybe I should document it better 🤦🏾

nlordell: Ouf... this feels like a big foot-gun:

- If you require Safenet, and it gets turned off (validators stop running), then you lose access to your account
- Making it an unconditional escape hatch (so you can pass a special signature which is a pre-image of a cosigner escape hatch approval transaction and have the cosigner always accept it), feels more correct, but also super scary that you have a way to get a signer to always accept a transaction (in particular, it can be used to extract funds using the `payment` feature of the Safe transaction without additional approvals). I don't know if this is actually an issue (the previous example, simple transfers are accepted by Safenet ATM, so it wouldn't change much, but might in the future once we add recipient checks).

I would love for us to discuss this tomorrow in our office hours!

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,239 @@
...
+    EpochState private $previousEpoch;
+
+    /**
+     * @notice True once the first `updateEpoch` call has populated `$previousEpoch`.
+     * @dev Guards `_resolveGroupKey` against treating the zero-initialised `$previousEpoch`
+     *      slot as a valid epoch entry before any rollover has occurred. Without this flag,
+     *      `_resolveGroupKey(0)` would return a zero group key — trivially forgeable via FROST.
+     */
+    // forge-lint: disable-next-line(mixed-case-variable)
+    bool private $hasPreviousEpoch;
```
nlordell: Again, I would just have 0 group keys be treated as "no group key".

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,244 @@
...
+     * @dev `$currentEpoch` shifts to `$previousEpoch` and the new epoch becomes `$currentEpoch`.
+     *      Only the two most recent epochs are retained; older keys are discarded on each rollover.
+     */
+    function updateEpoch(
+        uint64 proposedEpoch,
+        uint64 rolloverBlock,
+        Secp256k1.Point calldata newGroupKey,
+        FROST.Signature calldata signature
+    ) external {
+        require(proposedEpoch > $currentEpoch.epoch, EpochNotAdvancing());
```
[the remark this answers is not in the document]
nlordell: The rollover block is on the consensus chain, and different than the block number for this chain. This is invalid, although generally a good catch!

nlordell (commented):
One advantage of the cosigner is that the code is nice and simple. We have one point I would like to discuss around the escape hatch tomorrow.

contracts/src/cosigner/README.md
remedcu: For posterity, during today's meeting, it was discussed that it is important to have an escape hatch that works even without Safenet, so that in cases where it is n/n Safe, it should be possible to execute critical transactions (like removing `SafenetCosigner`) without Safenet Attestation.

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Updated 👍🏾

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,244 @@
...
+            (uint64 epoch, FROST.Signature memory sig) = abi.decode(_signature, (uint64, FROST.Signature));
+            bytes32 message = ConsensusMessages.transactionProposal(_CONSENSUS_DOMAIN_SEPARATOR, epoch, _hash);
+            FROST.verify(_resolveGroupKey(epoch), sig, message);
+            return EIP1271_MAGIC_VALUE;
+        }
+
+        uint256 executableAt = $allowedTransactions[msg.sender][_hash];
+        if (executableAt != 0 && block.timestamp >= executableAt) return EIP1271_MAGIC_VALUE;
+
+        revert AttestationNotFound();
```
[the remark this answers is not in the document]
remedcu: Updated `isValidSignature(...)` 👍🏾

contracts/src/cosigner/README.md
remedcu: Updated the escape hatch system (also updated the PR description) 👍🏾

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,360 @@
...
+        _CONSENSUS_DOMAIN_SEPARATOR = ConsensusMessages.domain(consensusChainId, consensusAddress);
+        _ALLOW_TX_DELAY = allowTransactionDelay;
+        // forge-lint: disable-next-item(asm-keccak256)
+        _DOMAIN_SEPARATOR = keccak256(
+            abi.encode(
+                keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)"),
+                keccak256("SafenetCosigner"),
+                keccak256("1"),
+                block.chainid,
+                address(this)
```
nlordell: nit: to be consistent with all our other contracts:


```suggestion
                keccak256("EIP712Domain(uint256 chainId,address verifyingContract)"),
                block.chainid,
                address(this)
```

nlordell: request change: the domain separator should not be an immutable. Chain IDs can change, and have in the past (see Ethereum vs Ethereum PoW).

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,360 @@
...
+     *      Any single Safe owner may call this directly — no Safenet attestation is required,
+     *      making the escape hatch available even when Safenet is unavailable.
+     *      `request.threshold` must equal the Safe's current threshold or current threshold minus one.
+     *      The registered hash is nonce-bound to the Safe's current nonce at registration time;
+     *      if other transactions advance the nonce before execution, re-registration is required.
+     *      To invalidate a pending registration, advance the Safe nonce with a dummy transaction.
+     */
+    function allowEscapeHatch(EscapeHatchRequest calldata request) external {
+        require(ISafe(payable(request.safe)).isOwner(msg.sender), NotSafeOwner());
+        _registerEscapeHatch(request);
```
nlordell: ~~request change~~ nit: I don't think allowing a single owner to do this is sound, as it allows them to unilaterally decide whether or not Safenet checks can be ignored in the context of the multisig. I would isntead suggest:

```
ISafe(payable(request.safe)).checkNSignatures(..., threshold - 1, ...)
```

nlordell (changes requested):

contracts/src/cosigner/SafenetCosigner.sol
rmeissner: +1

remedcu: Updated 👍🏾

contracts/src/cosigner/SafenetCosigner.sol
rmeissner: In the current process I agree. Because this basically makes the first step of the 2 step process threshold `1` instead of threshold `n - 1`.


I still think that allowing anyone to announce the escape hatch request  data would be simpler code wise. Just to be clear my proposal is a 3 step process:
- Announce data -> anyone can do this
- Execute safe transaction to the signer to allow list the tx (this can be checked now in the signer as we have the data to the hash) -> requires threshold n - 1 signers
- After delay passed execute the escape hatch as another safe transaction -> requires threshold n - 1 signers

nlordell:
> removeOwner(...) is allowed to be executed after a timedelay, and

I feel like this distinction adds unnecessary complication - why not just allow an arbitrary transaction? Can you elaborate on the rationale behind this design decision?

contracts/src/cosigner/SafenetCosigner.sol
nlordell: Now that I read through the whole PR, this makes a bit more sense to me - it isn't circumventing arbitrary transactions, but disabling Safenet only (akin to proposing this transaction).

One thing I don't like about this approach is that fixes other transaction parameters (like the gas ones), and makes it not possible to use with `multiSend` for example (so you can't use this to rotate to a new SAfenet version for example - in case the old network gets decommissioned).

Either way, goes from a `request change` to a `nit`.

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,356 @@
...
+                baseGas: request.baseGas,
+                gasPrice: request.gasPrice,
+                gasToken: request.gasToken,
+                refundReceiver: request.refundReceiver,
+                nonce: safeContract.nonce()
+            })
+        );
+    }
+
+    function _verifySignature(address signer, bytes32 hash, bytes calldata signature) private view {
```
rmeissner: nit: I am not really a fan that we implement signature verification logic here, when we have this readily available in the Safe

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,356 @@
...
+     *      allowing submission via a relay service where `msg.sender` is not the Safe owner.
+     *      Supports both ECDSA (EOA) and EIP-1271 (contract) owners. The signature must be
+     *      produced with `eth_signTypedData` over this contract's EIP-712 domain.
+     */
+    function allowEscapeHatchWithSig(EscapeHatchRequest calldata request, address signer, bytes calldata signature)
+        external
+    {
+        require(ISafe(payable(request.safe)).isOwner(signer), NotSafeOwner());
+        // forge-lint: disable-next-item(asm-keccak256)
+        bytes32 structHash = keccak256(
```
rmeissner: request change: this is replayable as no nonce is included in this struct

rmeissner (changes requested):

contracts/src/cosigner/SafenetCosigner.sol
rmeissner: It would be good to document expectations here:
- Do disable Safenet does it only require 1 time approval by the signers without Safenet itself?
- Or should it be a "double opt out" process (so 2 times the signers without Safenet have to approve)?

I was under the assumption it is the "double opt out", because this is how it is with a guard. While not critical for this prototype in a more complete solution I would argue that the "announcement" that the escape hatch is used has very low visibility right now (i.e. only one signer is by default aware).

remedcu: > While not critical for this prototype in a more complete solution I would argue that the "announcement" that the escape hatch is used has very low visibility right now (i.e. only one signer is by default aware).

Yes, but it still requires threshold owner signatures for this to be executed.

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Updated 👍🏾

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Updated 👍🏾

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,335 @@
...
+        Secp256k1.Point memory initialGroupKey,
+        uint256 allowTransactionDelay
+    ) {
+        require(consensusAddress != address(0), InvalidAddress());
+        require(allowTransactionDelay != 0, InvalidParameter());
+        Secp256k1.requireNonZero(initialGroupKey);
+        _CONSENSUS_DOMAIN_SEPARATOR = ConsensusMessages.domain(consensusChainId, consensusAddress);
+        _ALLOW_TX_DELAY = allowTransactionDelay;
+        // forge-lint: disable-next-item(asm-keccak256)
+        _DOMAIN_SEPARATOR = keccak256(
```
nlordell: I think GitHub lost this comment somehow...

request change: Local domain separator cannot be an immutable, as `chainId` can change (for example Ethereum PoW).

contracts/src/cosigner/SafenetCosigner.sol
nlordell: > Yes, but it still requires threshold owner signatures for this to be executed.

What isn't clear to me is _why_: the Guard requires 2 transactions (so the threshold to be reached twice), why not the cosigner? I'm not saying it is the wrong decision, I haven't thought enough about whether or not this makes sense, but the rationale is missing for me in order for me to determine whether or not I agree with the decision.

remedcu: For two steps with a full threshold owner signature, I thought about two approaches:
1. Have a particular message hash signed by each of the owners, and have an accounting system for that built within the smart contract.
2. Have an offline signature aggregation infrastructure for the threshold owner signature to be submitted once all the signers have signed it.

Neither method sounded good to me, so I thought this one was better.

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Updated 👍🏾

remedcu: Using a cached system, as `eq` checks are gas efficient compared to `keccak` recalculation

contracts/src/cosigner/SafenetCosigner.sol
nlordell: 1. Is built into the Safe with `approveHash`
2. Is the Safe transaction service or even Harbour (assuming we extend it for a different signature type)

Generally, infrastructure already exists to accumulate signatures, so I don't agree with downgrading security (owner threshold to agree to using the escape hatch, like for the guard) for this.

For example, would you modify the guard to make it so any owner can schedule a delayed transaction (or at least a delayed transaction to remove the guard)?

rmeissner: I agree, this should not require any additional offchain or onchain code, this can be achieved by either using `checkSignature` or by posting the pre-image of the hash to the cosigner, so that they know what they sign (what I proposed), you could even encode this info into the signature. All of these should work with the existing infrastructure.

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -327,6 +330,16 @@ contract SafenetCosigner is ISignatureValidator {
         );
     }
 
+    function _domainSeparator() private view returns (bytes32) {
+        if (block.chainid == _CACHED_CHAIN_ID) return _CACHED_DOMAIN_SEPARATOR;
```
rmeissner: nit: not sure why we choose a different approach here to our default (where we always calculate it), but since this is in place, can you share how much gas this saves :D ?

contracts/src/cosigner/SafenetCosigner.sol
remedcu: > Is built into the Safe with approveHash

Makes sense 👍🏾 

> Is the Safe transaction service or even Harbour (assuming we extend it for a different signature type)

But the signed message is different to that of a SafeTransaction (it has different parameters). Does the SafeAPI accept a different message schema?

---

Based on the first approach, will update the contract to have full threshold owner signature requirement.

rmeissner: > Does the SafeAPI accept a different message schema

No it won't. So we should decide if we want the escape hatch to work with the API or not

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Initial comment: https://github.com/safe-research/safenet/pull/354#discussion_r3203203753

Cached version cost is a chain read (2), EQ (3) and reading an immutable variable.

While if we calculate it every time, then it costs KECCAK (30), abi.encode for three 32 bytes (3 MSTORE and depending on the data), chain read (2) and address(this) (2)

So, definitely worth this approach.

contracts/src/cosigner/SafenetCosigner.sol
remedcu: As the first approach is still valid (with `approveHash`), I have made the change to require `max(`threshold - 1`, 1)` owner signature (max is selected, because for a safe of 1 of n, the threshold - 1 will lead to removal of all the signers, though SafenetCosigner should not be setup as a 1 of n signer)

remedcu: Updated the PR Description with the current approach, where the `threshold - 1` owner needs to sign to initiate any arbitrary transaction. More details have been added to the PR description (also explaining the previous approach). Let me know if more details need to be added.

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,244 @@
...
+     *      `execTransaction`. Accepts any Safe-compatible signature format (packed ECDSA,
+     *      EIP-1271 contract signature, or pre-approved hash); verification is delegated to
+     *      `ISafe.checkNSignatures`. No Safenet attestation is required, making this available
+     *      even when Safenet is unavailable.
+     *      `safeTx.nonce` must equal the Safe's current nonce at registration time; if other
+     *      transactions advance the nonce before execution, re-registration is required.
+     *      To invalidate a pending registration, advance the Safe nonce with a dummy transaction.
+     */
+    function allowTransaction(SafeTransaction.T calldata safeTx, bytes calldata signature) external {
+        require(safeTx.chainId == block.chainid, InvalidParameter());
```
rmeissner: nit: we are passing in parameters that we load from the chain and the user confirmed via the signature, I don't see a reason to pass them as parameters (as this increases the calldata costs). Here it is easier as it allows us to reuse the type.

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,244 @@
...
+     * @dev Called by Safe during `execTransaction`. `msg.sender` is the Safe.
+     *      Two execution paths:
+     *      1. FROST attestation: `_signature` = `abi.encode(uint64 epoch, FROST.Signature sig)`.
+     *         Verifies the FROST sig against `_hash` under the resolved group key.
+     *      2. Pre-approved transaction: `_signature` is empty. Approves if a matured `allowTransaction`
+     *         registration exists for `(msg.sender, _hash)`. Registrations are not deleted on use
+     *         because this function is `view` — replay is prevented by Safe's own nonce advancing
+     *         after execution.
+     */
+    function isValidSignature(bytes32 _hash, bytes memory _signature) external view override returns (bytes4) {
```
rmeissner: nit: I am not sure why we us `_` parameter names here.

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,244 @@
...
+     *      transactions advance the nonce before execution, re-registration is required.
+     *      To invalidate a pending registration, advance the Safe nonce with a dummy transaction.
+     */
+    function allowTransaction(SafeTransaction.T calldata safeTx, bytes calldata signature) external {
+        require(safeTx.chainId == block.chainid, InvalidParameter());
+        ISafe safeContract = ISafe(payable(safeTx.safe));
+        require(safeTx.nonce == safeContract.nonce(), InvalidNonce());
+        uint256 currentThreshold = safeContract.getThreshold();
+        uint256 requiredSignatures = currentThreshold > 1 ? currentThreshold - 1 : 1;
+        bytes32 safeTxHash = SafeTransaction.hash(safeTx);
```
rmeissner: request change: Aren't we with this essentially at 1 time the threshold has to sign?

The `safeTxHash` here is the same that has to be signed for the transaction to execute the allowed transaction, this means the `signature` can be reused for that one.

rmeissner (dismissed):

contracts/src/cosigner/SafenetCosigner.sol
remedcu: This design decision was specified in the PR description, and I also added the reasoning I had behind it. LMK if that still is not sound.

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Updated 👍🏾

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Updated 👍🏾

contracts/src/cosigner/SafenetCosigner.sol
nlordell: One-time signing is not sound IMO: the safe transaction signature is a permission to execute a transaction and should not be conflated with a permission to skip security checks! You are allowing a single signature to be reused for two things. As a rule of thumb, reusing signing message for two different things is a very bad idea.

This also opens up the possibility where a transaction gets signed while waiting for the Safenet attestation, the attacker then records this signature and triggers the time delay, executing the Safe transaction after the period. This is especially dangerous in a 1/1 (2/2 with the Safenet cosigner) where the you typically sign the transaction to submit it to the transaction service. This would allow any attacker to void Safenet protection after the delay period for any such transaction (unless the user diligently cancels every signed transaction for which they did not get a Safe attestation).

I'm not sure I understand the reluctance of signing twice for the escape hatch - this is the exact same process for the guard which we have determined to be sound. This is not the standard process for doing things with accounts, and the extra friction is a **feature** and not a UX issue IMO.

remedcu: Agreed, updated to have two step process.

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,268 @@
...
+     *         after execution.
+     */
+    function isValidSignature(bytes32 hash, bytes memory signature) external view override returns (bytes4) {
+        if (signature.length > 0) {
+            if (signature.length != 128) return bytes4(0);
+            (uint64 epoch, FROST.Signature memory sig) = abi.decode(signature, (uint64, FROST.Signature));
+            bytes32 message = ConsensusMessages.transactionProposal(_CONSENSUS_DOMAIN_SEPARATOR, epoch, hash);
+            Secp256k1.Point memory groupKey = _resolveGroupKey(epoch);
+            Secp256k1.requireNonZero(groupKey);
+            FROST.verify(groupKey, sig, message);
```
nlordell, line 184: Unrelated to this PR, but I feel like `FROST.verify` should possibly check `groupKey.requireNonZero()` internally.

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,268 @@
...
+    // ============================================================
+
+    function _resolveGroupKey(uint64 epoch) private view returns (Secp256k1.Point memory) {
+        if ($currentEpoch.epoch == epoch) return $currentEpoch.groupKey;
+        if ($previousEpoch.epoch == epoch) return $previousEpoch.groupKey;
+        revert InvalidEpoch();
+    }
+
+    function _domainSeparator() private view returns (bytes32) {
+        if (block.chainid == _CACHED_CHAIN_ID) return _CACHED_DOMAIN_SEPARATOR;
```
nlordell, line 273: nit: this feels like a premature optimization...

nlordell (approved):

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Gas saving comparison: https://github.com/safe-research/safenet/pull/354#discussion_r3207453804

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,268 @@
...
+        uint256 currentThreshold = safeContract.getThreshold();
+        uint256 requiredSignatures = currentThreshold > 1 ? currentThreshold - 1 : 1;
+        // forge-lint: disable-next-line(asm-keccak256)
+        bytes32 structHash = keccak256(abi.encode(_ALLOW_TRANSACTION_TYPEHASH, safeTxHash));
+        // forge-lint: disable-next-line(asm-keccak256)
+        bytes32 messageHash = keccak256(abi.encodePacked("\x19\x01", _domainSeparator(), structHash));
+        safeContract.checkNSignatures(address(0), messageHash, signature, requiredSignatures);
+        require($allowedTransactions[safe][safeTxHash] == 0, TransactionAlreadyAllowed());
+        uint256 executableAt = block.timestamp + _ALLOW_TX_DELAY;
+        $allowedTransactions[safe][safeTxHash] = executableAt;
```
nlordell, line 229: Might be worth commenting that this assumes that Safenet will **never** sign a `TransactionProposal` with a `safeTxHash` that has an `AllowTransaction` preimage.

nlordell (commented):
Can we document on Notion just how tricky the escape hatch design was? You, @rmeissner and myself all had different opinions on how this should work, and I imagine that what we end up writing in the end might be different depending on what is needed for the Wallet interface.

contracts/src/cosigner/SafenetCosigner.sol
remedcu: Updated 👍🏾

contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -0,0 +1,273 @@
...
+     *      `safeTxHash` must embed the Safe's current nonce; if other transactions advance the
+     *      nonce before execution, re-registration is required.
+     *      To invalidate a pending registration, advance the Safe nonce with a dummy transaction.
+     *      The `max(threshold - 1, 1)` requirement rests on the assumption that Safenet will never
+     *      sign a `TransactionProposal` for a `safeTxHash` that has an `AllowTransaction` preimage.
+     *      Such a signature could satisfy the cosigner's own EIP-1271 check within `checkNSignatures`,
+     *      reducing the effective human-owner signature requirement from `threshold - 1` to
+     *      `threshold - 2`.
+     */
+    function allowTransaction(address safe, bytes32 safeTxHash, bytes calldata signature) external {
```
rmeissner, line 218: documentation: I would like that we document that this does not work out of the box with the current Safe Transaction Service. Which is acceptable under the assumption that Safenet allows to disable Safenet (so the escape hatch is really only required if Safenet is down).

remedcu: Done 👍🏾

PR #355, opened 2026-05-04
feat(validator): add build commit to logs and metrics

rmeissner:
## Motivation

When investigating incidents or correlating behavior across validators in Grafana, there was no way to know which build was running. Adding version information closes this gap.

The commit SHA is baked into the Docker image at build time via a `COMMIT_SHA` build arg injected by GitHub Actions (`github.sha`). This means the value is always accurate — it reflects the exact commit the image was built from, with no runtime dependency on git.

## Solution

- Create a version.ts that is populated at build time. By default it will return `dev`
- **Prometheus build info gauge** (`validator_build_info{commit="..."} 1`): the established Prometheus convention for exposing version information. In Grafana this can be displayed as a Stat panel using `label_values(validator_build_info, commit)`, giving a clean at-a-glance widget without any log parsing.

## References

- https://www.robustperception.io/exposing-the-software-version-to-prometheus/

_Generated by [Claude Code](https://claude.ai/code/session_01FYK4zSgARKs74KY87w1ry6)_

### DIFF
--- a/validator/src/utils/logging.ts
+++ b/validator/src/utils/logging.ts
@@ -17,6 +17,7 @@ export type Logger = Pick<WinstonLogger, LogLevel>;
 export type LoggingOptions = {
 	level?: LogLevel | "silent";
 	pretty?: boolean;
+	commit?: string;
 };
 
 const SPLAT = Symbol.for("splat");
@@ -51,10 +52,12 @@ export const createLogger = (options: LoggingOptions): Logger => {
 		options.pretty === true
 			? winston.format.combine(winston.format.timestamp(), winston.format.colorize(), prettyFormat)
 			: winston.format.json({ replacer: jsonReplacer });
+	const defaultMeta = options.commit !== undefined ? { commit: options.commit } : undefined;
 	return winston.createLogger({
 		...level,
 		levels: LEVELS,
 		format,
+		defaultMeta,
 		transports: [new winston.transports.Console()],
 	});
 };
--- a/validator/src/utils/logging.ts
+++ b/validator/src/utils/logging.ts
@@ -17,7 +17,6 @@ export type Logger = Pick<WinstonLogger, LogLevel>;
 export type LoggingOptions = {
 	level?: LogLevel | "silent";
 	pretty?: boolean;
-	commit?: string;
 };
 
 const SPLAT = Symbol.for("splat");
@@ -52,12 +51,10 @@ export const createLogger = (options: LoggingOptions): Logger => {
 		options.pretty === true
 			? winston.format.combine(winston.format.timestamp(), winston.format.colorize(), prettyFormat)
 			: winston.format.json({ replacer: jsonReplacer });
-	const defaultMeta = options.commit !== undefined ? { commit: options.commit } : undefined;
 	return winston.createLogger({
 		...level,
 		levels: LEVELS,
 		format,
-		defaultMeta,
 		transports: [new winston.transports.Console()],
 	});
 };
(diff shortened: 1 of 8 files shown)


nlordell:
question: Personally, I'm not a fan of this approach for piping the data to the runtime (for example, if I run `-e COMMIT_SHA=something_nonsensical`, then my metrics will be incorrect). I would like to understand why this approach was taken over generating a build artifact that includes this information.

Specifically, I think that we should instead add a build step that writes the commit hash to the `dist` directory. While typescript doesn't support this natively, I think it is very easy to achieve using standard tooling:

1. create a `version.ts` module that just exports a `COMMIT_SHA` constant that is initially `dev` or something similar
2. modify our Docker container to `sed -i 's/COMMIT_SHA = "dev"/COMMIT_SHA = "'$(some git command)'"/g' dist/version.js`

This way we build the commit SHA value directly into our build output instead of relying on environment variables (which is not as robust). This is not any more complex.

validator/src/utils/logging.ts
```diff
@@ -51,10 +52,12 @@ export const createLogger = (options: LoggingOptions): Logger => {
 		options.pretty === true
 			? winston.format.combine(winston.format.timestamp(), winston.format.colorize(), prettyFormat)
 			: winston.format.json({ replacer: jsonReplacer });
+	const defaultMeta = options.commit !== undefined ? { commit: options.commit } : undefined;
```
nlordell: request change: I don't think it is a good idea to include this as an additional logging tag for all our logs. It adds a lot of additional data with our logs that is specific to our requirements and (AFAIU) can be achieved must less intrusively.

In particular, I believe we can use:

1. Set `LogOpt=tag='{{ index .Config.Labels "org.opencontainers.image.revision" }}'` ([source](https://docs.podman.io/en/stable/markdown/podman-create.1.html#log-opt-name-value) and [source](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#container-units-container), and we are already populating that label with the commit hash as part of our build process). This will add an additional log tag that will be included with the `journald` logs from our quadlets.
2. Add a relabel rule to create a Loki log label for the custom `journald` log tag

This will get us a similar result (a column with the commit SHA hash in Grafana) but in a way that is less intrusive, doesn't pollute the log line with low cardinality data, gets indexed more efficiently by Loki by nature of being a label.

validator/src/utils/metrics.ts
```diff
@@ -25,9 +26,19 @@ export class MetricsService {
...
 		this.#logger = logger;
 		this.#register = new Registry();
+
+		if (commit !== undefined) {
+			new Gauge({
+				name: "validator_build_info",
+				help: "Build information for the validator",
+				labelNames: ["commit"],
+				registers: [this.#register],
+			}).set({ commit }, 1);
```
nlordell, line 39: This trick, I love!

nlordell (changes requested):
Not a fan of using the environment for providing this data to our runtime, I feel like having our Dockerfile write this information to a file is more robust and just as simple. Additionally, I would prefer to not include commit SHA (which is extremely low cardinality) with our log messages when it can be added in more efficient ways.

validator/src/utils/logging.ts
rmeissner: Tbh I don't see a requirement to have this on the logs, for me the Gauge is sufficient, therefore I would remove this if you think it is too much overhead

nlordell: I think we should. As I mentioned, we can add it to the logs with other less intrusive mechanisms.

nlordell (approved):

PR #356, opened 2026-05-05
Verify and display signature in explorer

rmeissner:
## Motivation

Currently users have to blindly trust that the signature is valid and there is no easy way to extract the signature to use it in a different context (i.e. to compare it with the signature submitted to the Safe API). 

## Solution

- Display Signature as part of the proposal info popup
- Make sure to only display attestations with valid signature
  - Signatures for unknown groups are considered invalid and should not be displayed
- Allow to copy signatures

image


explorer/src/components/transaction/SafeTxProposals.tsx
```diff
@@ -53,6 +54,19 @@ function ProposalInfoButton({ proposal }: { proposal: TransactionProposal }) {
 					<InlineHash hash={status.data.groupId} />
 					<CopyButton value={status.data.groupId} />
 				</div>
+				{status.data.completed && status.data.signature && (
+					<>
+						<span className="text-muted">Signature:</span>
+						<div className="flex items-center gap-1 whitespace-nowrap">
+							<InlineHash
+								hash={`0x${status.data.signature.r.x.toString(16).padStart(64, "0")}${status.data.signature.r.y.toString(16).padStart(64, "0")}${status.data.signature.z.toString(16).padStart(64, "0")}`}
```
rmeissner: The concatenated signature is used multiple time. Extract this signature part into a separate components to avoid duplicating the encoding.

explorer/src/components/transaction/SafeTxProposals.tsx
```diff
@@ -41,6 +42,27 @@ function ProposalInfoButton({ proposal }: { proposal: TransactionProposal }) {
 	);
 	if (status.data === null) return null;
 
+	if (!status.data.completed || !status.data.signature) {
+		return (
+			<InfoPopover trigger={<InformationCircleIcon className="h-4 w-4 text-muted" />}>
```
rmeissner: Now we have code duplication for the signature id and group id. Make sure to reduce code duplication. 

Another way is to make the signatureHex optional (so it can be undefined) and only display the signature section if it is available

explorer/src/lib/coordinator/signing.ts
```diff
@@ -38,6 +40,14 @@ export type AttestationParticipation = {
 	block: bigint;
 };
 
+export type Signature = {
+	r: {
```
rmeissner: Please introduce a FrostPoint type that can be used here and for the public key.

explorer/src/lib/coordinator/signing.ts
```diff
@@ -175,13 +192,101 @@ export const loadLatestAttestationStatus = async ({
 			// If both or none are completed return the recentrly updated one
 			return left.lastUpdate < right.lastUpdate ? 1 : -1;
 		})[0];
+
+	// If we have a completed signing, fetch the group public key for verification
+	const attestationStatus: AttestationStatus | null = rawAttestationStatus
```
rmeissner: it would be cleaner to early return with `null` if `attestationStatus` is null, not completed, or no signature is available. Then load the group key and merge this.

explorer/src/lib/coordinator/signing.ts
```diff
@@ -175,13 +192,101 @@ export const loadLatestAttestationStatus = async ({
...
 type StatusAggregation = {
 	lastUpdate: bigint;
 	selectionRoot?: Hex;
 	committed: AttestationParticipation[];
 	signedBySelection: Record<string, AttestationParticipation[]>;
+	signature?: Signature;
+	groupPublicKey?: { x: bigint; y: bigint };
+};
+
+const GROUP_KEY_ABI = [
```
rmeissner: please move this to the lib/coordinator/abi.ts file, as there are already abi definitions (the events). Please check lib/consensus/abi.ts on how to define the functions

explorer/src/lib/coordinator/signing.ts
```diff
@@ -118,8 +132,8 @@ export const loadLatestAttestationStatus = async ({
 
 	const aggregate = eventLogs.reduce(
 		(agg, log) => {
-			const status = agg[log.args.sid] ?? { committed: [], signedBySelection: {}, lastUpdate: 0n };
```
rmeissner, line 121: `lastUpdate` should be initialized to `0n`, as per type definition this should not be undefined

explorer/src/lib/coordinator/signing.ts
```diff
@@ -118,8 +132,8 @@ export const loadLatestAttestationStatus = async ({
...
-			if (status.lastUpdate < log.blockNumber) {
+			const status = agg[log.args.sid] ?? { committed: [], signedBySelection: {} };
+			if (status.lastUpdate === undefined || status.lastUpdate < log.blockNumber) {
```
rmeissner: `status.lastUpdate` should never be undefined

explorer/src/lib/coordinator/signing.ts
```diff
@@ -175,13 +194,111 @@ export const loadLatestAttestationStatus = async ({
...
+			};
+		}
+	}
+
+	return attestationStatus;
 };
 
 type StatusAggregation = {
-	lastUpdate: bigint;
+	lastUpdate?: bigint;
```
rmeissner: This should not be optional

explorer/src/lib/coordinator/signing.ts
```diff
@@ -175,13 +194,111 @@ export const loadLatestAttestationStatus = async ({
...
 	selectionRoot?: Hex;
 	committed: AttestationParticipation[];
 	signedBySelection: Record<string, AttestationParticipation[]>;
+	signature?: Signature;
+	groupPublicKey?: FrostPoint;
```
rmeissner: groupPublicKey is actually not used. Only to indicate if the signature is valid.

explorer/src/components/transaction/SafeTxProposals.tsx
```diff
@@ -53,6 +60,15 @@ function ProposalInfoButton({ proposal }: { proposal: TransactionProposal }) {
 					<InlineHash hash={status.data.groupId} />
 					<CopyButton value={status.data.groupId} />
 				</div>
+				{signatureHex && (
+					<>
+						<span className="text-muted">Signature:</span>
+						<div className="flex items-center gap-1 whitespace-nowrap">
+							<InlineHash hash={signatureHex} />
```
nlordell: I feel like this component is inaptly named :P

explorer/src/lib/coordinator/signing.ts
```diff
@@ -46,6 +62,7 @@ export type AttestationStatus = {
 	committed: AttestationParticipation[];
 	signed: AttestationParticipation[];
 	completed: boolean;
+	signature?: Signature;
```
nlordell: nit: I feel like this should be:


```suggestion
{
    completed: false;
	signature?: undefined;
} | {
    completed: true;
	signature: Signature;
}
```

I would prefer enforcing at a type level invariants about the data (complete set to `false` with a signature doesn't make sense).

explorer/src/lib/coordinator/signing.ts
```diff
@@ -175,13 +194,65 @@ export const loadLatestAttestationStatus = async ({
 			// If both or none are completed return the recentrly updated one
 			return left.lastUpdate < right.lastUpdate ? 1 : -1;
 		})[0];
+
+	if (!attestationStatus.completed || !attestationStatus.signature) return attestationStatus;
```
nlordell: In particular, this `||` is a code smell because of the looser typing on the attestation data.

explorer/src/lib/coordinator/signing.ts
```diff
@@ -175,13 +194,65 @@ export const loadLatestAttestationStatus = async ({
...
+	const groupKey = await loadGroupPublicKey(provider, coordinator, attestationStatus.groupId);
+	if (groupKey === undefined) return attestationStatus;
+
+	const signature = attestationStatus.signature;
+	try {
+		const isValid = verifySignature(toPoint(signature.r), signature.z, toPoint(groupKey), message);
+		if (!isValid) {
+			// Log and then remove invalid signature
+			console.error(`Detected invalid signature ${signature} for ${attestationStatus.sid}`);
+			attestationStatus.signature = undefined;
```
nlordell: question: What does `complete: true` with an invalid signature mean? I feel like this is an additional error condition that isn't correctly represented type-wise. I added an "error" status suggestion above.

explorer/src/lib/coordinator/signing.ts
nlordell: **Edit**: given the error state below, maybe:

```suggestion
{
    status: "pending" | "error";
	signature?: undefined;
} | {
    statue: "completed";
	signature: Signature;
}
```

explorer/src/lib/coordinator/signing.ts
```diff
@@ -175,13 +194,65 @@ export const loadLatestAttestationStatus = async ({
...
 type StatusAggregation = {
 	lastUpdate: bigint;
 	selectionRoot?: Hex;
 	committed: AttestationParticipation[];
 	signedBySelection: Record<string, AttestationParticipation[]>;
+	signature?: Signature;
+};
+
+export const formatSignatureHex = (signature: Signature): Hex => {
+	return `0x${[signature.r.x, signature.r.y, signature.z].map((v) => v.toString(16).padStart(64, "0")).join("")}`;
```
nlordell: nit: I would prefer using Viem to ABI encode instead of rolling our own.

explorer/src/lib/frost/hashes.ts
```diff
@@ -0,0 +1,23 @@
+import { type H2COpts, hash_to_field } from "@noble/curves/abstract/hash-to-curve.js";
```
nlordell, line 1: What a well-designed library!

explorer/package.json
```diff
@@ -20,8 +20,8 @@
 		"@fontsource-variable/geist-mono": "^5.2.7",
 		"@heroicons/react": "^2.2.0",
 		"@hookform/resolvers": "^5.2.2",
-		"@noble/curves": "^2.0.1",
-		"@noble/hashes": "^2.0.1",
+		"@noble/curves": "^2.2.0",
+		"@noble/hashes": "^2.2.0",
```
nlordell: request change: given our PR etiquette - it isn't clear to me why this was included as part of this PR, I don't think we should be adjusting our dependencies in the same PR as our logic changes (in the spirit of making PRs only do "one thing"). If it is required, I would appreciate an explanation as to why in the description.

nlordell (changes requested):
PR generally looks good to me. I left mostly nits and one minor change request related to our recent PR etiquette.

Generally, I would appreciate a more embellished PR description:
- Why are you verifying signatures in the explorer, how is this information surfaced to the user, and what are they expected do with this information?
- How does the new interface component look like?
- How are error conditions (i.e. invalid signature) expected to be handled, and the motivation behind the decision?

explorer/src/lib/coordinator/signing.ts
rmeissner: Probably signature does not need to be defined at all if we use a more complex datatype

explorer/package.json
rmeissner: reverted

explorer/src/components/transaction/SafeTxProposals.tsx
rmeissner: hmm not sure what a better name is. Maybe `InlineHex` or `ShortenedHex`, would refactor this in a separate PR

nlordell (approved):

PR #359, opened 2026-05-08
Implement SentinelsGame Phase 1: competitive bonded transaction checker oracle

rmeissner:
## Summary

Introduces `CheckerOracle`, a new `IOracle` implementation that enables a permissioned set of checker nodes to competitively post bonded votes ("Approve" or "Deny") on transactions within a time-boxed window. This is Phase 1 of the WardensGame feature, covering core request lifecycle, bonding mechanics, unanimous resolution, and fee distribution.

https://claude.ai/code/session_018T2hGc2dq22SXdf556EFmj


contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,436 @@
...
+            checkerCount: 0,
+            totalScore: 0,
+            approvedOutcome: false,
+            arbitrated: false
+        });
+
+        FEE_TOKEN.safeTransferFrom(msg.sender, address(this), fee);
+
+        emit NewRequest(requestId, msg.sender, fee, bondTarget, deadline);
+    }
```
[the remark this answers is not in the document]
rmeissner: This is intentional and consistent with the spec and the existing oracle pattern. `SimpleOracle` records `msg.sender` (the Consensus contract) as the proposer in exactly the same way. The spec explicitly states `proposer: Consensus contract address` in the `Request` struct definition.

For the fee to flow correctly in practice, the Consensus contract (or whatever calls `postRequest`) must hold a sufficient token balance and have approved this contract to pull it — the same setup requirement as for checker bonds. On a timeout the fee is refunded to the proposer (Consensus), which can then route it back to the user as needed; this keeps the oracle contract decoupled from Safe/user identity.

---

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,436 @@
...
+        require(block.number > req.deadline, VotingWindowOpen());
+
+        bool approveMet = req.totalApproveBond >= req.approveBondTarget;
+        bool denyMet = req.totalDenyBond >= req.approveBondTarget;
+
+        if (approveMet && denyMet) {
+            // Conflict detected — freeze for Phase 2 arbitration. No OracleResult yet.
+            req.state = State.FROZEN;
+            return;
+        }
```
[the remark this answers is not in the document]
rmeissner: This is intentional — the `FROZEN` state is a deliberate holding state designed for Phase 2. Phase 2 adds `triggerArbitration` and `resolveDispute` which unfreeze the request, slash the losing side, and disburse all locked funds. The spec explicitly scopes conflict resolution to Phase 2 and notes that conflicts are expected to be rare with the permissioned checker set. Adding a partial recovery mechanism now would duplicate logic that Phase 2 will implement properly.

---

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,436 @@
...
+    {
+        address[] storage order = $checkerOrder[requestId];
+        uint256 len = order.length;
+        for (uint256 i = 0; i < len; i++) {
+            Commitment storage c = $commitments[requestId][order[i]];
+            if (c.approved == approve) {
+                totalScore += c.bondAmount * (winnerCount + 1 - c.position);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
rmeissner: The spec caps the permissioned checker set at ≤ 20 nodes, so the loop is bounded and inexpensive in practice. The incremental formula (`TotalScore = (winnerCount + 1) × totalBond − Σ bond_i × position_i`) is mathematically correct and is a good optimisation to carry forward, but with a max of ~20 iterations the current approach is readable and safe for V1.

---

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,440 @@
...
+            req.checkerCount = req.denyCheckerCount;
+            req.totalScore = _computeTotalScore(requestId, false, req.denyCheckerCount);
+            emit Resolved(requestId, false, ResolveReason.UNANIMOUS_DENY);
+            emit OracleResult(requestId, req.proposer, abi.encode(ResolveReason.UNANIMOUS_DENY), false);
+        } else {
+            // Timeout / undercapitalised — refund fee to proposer, reject by default.
+            req.approvedOutcome = false;
+            uint256 fee = req.fee;
+            req.fee = 0;
+            FEE_TOKEN.safeTransfer(req.proposer, fee);
```
rmeissner: this should be a pull model, pushing tokens to a receiver is not the best solution, especially if we expect that this is not the actual fee payer (but rather the intermediate consensus contract).

Please add a todo on the fee payment related parts that this has to be properly evaluate how fee payment should happen.

rmeissner: Added a TODO to `postRequest`'s NatDoc noting that the current push-to-receiver model needs evaluation — specifically that the Consensus contract (not the actual fee payer) is both the pull source and the refund target, and that a pull model involving the actual payer may be preferable.

---

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,445 @@
...
+        Commitment storage c = $commitments[requestId][msg.sender];
+        require(c.bondAmount > 0, NothingToClaim());
+        require(!c.claimed, AlreadyClaimed());
+        c.claimed = true;
+
+        uint256 bondReturn = c.bondAmount;
+        uint256 feeReward = 0;
+
+        // Winning-side checkers receive a proportional share of the request fee.
+        if (req.checkerCount > 0 && c.approved == req.approvedOutcome) {
```
rmeissner: Why would we allow to claim when the approved outcome is not the same as the committed response? 

If you answered wrongly, then you should lose your bond.

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,445 @@
...
+    /**
+     * @inheritdoc ICheckerOracle
+     */
+    function scheduleBondMultiplier(uint256 newValue) external override(ICheckerOracle) {
+        require(msg.sender == ARBITRATOR, NotArbitrator());
+        require(newValue > 0, InvalidMultiplier());
+
+        uint256 activeAt = block.number + GOVERNANCE_DELAY;
+        pendingBondMultiplier = newValue;
+        bondMultiplierActiveAt = activeAt;
```
rmeissner: the naming of `bondMultiplierActiveAt` is slightly misleading as it could be understood as an automatic activation. Therefore I would recommend to either implement it as an automatic activation or rename it.

contracts/src/CheckerOracle.sol
rmeissner: Fixed in commit 80eaf21. Losing-side checkers in unanimous outcomes now forfeit their bond — `claim` emits `Claimed(requestId, checker, 0, 0)` and returns without transferring anything. Bonds are still returned in the timeout case (`checkerCount == 0`) since neither side was definitively wrong there.

---

contracts/src/CheckerOracle.sol
rmeissner: Renamed to `pendingBondMultiplierActiveAt` in commit 80eaf21. This pairs clearly with `pendingBondMultiplier` and makes it explicit that it tracks the activation block of the *staged* value rather than implying any automatic behaviour.

---

contracts/src/CheckerOracle.sol
```diff
@@ -261,11 +259,19 @@ contract CheckerOracle is IOracle, ICheckerOracle {
         require(!c.claimed, AlreadyClaimed());
         c.claimed = true;
 
+        bool unanimousOutcome = req.checkerCount > 0;
```
rmeissner: This is super misleading, just because the `checkerCount > 0` it is not a unanimous outcome. this should be renamed to something related to a timeout. 

Also can this state (RESOLVED and checkerCount == 0) even happen?

rmeissner: Good catch — `checkerCount > 0` was indeed not the right condition. Renamed to `isTimeout = req.checkerCount == 0` in commit df70129.

To answer the second question: yes, RESOLVED with `checkerCount == 0` is exactly the timeout path. In `finalize`, when neither side reaches its bond target, we set `state = RESOLVED` but leave `checkerCount` at its zero default (we only assign `req.checkerCount = req.approveCheckerCount` or `req.denyCheckerCount` on a unanimous outcome). So `isTimeout = checkerCount == 0` is the correct way to distinguish the two RESOLVED sub-cases.

---

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,461 @@
...
+        if (approveMet && denyMet) {
+            // Conflict detected — freeze for Phase 2 arbitration. No OracleResult yet.
+            req.state = State.FROZEN;
+            return;
+        }
+
+        req.state = State.RESOLVED;
+
+        if (approveMet) {
+            req.approvedOutcome = true;
```
rmeissner: The outcome can be encoded into the state (i.e. RESOLVED_APPROVED and RESOLVED_DENIED).

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,461 @@
...
+            emit OracleResult(requestId, req.proposer, abi.encode(ResolveReason.UNANIMOUS_APPROVE), true);
+        } else if (denyMet) {
+            req.approvedOutcome = false;
+            req.checkerCount = req.denyCheckerCount;
+            req.totalScore = _computeTotalScore(requestId, false, req.denyCheckerCount);
+            emit Resolved(requestId, false, ResolveReason.UNANIMOUS_DENY);
+            emit OracleResult(requestId, req.proposer, abi.encode(ResolveReason.UNANIMOUS_DENY), false);
+        } else {
+            // Timeout / undercapitalised — refund fee to proposer, reject by default.
+            req.approvedOutcome = false;
```
rmeissner: we should use a separate state here: TIMED_OUT

features/2026_05_08_sentinelsgame_phase1.md
```diff
@@ -0,0 +1,102 @@
+# Feature Proposal: SentinelsGame — Competitive Transaction Checker Oracle (Phase 1)
```
rmeissner: Please remove this feature spec. This is covered by the overall feature spec for the sentinels game

contracts/test/CheckerOracle.t.sol
```diff
@@ -0,0 +1,677 @@
...
+        assertEq(req.proposer, proposer);
+        assertEq(req.fee, REQUEST_FEE);
+        assertEq(req.approveBondTarget, BOND_TARGET);
+        assertEq(uint256(req.state), uint256(ICheckerOracle.State.PENDING));
+    }
+
+    function test_PostRequest_DuplicateReverts() public {
+        _postRequest();
+        vm.expectRevert(CheckerOracle.RequestAlreadyExists.selector);
+        vm.prank(proposer);
```
rmeissner: Why not use _postRequest here again?

contracts/test/CheckerOracle.t.sol
```diff
@@ -0,0 +1,677 @@
...
+import {ICheckerOracle} from "@/interfaces/ICheckerOracle.sol";
+import {CheckerOracle} from "@/CheckerOracle.sol";
+import {MockERC20} from "@test/util/MockERC20.sol";
+
+contract CheckerOracleTest is Test {
+    // ============================================================
+    // CONSTANTS
+    // ============================================================
+
+    uint256 constant REQUEST_FEE = 1000e18;
```
rmeissner: Not sure why such big numbers are used. Most realistic would be to align it with usdc which has 6 decimals and make the base fee 1 cent (So 10000). All other values should be adjusted accordingly.

contracts/test/CheckerOracle.t.sol
```diff
@@ -0,0 +1,677 @@
...
+        uint256 balBefore = token.balanceOf(checker1);
+
+        vm.expectEmit(true, true, false, true);
+        emit ICheckerOracle.Claimed(REQUEST_ID, checker1, 0, 0);
+
+        vm.prank(checker1);
+        oracle.claim(REQUEST_ID);
+
+        // Losing-side checker's bond is slashed — nothing returned.
+        assertEq(token.balanceOf(checker1), balBefore, "losing checker bond should be slashed");
```
rmeissner: Please add a todo for this, as it is questionable if the losing bond should be lost here, since no happened arbitration

contracts/test/CheckerOracle.t.sol
```diff
@@ -277,10 +276,10 @@ contract CheckerOracleTest is Test {
...
         assertEq(uint256(req.state), uint256(ICheckerOracle.State.RESOLVED_APPROVED));
 
-        // Score_1 = 1500e18 / 1 = 1500e18
-        // Score_2 = 500e18  / 2 = 250e18
-        // approveTotalScore = 1750e18
-        uint256 expectedTotalScore = 1500e18 / 1 + 500e18 / 2;
+        // Score_1 = 15_000 / 1 = 15_000
+        // Score_2 = 5_000  / 2 = 2_500
+        // approveTotalScore = 17_500
+        uint256 expectedTotalScore = 15_000 / 1 + 5_000 / 2;
```
rmeissner: Since we know the final numbers (as they are in the comment) why not use them for the assertions

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,431 @@
...
+
+        Commitment storage c = $commitments[requestId][msg.sender];
+        require(c.bondAmount > 0, NothingToClaim());
+        require(!c.claimed, AlreadyClaimed());
+        c.claimed = true;
+
+        bool isTimeout = state == State.TIMED_OUT;
+
+        // On timeout all bonds are returned (no side was definitively wrong).
+        // On a unanimous outcome the losing side forfeits their bond.
```
rmeissner: @nlordell @tschubotz what is your opinion on this? 

The specific case assumes that there is no arbitration as the losing side did not reach the threshold, should this still lead to a loss of the bonds?

tschubotz: Are you saying that 

- Side A reached bond threshold
- Side B had also some bonds, but did NOT reach the bond threshodl

And in that case:
- Side A wins
- Side B loses
- No arbitration
- Side B does NOT lose bonds since threshold not met?

rmeissner: Currently side B does lose the bond. But my question is if this makes sense. Two perspectives:
- Yes makes sense to prevent spam of low amount bonds without this being a punishment
- No does not make sense as no additional work (and therefore no spam) has happened

nlordell: Brief update - We discussed in a sync today that we should never penalize bonds without arbitration.

nlordell:
This PR is very large and, IMO, does not satisfy the "keep changes focused" rule from our PR etiquette. In particular a deployment script is unrelated to the checker oracle implementation, which is itself unrelated to the checker oracle interface proposal.

I would request for this PR to be split into smaller pieces to facilitate review. There were already quite some discussions this morning on implementation details, where we wanted to "merge this PR as is and address the details later", which for me is a symptom of "this PR is doing too much at once".

**Edit**: I would also suggest using libraries as a mechanism to modularize logic in testable and reviewable pieces and allow for large features like this to be split into smaller PRs.

contracts/package.json
```diff
@@ -9,6 +9,7 @@
 		"cmd:propose": "forge script ProposeScript",
 		"cmd:deploy:testing-erc20": "forge script DeployERC20Script",
 		"cmd:deploy:staking": "forge script DeployStakingScript",
+		"cmd:deploy:checker-oracle": "forge script DeployCheckerOracleScript",
```
rmeissner: This should be removed, as this script doesn't exist right now

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,324 @@
...
+
+        FEE_TOKEN.safeTransfer(msg.sender, bondReturn + feeReward);
+        emit Claimed(requestId, msg.sender, bondReturn, feeReward);
+    }
+
+    // ============================================================
+    // BOND MULTIPLIER GOVERNANCE
+    // ============================================================
+
+    function scheduleBondMultiplier(uint256 newValue) external onlyArbitrator {
```
rmeissner: Please move this logic to a separate contract that can be inherited

contracts/src/SentinelManager.sol
```diff
@@ -0,0 +1,91 @@
...
+    error SentinelAlreadyScheduled();
+    error SentinelNotScheduled();
+
+    // ============================================================
+    // MODIFIERS
+    // ============================================================
+
+    // forge-lint: disable-start(unwrapped-modifier-logic)
+
+    modifier onlyArbitrator() {
```
rmeissner: it would be cleaner to handle this in the oracle contract, or a separate contract.

The functions relying on this should be internal and can then be overwritten as external methods with the modifier.

rmeissner:
@nlordell I tried to trim down the code to the bare minimum and also split it into separate contract. The test only test complete flows for now and natspec have been removed (both can be added in separate PRs later). ~Can you please review again~

Edit: I will split into multiple PRs

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,322 @@
...
+
+        if (approveMet && denyMet) {
+            // Conflict detected — freeze for Phase 2 arbitration. No OracleResult yet.
+            req.state = State.FROZEN;
+            return;
+        }
+
+        if (approveMet) {
+            req.state = State.RESOLVED_APPROVED;
+            emit Resolved(requestId, true, ResolveReason.UNANIMOUS_APPROVE);
```
rmeissner: The resolve events feel duplicate and can be removed (as the same information is present in the OracleResult)

contracts/src/BondMultiplierGovernance.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+import {SentinelManager} from "@/SentinelManager.sol";
+
+abstract contract BondMultiplierGovernance is SentinelManager {
```
rmeissner: The inheritance of the SentinelManager should be removed. For the GOVERNANCE_DELAY in the _scheduleBondMultiplier this should be passed in as a parameter.

contracts/src/BondMultiplierGovernance.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+abstract contract BondMultiplierGovernance {
```
rmeissner: To align with the other contract we should convert this to a library with a type `T` which holds the storage in a struct. 

The methods should then be library methods on that type. For reference see https://github.com/safe-research/safenet/blob/main/contracts/src/libraries/FROSTParticipantMap.sol.

Also make sure to move that library contract to the `libraries` folder.

contracts/src/SentinelManager.sol
```diff
@@ -0,0 +1,74 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+abstract contract SentinelManager {
```
rmeissner: Please convert this contract also to a library. GOVERNANCE_DELAY should be passed in as a parameter whereever needed

contracts/src/libraries/SentinelManager.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+library SentinelManager {
```
rmeissner: Please rename it to SentinelMap

contracts/src/libraries/BondMultiplierGovernance.sol
```diff
@@ -0,0 +1,58 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+library BondMultiplierGovernance {
```
rmeissner: Please rename it to BondConfig

contracts/src/libraries/BondMultiplierGovernance.sol
```diff
@@ -0,0 +1,58 @@
...
+    function schedule(T storage self, uint256 newValue, uint256 governanceDelay) internal {
+        require(newValue > 0, InvalidMultiplier());
+
+        uint256 activeAt = block.number + governanceDelay;
+        self.pendingBondMultiplier = newValue;
+        self.pendingBondMultiplierActiveAt = activeAt;
+        emit BondMultiplierScheduled(newValue, activeAt);
+    }
+
+    function applyMultiplier(T storage self) internal {
```
rmeissner: please rename to `applyPending`

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,352 @@
...
+        GOVERNANCE_DELAY = governanceDelay;
+        $bondMultiplier.init(initialMultiplier);
+    }
+
+    // ============================================================
+    // IOracle IMPLEMENTATION
+    // ============================================================
+
+    function postRequest(bytes32 requestId) external override(IOracle) {
+        require($requests[requestId].proposer == address(0), RequestAlreadyExists());
```
rmeissner: Please extract all the request logic into a `SentinelOracleRequest` library that follows the library pattern used for the other types.

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,352 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+import {IERC20} from "@oz/token/ERC20/IERC20.sol";
+import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
+import {IOracle} from "@/interfaces/IOracle.sol";
+import {BondMultiplierGovernance} from "@/libraries/BondMultiplierGovernance.sol";
+import {SentinelManager} from "@/libraries/SentinelManager.sol";
+
+contract CheckerOracle is IOracle {
```
rmeissner: Please rename to SentinelOracle

contracts/src/CheckerOracle.sol
```diff
@@ -0,0 +1,352 @@
...
+    BondMultiplierGovernance.T private $bondMultiplier;
+
+    // forge-lint: disable-next-line(mixed-case-variable)
+    SentinelManager.T private $sentinels;
+
+    // forge-lint: disable-next-line(mixed-case-variable)
+    mapping(bytes32 requestId => Request) private $requests;
+
+    // forge-lint: disable-next-line(mixed-case-variable)
+    mapping(bytes32 requestId => mapping(address sentinel => Commitment)) private $commitments;
```
rmeissner: Please extract all the commitments logic into a `SentinelOracleCommitments` library that follows the library pattern used for the other types.

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -0,0 +1,186 @@
...
+        UNANIMOUS_DENY,
+        TIMEOUT,
+        ARBITRATION
+    }
+
+    // ============================================================
+    // STRUCTS
+    // ============================================================
+
+    struct Request {
```
rmeissner, line 28: I was thinking that this can be further split. One library for the map and one for the request. They can be in the same file.

Same is probably possbile for the commitments.

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -0,0 +1,191 @@
+// SPDX-License-Identifier: GPL-3.0-only
```
rmeissner, line 1: Please rename the file to be plural

rmeissner:
Closed in favor of #374

PR #369, opened 2026-05-13
Sentinel Oracle - Phase 2

rmeissner:
Implement Phase 2 of the Sentinel Game, especially the Arbitration contract logic

https://claude.ai/code/session_01GZUJ34rV4WC1WkvsyQYTod

### DIFF
--- a/contracts/src/SentinelOracle.sol
+++ b/contracts/src/SentinelOracle.sol
@@ -22,6 +22,7 @@ contract SentinelOracle is IOracle {
     // EVENTS
     // ============================================================
 
+    event ArbitrationTriggered(bytes32 indexed requestId);
     event DisputeResolved(bytes32 indexed requestId, SentinelOracleRequest.State outcome, uint256 slashed);
     event Claimed(bytes32 indexed requestId, address indexed sentinel, uint256 bondReturn, uint256 feeReward);
 
@@ -106,13 +107,13 @@ contract SentinelOracle is IOracle {
     // IOracle IMPLEMENTATION
     // ============================================================
 
-    function postRequest(bytes32 requestId) external override(IOracle) {
+    function postRequest(bytes32 requestId, address feePayer) external override(IOracle) {
         require(msg.sender == CONSENSUS, NotConsensus());
         uint256 fee = REQUEST_FEE;
         uint256 bondTarget = fee * $bondConfig.bondMultiplier;
         uint256 deadline = block.number + VOTING_WINDOW;
         $requests.create(requestId, msg.sender, fee, bondTarget, deadline);
-        FEE_TOKEN.safeTransferFrom(msg.sender, address(this), fee);
+        FEE_TOKEN.safeTransferFrom(feePayer, address(this), fee);
     }
 
     // ============================================================
@@ -169,16 +170,16 @@ contract SentinelOracle is IOracle {
     // ============================================================
 
     function triggerArbitration(bytes32 requestId) external onlyArbitrator {
-        SentinelOracleRequest.Request storage req = $requests.get(requestId);
-        req.triggerArbitration();
+        $requests.get(requestId).triggerArbitration();
+        emit ArbitrationTriggered(requestId);
     }
 
     function resolveDispute(bytes32 requestId, bool approveWins) external onlyArbitrator {
         SentinelOracleRequest.Request storage req = $requests.get(requestId);
         address proposer = req.proposer;
-        (uint256 slashed, uint256 fee) = req.resolveDispute(approveWins);
+        uint256 slashed = req.resolveDispute(approveWins);
         SentinelOracleRequest.State outcome = req.state;
-        FEE_TOKEN.safeTransfer(ARBITRATOR, slashed - fee);
+        FEE_TOKEN.safeTransfer(ARBITRATOR, slashed);
         emit DisputeResolved(requestId, outcome, slashed);
         emit OracleResult(requestId, proposer, abi.encode(SentinelOracleRequest.ResolveReason.ARBITRATION), approveWins);
     }
--- a/contracts/src/SentinelOracle.sol
+++ b/contracts/src/SentinelOracle.sol
@@ -22,7 +22,6 @@ contract SentinelOracle is IOracle {
     // EVENTS
     // ============================================================
 
-    event ArbitrationTriggered(bytes32 indexed requestId);
     event DisputeResolved(bytes32 indexed requestId, SentinelOracleRequest.State outcome, uint256 slashed);
     event Claimed(bytes32 indexed requestId, address indexed sentinel, uint256 bondReturn, uint256 feeReward);
 
@@ -168,11 +167,6 @@ contract SentinelOracle is IOracle {
     // ============================================================
     // ARBITRATION
     // ============================================================
-
-    function triggerArbitration(bytes32 requestId) external onlyArbitrator {
-        $requests.get(requestId).triggerArbitration();
-        emit ArbitrationTriggered(requestId);
-    }
 
     function resolveDispute(bytes32 requestId, bool approveWins) external onlyArbitrator {
         SentinelOracleRequest.Request storage req = $requests.get(requestId);
--- a/contracts/src/SentinelOracle.sol
+++ b/contracts/src/SentinelOracle.sol
@@ -107,13 +107,13 @@ contract SentinelOracle is IOracle {
     // IOracle IMPLEMENTATION
     // ============================================================
 
-    function postRequest(bytes32 requestId, address feePayer) external override(IOracle) {
+    function postRequest(bytes32 requestId) external override(IOracle) {
         require(msg.sender == CONSENSUS, NotConsensus());
         uint256 fee = REQUEST_FEE;
         uint256 bondTarget = fee * $bondConfig.bondMultiplier;
         uint256 deadline = block.number + VOTING_WINDOW;
         $requests.create(requestId, msg.sender, fee, bondTarget, deadline);
-        FEE_TOKEN.safeTransferFrom(feePayer, address(this), fee);
+        FEE_TOKEN.safeTransferFrom(msg.sender, address(this), fee);
     }
 
     // ============================================================
--- a/contracts/src/SentinelOracle.sol
+++ b/contracts/src/SentinelOracle.sol
@@ -22,7 +22,7 @@ contract SentinelOracle is IOracle {
     // EVENTS
     // ============================================================
 
-    event DisputeResolved(bytes32 indexed requestId, address winner, address loser, uint256 slashed);
+    event DisputeResolved(bytes32 indexed requestId, SentinelOracleRequest.State outcome, uint256 slashed);
     event Claimed(bytes32 indexed requestId, address indexed sentinel, uint256 bondReturn, uint256 feeReward);
 
     // ============================================================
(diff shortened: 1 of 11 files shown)


contracts/src/interfaces/ICheckerOracle.sol
```diff
@@ -152,9 +152,12 @@ interface ICheckerOracle {
...
 
     /**
-     * @notice Emitted when a dispute is resolved by the arbitrator (Phase 2).
+     * @notice Emitted when a dispute is resolved by the arbitrator.
+     * @param requestId  The resolved request.
+     * @param approveWins True if the Approve side won; false if the Deny side won.
+     * @param slashed    Total bond amount slashed from the losing side.
      */
-    event DisputeResolved(bytes32 indexed requestId, address winner, address loser, uint256 slashed);
+    event DisputeResolved(bytes32 indexed requestId, bool approveWins, uint256 slashed);
```
rmeissner: How about using a status here (can reuse the status of the request)

contracts/test/CheckerOracle.t.sol
```diff
@@ -263,5 +263,56 @@ contract CheckerOracleTest is Test {
...
+        vm.expectRevert(CheckerOracle.NotArbitrator.selector);
+        oracle.triggerArbitration(REQUEST_ID);
+
+        // Arbitrator triggers arbitration.
+        vm.expectEmit(true, false, false, false);
+        emit CheckerOracle.ArbitrationTriggered(REQUEST_ID);
+        vm.prank(arbitrator);
+        oracle.triggerArbitration(REQUEST_ID);
+
+        // Cannot trigger twice.
```
rmeissner: This should not be tested here, as this just tests the whole flow.

contracts/src/CheckerOracle.sol
```diff
@@ -251,6 +255,39 @@ contract CheckerOracle is IOracle, BondMultiplierGovernance {
...
+        Request storage req = $requests[requestId];
+        require(req.proposer != address(0), RequestNotFound());
+        require(req.state == State.FROZEN, RequestNotFrozen());
+        require(req.arbitrated, ArbitrationNotTriggered());
+
+        uint256 slashed = approveWins ? req.totalDenyBond : req.totalApproveBond;
+        req.state = approveWins ? State.RESOLVED_APPROVED : State.RESOLVED_DENIED;
+
+        uint256 fee = req.fee;
+        req.fee = 0;
```
rmeissner: The fee should be refunded from the slashed bond, as the winning parties should still receive the fee. 

If I see this correctly it is currently possible that the winners cannot claim as the contract does not have enough funds.

contracts/src/CheckerOracle.sol
```diff
@@ -251,6 +255,39 @@ contract CheckerOracle is IOracle, BondMultiplierGovernance {
...
+        FEE_TOKEN.safeTransfer(req.proposer, fee);
+        FEE_TOKEN.safeTransfer(ARBITRATOR, slashed);
+
+        emit DisputeResolved(requestId, req.state, slashed);
+        emit Resolved(requestId, approveWins, ResolveReason.ARBITRATION);
```
rmeissner: Theses event feel duplicated, as DisputeResolved provides the same information

contracts/test/CheckerOracle.t.sol
```diff
@@ -263,5 +263,51 @@ contract CheckerOracleTest is Test {
...
+
+        // Fee refunded to proposer; loser (deny) bonds slashed to arbitrator.
+        assertEq(token.balanceOf(proposer), proposerBalBefore + REQUEST_FEE, "fee refunded to proposer");
+        assertEq(token.balanceOf(arbitrator), arbitratorBalBefore + BOND_TARGET, "deny bonds slashed to arbitrator");
+
+        // Winning approve sentinel (sentinel1) gets bond back, no fee reward.
+        uint256 s1Before = token.balanceOf(sentinel1);
+        vm.prank(sentinel1);
+        oracle.claim(REQUEST_ID);
+        assertEq(token.balanceOf(sentinel1), s1Before + BOND_TARGET, "sentinel1 bond returned");
```
rmeissner: the winning party should get the bond and the fee

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -126,6 +129,32 @@ library SentinelOracleRequest {
         uint256 totalScore = state == State.RESOLVED_APPROVED ? self.approveTotalScore : self.denyTotalScore;
         return self.fee * score / totalScore;
     }
+
+    function calcBondReturn(Request storage self, bool approved, uint256 bondAmount, State state)
+        internal
+        view
+        returns (uint256)
+    {
+        if (state == State.TIMED_OUT) return bondAmount;
```
rmeissner: it feels weird that this method returns "wrong" numbers if called on a non resolved request. I understand that this is checked before calling this method, but this looks like something that can easily go wrong. 

Also calculating the bond return amount is not really part of the request logic. The request should just check if the request is eligible to claim the bond.

contracts/src/libraries/SentinelOracleRequests.sol
rmeissner: while not part of this PR but probably should be adjusted: A lot of the caluclation here actually depends on teh commitment state rather than on the request state. Probably should be refactored to be part of the commitment lib.

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -126,6 +129,32 @@ library SentinelOracleRequest {
         uint256 totalScore = state == State.RESOLVED_APPROVED ? self.approveTotalScore : self.denyTotalScore;
         return self.fee * score / totalScore;
     }
+
+    function calcBondReturn(Request storage self, bool approved, uint256 bondAmount, State state)
```
rmeissner: Passing in the state even so it can be loaded from `self` feels very counterintuitive. I understand that this is done to reduce gas usage, but in this case it might make sense to overthink in general the usage of `storage` vs `memory`. For this PR I would say don't worry about gas usage and just access it from storage.

contracts/src/libraries/SentinelOracleRequests.sol
rmeissner: this still has the problem that if the state is not a resolved state then it returns something unexpected. I.e. if the state is FROZEN, then this will return `true` for calls with `approved = true`, which makes no sense

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -132,7 +132,7 @@ library SentinelOracleRequest {
 
     function isBondSlashed(Request storage self, bool approved) internal view returns (bool) {
         if (!self.arbitrated) return false;
-        if (self.state == State.TIMED_OUT) return false;
+        if (self.state != State.RESOLVED_APPROVED && self.state != State.RESOLVED_DENIED) return false;
```
rmeissner: this is also not correct. isBondSlashed should revert if not in a resolved state.

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -131,9 +131,14 @@ library SentinelOracleRequest {
     }
 
     function isBondSlashed(Request storage self, bool approved) internal view returns (bool) {
-        require(self.state == State.RESOLVED_APPROVED || self.state == State.RESOLVED_DENIED, RequestNotResolved());
+        State state = self.state;
```
rmeissner: why not call `requireResolved` it event returns the state for further usage.

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -114,18 +117,40 @@ library SentinelOracleRequest {
...
-    function calcFeeReward(Request storage self, bool approved, uint256 bondAmount, uint256 position, State state)
+    function calcFeeReward(Request storage self, bool approved, uint256 bondAmount, uint256 position)
         internal
         view
         returns (uint256)
     {
-        if (state == State.TIMED_OUT) return 0;
-        bool isWinner = approved == (state == State.RESOLVED_APPROVED);
+        if (self.state == State.TIMED_OUT) return 0;
+        bool isWinner = approved == (self.state == State.RESOLVED_APPROVED);
```
rmeissner: rename `isWinner` to `isEligibleForFee`

contracts/src/SentinelOracle.sol
remedcu, line 115: request change: `Consensus.sol` imports no ERC-20 interfaces, holds no token balance, and never calls `approve()`. `safeTransferFrom(CONSENSUS, ...)` will always revert with `ERC20InsufficientAllowance`, making it impossible to ever create an oracle request.

**Possible fix:** Pull the fee from the originating proposer, not from Consensus. One approach is to pass the payer address through:

```solidity
function postRequest(bytes32 requestId, address feePayer) external {
    ...
    FEE_TOKEN.safeTransferFrom(feePayer, address(this), fee);
}
```

contracts/src/SentinelOracle.sol
```diff
@@ -156,11 +156,31 @@ contract SentinelOracle is IOracle {
...
+    // ============================================================
+    // ARBITRATION
+    // ============================================================
+
+    function triggerArbitration(bytes32 requestId) external onlyArbitrator {
+        SentinelOracleRequest.Request storage req = $requests.get(requestId);
+        req.triggerArbitration();
+    }
+
+    function resolveDispute(bytes32 requestId, bool approveWins) external onlyArbitrator {
```
remedcu, line 171: request change: `resolveDispute` permanently locks `REQUEST_FEE` tokens on every arbitration**

The fee accounting in `resolveDispute` double-counts the fee, leaving `REQUEST_FEE` tokens permanently stuck in the contract after every arbitrated dispute.

**Token trace (REQUEST_FEE = 10, bondTarget = 20, approve wins):**

| Step | Transfer | Contract balance |
|---|---|---|
| `postRequest` | Consensus → contract: 10 | 10 |
| Approve sentinels commit | sentinels → contract: 20 | 30 |
| Deny sentinels commit | sentinels → contract: 20 | **50** |
| `resolveDispute(true)` | contract → ARBITRATOR: `slashed − fee` = 10 | 40 |
| Winning sentinel `claim()` | contract → sentinel: bond(20) + `calcFeeReward(self.fee=10)` = 30 | 10 |
| Losing sentinels `claim()` | 0 (slashed) | **10 - permanently locked** |

**Root cause:** `resolveDispute` retains `fee = self.fee` from the slashed bonds (by sending only `slashed - fee` to the ARBITRATOR) intending to fund the winning sentinel fee reward. But `calcFeeReward` draws from `self.fee` - the *original* `REQUEST_FEE` deposited at request creation, which is a separate pool also still in the contract. Both pools exist simultaneously; only one is paid out. The retained slashed portion has no claimer.

Additionally, the spec defines the arbitration waterfall as: **1. Full user fee refund to proposer. 2. Remainder to ARBITRATOR.** The implementation gives the fee to winning sentinels instead of refunding the user, deviating from the spec.

**Possible fix - aligns with spec (user gets fee refund):**

```solidity
(uint256 slashed, uint256 fee) = req.resolveDispute(approveWins);
FEE_TOKEN.safeTransfer(proposer, fee);             // refund to user
FEE_TOKEN.safeTransfer(ARBITRATOR, slashed - fee); // remainder to arbitrator
// zero out self.fee in resolveDispute so calcFeeReward returns 0 for arbitrated requests
```

contracts/src/SentinelOracle.sol
```diff
@@ -156,11 +156,31 @@ contract SentinelOracle is IOracle {
...
+        uint256 bondReturn = req.isBondSlashed(c.approved) ? 0 : c.bondAmount;
+        FEE_TOKEN.safeTransfer(msg.sender, bondReturn + feeReward);
+        emit Claimed(requestId, msg.sender, bondReturn, feeReward);
+    }
+
+    // ============================================================
+    // ARBITRATION
+    // ============================================================
+
+    function triggerArbitration(bytes32 requestId) external onlyArbitrator {
```
remedcu: request change: `triggerArbitration` sets `arbitrated = true` in storage but emits no event.

The spec explicitly defines `event ArbitrationTriggered(bytes32 indexed requestId)` and states the Phase 3 validator service must *"monitor `Resolved` and `ArbitrationTriggered` events to trigger `claim`."*

contracts/src/SentinelOracle.sol
```diff
@@ -156,11 +156,31 @@ contract SentinelOracle is IOracle {
...
+        SentinelOracleRequest.Request storage req = $requests.get(requestId);
+        req.triggerArbitration();
```
remedcu: nit:
```suggestion
        $requests.get(requestId).triggerArbitration();
```

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -114,18 +117,40 @@ library SentinelOracleRequest {
...
+        self.arbitrated = true;
+    }
+
+    function resolveDispute(Request storage self, bool approveWins) internal returns (uint256 slashed, uint256 fee) {
+        require(self.state == State.FROZEN, RequestNotFrozen());
+        require(self.arbitrated, ArbitrationNotTriggered());
+        slashed = approveWins ? self.totalDenyBond : self.totalApproveBond;
+        self.state = approveWins ? State.RESOLVED_APPROVED : State.RESOLVED_DENIED;
+        fee = self.fee;
+        return (slashed, fee);
```
remedcu: nit: explicit `return` not required.

```suggestion
```

remedcu (changes requested):

contracts/src/SentinelOracle.sol
rmeissner: This is explicitly mentioned for phase 4 (https://github.com/safe-research/safenet/blob/stack/sentinels_game/features/2026_05_05_transaction_checker_oracle_v1.md#phase-4--final-cleanup-pr-4-depends-on-previous-phases) as this requires adjustments to the consensus contract, which I would not do here.

@claude please revert the adjustments and add a comment to phase 4 of the feature specs

contracts/src/SentinelOracle.sol
rmeissner: The specs for phase 3 are inaccurate. `claim` should only be triggered after a dispute has been resolved. @claude please adjust the specs

contracts/src/SentinelOracle.sol
rmeissner: @claude please revert the adjustments and add a comment to phase 4 of the feature specs as this would adjust the current oracle interface. 

Note: might have to be adjusted in some form already in phase 3 for integration tests.

contracts/src/SentinelOracle.sol
rmeissner: @claude why do we need `triggerArbitration` anyways?

the FROZEN state is already an indicator for that arbitration is required. FROZEN and arbitration behave in sync and therefore we can remove the arbitration flag and the `triggerArbitration` step.

remedcu (approved):
Based on internal discussion, approving in favor of a new PR to handle the fee handling.

PR #371, opened 2026-05-13
Define libs over inheritance for claude

rmeissner:
### Context and Motivation

We want to modularise code to make it is to review and reuse. In our current code we use Libraries with structs for this, and Claude should follow this standard.

### Decisions and Tradeoffs

- We want to apply this pattern for all our contracts whenever possible and prefer this over inheritance.
- Update CLAUDE.md to explain this pattern and mention that it should be used over inheritance.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -114,4 +114,4 @@ Always use a name that makes the purpose of the branch immediately clear to anyo
 
 #### Use Libraries over inheritance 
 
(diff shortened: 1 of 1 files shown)


nlordell (approved):
For reference, This is an OCaml-ism that I brought over to Solidity.

remedcu (approved):

PR #373, opened 2026-05-15
fix: add legacy EIP-1271 interface support to SafenetCosigner

remedcu:
Add legacy EIP-1271 interface overload to `SafenetCosigner` to support Safe 1.4.1.

### Context and Motivation

Safe 1.4.1 (and other older Safe versions) was compiled against the original `ISignatureValidator` interface, which defines `isValidSignature(bytes,bytes)` with selector `0x20c13b0b`. The modern EIP-1271 standard uses `isValidSignature(bytes32,bytes)` with selector `0x1626ba7e`. `SafenetCosigner` only implemented the modern interface, so when `execTransaction` on a Safe 1.4.1 instance called the cosigner for signature validation, it received no magic value back and rejected the signature — causing a bare revert with no error data.

This PR adds the legacy overload so the cosigner works across both Safe versions. The `_data` argument in the legacy interface is the raw EIP-712 encoding (`\x19\x01 || domainSeparator || structHash`, 66 bytes), whose `keccak256` equals the `safeTxHash` — so verification logic is identical; only the input hashing step differs.

### Decisions and Tradeoffs

- The verification logic shared between both overloads was extracted into a `_isVerified(bytes32, bytes)` private helper to avoid duplication and keep both public functions thin.
- The legacy overload returns `bytes4(0x20c13b0b)` on success (not the `EIP1271_MAGIC_VALUE` constant `0x1626ba7e`), because Safe 1.4.1 checks for the old selector. Returning the wrong magic value would silently fail validation.

### DIFF
--- a/contracts/src/cosigner/SafenetCosigner.sol
+++ b/contracts/src/cosigner/SafenetCosigner.sol
@@ -58,6 +58,8 @@ contract SafenetCosigner is ISignatureValidator {
     // CONSTANTS
     // ============================================================
 
+    bytes4 private constant LEGACY_EIP1271_MAGIC_VALUE = bytes4(0x20c13b0b);
+
     // forge-lint: disable-next-item(asm-keccak256)
     bytes32 private constant _EIP712_DOMAIN_TYPEHASH =
         keccak256("EIP712Domain(uint256 chainId,address verifyingContract)");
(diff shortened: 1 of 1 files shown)


contracts/src/cosigner/SafenetCosigner.sol
```diff
@@ -58,6 +58,8 @@ contract SafenetCosigner is ISignatureValidator {
     // CONSTANTS
     // ============================================================
 
+    bytes4 private constant LEGACY_EIP1271_MAGIC_VALUE = bytes4(0x20c13b0b);
```
nlordell: request change: 
```suggestion
    bytes4 private constant _LEGACY_EIP1271_MAGIC_VALUE = bytes4(0x20c13b0b);
```

nlordell (approved):

nlordell:
Optimistically approving to reduce back and forth.

PR #374, opened 2026-05-16
SentinelOracle - Phase 1

rmeissner:
## Summary

Introduces `SentinelOracle`, a new `IOracle` implementation that enables a permissioned set of checker nodes to competitively post bonded votes ("Approve" or "Deny") on transactions within a time-boxed window. This is Phase 1 of the SentinelOracle feature, covering core request lifecycle, bonding mechanics, unanimous resolution, and fee distribution.

## Specs

[Feature Spec Phase 1](https://github.com/safe-research/safenet/blob/main/features/2026_05_05_transaction_checker_oracle_v1.md#phase-1--core-contract-pr-1)

The specs have been slightly adjusted to use `Sentinels` instead of `Transaction Checkers`.

## Review Notes

The PR is separated into each commit that can be independently reviewed (in order):
1. e85bc3dda9705f18c8311aefe2cf1251022f01ea - Library to manage bond configuration
2. abcdec7888405922ec6e5f852c5caa7653330353 - Library to manage Sentinels
3. d4f0a937dfd6202e93043ab3097fd40126fa0dc9 - Library to manage requests
4. 91a10ad8b9f1ebe1b8b4af36d439bc5ef2ab66d6 - Library to manage commitments
5. 18eba26b0b49c6b6be61e86752cc75d6e8e12073 - Oracle contract that uses the library contracts
6. e00db1dc840e6cc12f0c80a666824ab072fde707 - Test for flows of the oracle

Any additional commits are adjustments based on comments.

https://claude.ai/code/session_018T2hGc2dq22SXdf556EFmj


contracts/test/SentinelOracle.t.sol
```diff
@@ -258,9 +261,254 @@ contract SentinelOracleTest is Test {
...
+        vm.expectEmit(true, false, false, false);
+        emit SentinelOracle.ArbitrationTriggered(REQUEST_ID);
+
         oracle.finalize(REQUEST_ID);
 
         SentinelOracleRequest.Request memory req = oracle.getRequest(REQUEST_ID);
         assertEq(uint256(req.state), uint256(SentinelOracleRequest.State.FROZEN), "conflicted request should be frozen");
     }
+
+    // ============================================================
```
rmeissner: This PR should only have e2e flow tests, to minimize the reviewable code. Please remove any other tests.

contracts/src/SentinelOracle.sol
```diff
@@ -131,7 +131,10 @@ contract SentinelOracle is IOracle {
         address proposer = req.proposer;
         (SentinelOracleRequest.State newState, uint256 refundFee) = req.finalize();
 
-        if (newState == SentinelOracleRequest.State.FROZEN) return;
+        if (newState == SentinelOracleRequest.State.FROZEN) {
+            emit ArbitrationTriggered(requestId);
```
rmeissner: This should be handled in the Phase 2 PR. I would remove the event for now.

contracts/src/SentinelOracle.sol
```diff
@@ -0,0 +1,225 @@
...
+        VOTING_WINDOW = votingWindow;
+        GOVERNANCE_DELAY = governanceDelay;
+        $bondConfig.init(initialMultiplier);
+    }
+
+    // ============================================================
+    // IOracle IMPLEMENTATION
+    // ============================================================
+
+    function postRequest(bytes32 requestId) external override(IOracle) {
```
remedcu, line 109: request change: `SentinelOracle.postRequest` can be called by any address, not just `Consensus`. The `requestId` is fully deterministic from public on-chain data:

```
message = domainSeparator().oracleTransactionProposal(epochs.active, oracle, safeTxHash)
```

An attacker can front-run any `proposeOracleTransaction` call, register the request with themselves as proposer, and block the Consensus call from ever succeeding (it reverts with `RequestAlreadyExists`). After the voting window the attacker calls `finalize()`, the request times out, and the `REQUEST_FEE` is **refunded to them**. Net cost: gas only. The `(epoch, oracle, safeTxHash)` slot is permanently occupied for the current epoch.

Suggested fix — restrict `postRequest` to the Consensus contract:

```solidity
address public immutable CONSENSUS;
error NotConsensus();

function postRequest(bytes32 requestId) external override {
    ...
}
```

`CONSENSUS` set at construction time alongside `ARBITRATOR`.

contracts/src/SentinelOracle.sol
```diff
@@ -0,0 +1,225 @@
...
+        uint256 initialMultiplier
+    ) {
+        require(arbitrator != address(0), InvalidAddress());
+        require(feeToken != address(0), InvalidAddress());
+        require(requestFee > 0, ZeroFee());
+        ARBITRATOR = arbitrator;
+        FEE_TOKEN = IERC20(feeToken);
+        REQUEST_FEE = requestFee;
+        VOTING_WINDOW = votingWindow;
+        GOVERNANCE_DELAY = governanceDelay;
```
remedcu, line 101: nit:
```suggestion
        require(votingWindow > 0, InvalidVotingWindow());
        require(governanceDelay > 0, InvalidGovernanceDelay());
        ARBITRATOR = arbitrator;
        FEE_TOKEN = IERC20(feeToken);
        REQUEST_FEE = requestFee;
        VOTING_WINDOW = votingWindow;
        GOVERNANCE_DELAY = governanceDelay;
```

A misconfiguration with zero value is more easy to make compared to an intentional small value.

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -0,0 +1,191 @@
...
+
+    function calcFeeReward(Request storage self, bool approved, uint256 bondAmount, uint256 position, State state)
+        internal
+        view
+        returns (uint256)
+    {
+        if (state == State.TIMED_OUT) return 0;
+        bool isWinner = approved == (state == State.RESOLVED_APPROVED);
+        if (!isWinner) return 0;
+        uint256 score = (bondAmount * 1e18) / position;
```
remedcu, line 125: feedback: Design feedback, not a bug: A sentinel with better transaction-evaluation quality but limited ability to win gas auctions loses disproportionately to one with MEV expertise or a direct block-builder relationship, independent of how accurately they assess the transaction. Over time this could concentrate fee earnings among the most technically sophisticated operators rather than the most security-conscious ones.

remedcu (commented):
The ~`ICheckerOracle.sol`~ `ISentinelOracle.sol` and the deployment script is missing from the Phase 1 as per spec.

contracts/src/SentinelOracle.sol
rmeissner: Either that or we should have the proposer be part of the hash (which I would prefer)

contracts/src/SentinelOracle.sol
rmeissner: I would argue it is similarly bad and would not require any checks. As these are immutable they should be checked after deployment in any case

contracts/src/libraries/SentinelOracleRequests.sol
rmeissner: This is a known issue and has been discussed. As a solution we are planning to evaluate secret voting (this is documented on notion)

contracts/script/DeploySentinelOracle.s.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: GPL-3.0-only
```
rmeissner: Please remove the deployment script (to keep the PR small), as this will be added in a later step. Please adjust the feature spec and add a phase 4

contracts/src/SentinelOracle.sol
remedcu: Either way is fine, the aim is to avoid this type of attack.

remedcu (changes requested):
`request change` comment updated.

contracts/src/SentinelOracle.sol
rmeissner: Implemented the hash-based approach per @rmeissner's preference — `postRequest` is now permissionless and `CONSENSUS` has been removed. The front-run protection will be enforced at the Consensus layer: the Consensus contract will compute `requestId = keccak256(abi.encode(transactionData, address(this)))`, making it impossible for an attacker to pre-register the same ID from a different address.

---

rmeissner: @claude Now the change was just reverted, this is not a solution. As the implementation of using the proposer for the mapping key introduces other pitfalls, we should stick with limiting to a specific Consensus contract (as originally proposed by the comment)

contracts/src/interfaces/ISentinelOracle.sol
```diff
@@ -0,0 +1,64 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+import {IOracle} from "@/interfaces/IOracle.sol";
+import {SentinelOracleRequest} from "@/libraries/SentinelOracleRequests.sol";
+import {SentinelOracleCommitment} from "@/libraries/SentinelOracleCommitments.sol";
+
+interface ISentinelOracle is IOracle {
```
rmeissner: Please remove the interface, it provides no value right now and can be introduced later once the required.

remedcu (approved):

PR #380, opened 2026-05-19
SentinelOracle - Phase 3: Introduce transaction manager

rmeissner:
### Context and Motivation

Task in Phase 3 of the Sentinel Oracle ([Feature Spec](https://github.com/safe-research/safenet/blob/main/features/2026_05_05_transaction_checker_oracle_v1.md#phase-3--checker-node-service-pr-3-can-be-parallelized-with-phase-2)). Complete list of tasks can be found in the [milestone](https://github.com/safe-research/safenet/milestone/1)

This extracts the logic of transaction management from the safenet onchain protocol class into a separate reusable class. This way this logic can be reused for the sentinel service to monitor  and retry transactions.

PR #381, opened 2026-05-19
SentinelOracle - Phase 3: Introduce Sentinel Protocol

rmeissner:
### Context and Motivation

Task in Phase 3 of the Sentinel Oracle ([Feature Spec](https://github.com/safe-research/safenet/blob/main/features/2026_05_05_transaction_checker_oracle_v1.md#phase-3--checker-node-service-pr-3-can-be-parallelized-with-phase-2)). Complete list of tasks can be found in the [milestone](https://github.com/safe-research/safenet/milestone/1)

The protocol is responsible for converting actions to transactions and
making sure that these transactions get executed onchain. For this the
transaction manager and base action queue are reused.


validator/src/sentinel/protocol.ts
```diff
@@ -0,0 +1,106 @@
...
+
+	protected async performAction(action: SentinelAction): Promise<SubmittedAction> {
+		switch (action.id) {
+			case "sentinel_approve_token":
+				return this.#txManager.submitAction({
+					to: this.#feeToken,
+					data: encodeFunctionData({
+						abi: ERC20_FUNCTIONS,
+						functionName: "approve",
+						args: [this.#sentinelOracle, action.bondTarget],
```
remedcu: nit: Is `bondTarget` required here, or `bondAmount` is fine?

validator/src/sentinel/protocol.ts
```diff
@@ -0,0 +1,106 @@
+import type { Database } from "better-sqlite3";
+import { type Address, encodeFunctionData } from "viem";
+import { z } from "zod";
+import { BaseActionQueue, type SubmittedAction } from "../consensus/protocol/base.js";
+import type { TransactionManager } from "../consensus/protocol/transaction.js";
```
remedcu, line 5: Found with AI, and I think it makes sense:

**Concurrent `submitAction` calls are now reachable — `protocol.ts:43`, `transaction.ts:192`**

`SentinelProtocol` and `OnchainProtocol` share one `TransactionManager` but have independent `#currentAction` guards. Both queues can call `submitAction` simultaneously. This makes the previously-safe silent-dequeue branch reachable:

```ts
// transaction.ts:199
return { nonce, hash: null }; // no throw → action dequeued as "submitted"
```

If `OnchainProtocol` and `SentinelProtocol` both submit concurrently and one fails mid-flight, the failing action is dequeued silently. The Ethereum tx is retained in storage and retried by `#checkPendingActions` — but with no protocol-level `validUntil` applied. For `attest_transaction` or `sign_request`, a late on-chain arrival after the ceremony window closes wastes the nonce and produces an unrecorded attestation.

This was safe with a single consumer (PR #380). It is now load-bearing.

validator/src/sentinel/abis.ts
```diff
@@ -0,0 +1,42 @@
+import { parseAbi, parseAbiItem } from "viem";
+
+export const SENTINEL_NEW_REQUEST_EVENT = parseAbiItem(
+	"event NewRequest(bytes32 indexed requestId, address indexed proposer, uint256 fee, uint256 bondTarget, uint256 deadline)",
+);
+
+export const SENTINEL_ORACLE_RESULT_EVENT = parseAbiItem(
```
remedcu: Are these the same: https://github.com/safe-research/safenet/blob/feat/sentinel_protocol/validator/src/types/abis.ts#L19-L21

validator/src/sentinel/protocol.ts
rmeissner: validator and sentinel should be run as separate services with separate keys. They currently live in the same code base as they share a lot of code and doing a bigger refactor to create a shared code package is quite involved for a PoC, especially considering that we want to migrate the services away from Typescript.

remedcu (approved):
Approving, as the rest 2 are small nits.

PR #388, opened 2026-05-20
Explicit Tx Rejection Spec

remedcu:
Adds a feature specification for explicit transaction rejection, replacing the current silent validator drop with an on-chain decline that the explorer surfaces as a distinct `REJECTED` status.

### Context and Motivation

- Currently, when a validator rejects a transaction as invalid it silently ignores it. The signing ceremony times out with no record. From the operator's perspective, a rejected transaction looks identical to a network fault or timeout.
- This spec defines the full feature: how the decline is recorded on-chain, how it propagates to `Consensus`, how the validator state machine handles it, and how the explorer surfaces it.

### Decisions and Tradeoffs

- **Decline lives in `FROSTCoordinator`, not `Consensus`**: a decline is a signing-protocol concept ("I am not participating in this ceremony"), not a consensus-layer one. The coordinator already owns all per-participant signing state.
- **Threshold-based stopping, not advisory**: `signShare` reverts once `count - threshold + 1` validators have declined, that is the exact point at which the ceremony is mathematically uncompletable. Before that threshold, signing is unaffected. Attestation is never blocked regardless of declines.
- **Callback pattern mirrors `signShareWithCallback` → `onSignCompleted`**: `signDeclineWithCallback` triggers `onSignRejected` on `Consensus` when the threshold is first crossed, emitting `TransactionRejected` on the same contract as `TransactionProposed` and `TransactionAttested`. This keeps the explorer's status derivation to a single contract with no cross-contract SID correlation.
- **`signDeclineWithCallback` deferred to Phase 2**: it calls `callback.target.onSignRejected` which does not exist on `IFROSTCoordinatorCallback` until Phase 2. The coordinator function and the interface method must be introduced together.
- **No "reason" code**: avoids keeping a Solidity/TypeScript enum in sync and a contract upgrade path each time a new validation check is added.
- **5 PRs**: Phase 1 (coordinator), Phase 2 (consensus callback), Phase 3 (validator), Phase 4 (explorer status), Phase 5 (explorer declined breakdown). Phases 4 and 5 are split because they touch entirely separate data sources (`Consensus` vs `FROSTCoordinator`) and can be reviewed independently.

### DIFF
--- a/features/2026_05_19_explicit_transaction_rejection.md
+++ b/features/2026_05_19_explicit_transaction_rejection.md
@@ -1,0 +1,409 @@
+# Feature Proposal: Explicit Transaction Rejection Response
+Component: `all`
+
+---
+
+## Overview
+
+Currently, when a validator determines a transaction is invalid, it silently drops it. The FROST signing ceremony times out with no on-chain record. From the user's and operator's perspective, "rejected" and "error/timeout" are indistinguishable.
+
+This feature adds an explicit on-chain rejection mechanism: a validator that determines a transaction is invalid submits a `rejectTransaction` (or `rejectOracleTransaction`) call to `Consensus.sol`, storing the rejection reason on-chain. The explorer surfaces this as a distinct "Rejected" status with the specific reason code.
+
+**Phases (separate PRs):**
+1. **Contracts** — Add `rejectTransaction`, `rejectOracleTransaction`, `getTransactionRejection`, `getOracleTransactionRejection`, `TransactionRejected`, `OracleTransactionRejected` events, and `RejectionReason` enum.
+2. **Validator** — Emit `consensus_reject_transaction` / `consensus_reject_oracle_transaction` action on invalid transactions instead of silently returning.
+3. **Explorer** — Query `TransactionRejected` / `OracleTransactionRejected` events and display rejection status with reason.
+
+Phases 2 and 3 can be developed in parallel once Phase 1 is merged.
+
+---
+
+## Architecture Decision
+
+### On-chain rejection over off-chain signaling
+
+Rejection is recorded on-chain via new `rejectTransaction` / `rejectOracleTransaction` functions rather than via off-chain storage or P2P messages. This is consistent with the existing architecture where all protocol communication happens through contract events, and it ensures transparency and verifiability.
+
+### Rejection is advisory — attestation is never blocked
+
+A rejection event provides fast feedback and observability, but **does not prevent a subsequent attestation**. `attestTransaction` is not modified. If an attestation is recorded for a previously rejected proposal, the explorer shows `ATTESTED` (attestation takes precedence).
+
+**Why**: The allowed-modules, allowed-guards, and allowed-fallback-handler lists are hardcoded per validator binary version (see `checks/config/modules.ts`, `guards.ts`, `fallback.ts`). During a rolling upgrade where some validators run a newer version with an expanded allowlist, a validator on the old version could reject a transaction that the majority considers valid. Making rejection block attestation would give a single outdated validator the power to permanently deny any transaction in that epoch, with no override mechanism. Keeping rejection advisory eliminates this DoS vector entirely.
+
+### First validator wins (race-to-reject)
+
+The first validator to call `rejectTransaction` establishes the on-chain record; subsequent calls revert with `AlreadyRejected`. This keeps storage and gas costs to a single slot and a single tx per proposal.
+
+`rejectTransaction` still guards against rejecting an already-attested proposal (`require($attestations[message].isZero(), AlreadyAttested())`) to avoid emitting misleading rejection events after a successful attestation.
+
+### Access control: FROST group participants only
+
+`rejectTransaction` calls `_COORDINATOR.participantKey($groups[epoch], msg.sender)`, which reverts with `InvalidParticipant` (from `FROSTParticipantMap.getKey`) if the caller is not a member of the epoch's signing group. This revert propagates and serves as the access control guard. No return-value check is needed.
+
+### Error codes as a `uint8` enum
+
+Rejection reasons are stored on-chain as a `uint8` enum (`RejectionReason`), matching the current set of `TransactionCheckErrorCode` values in the validator. Adding new codes requires a contract upgrade.
+
+### Oracle transaction parity
+
+`proposeOracleTransaction` has a parallel lifecycle with a different message derivation (includes oracle address). This feature adds `rejectOracleTransaction` to provide the same rejection visibility for oracle-checked proposals.
+
+### Alternatives Considered
+
+**Off-chain validator database**: Validators could write rejections to a shared or local DB for the explorer to query. Rejected because it breaks the transparency guarantee and requires additional infrastructure.
+
+**Threshold-based rejection**: Require M-of-N validators to reject before marking as rejected. Would handle version-mismatch cases more gracefully, but adds significant complexity (per-validator storage, threshold parameter). Advisory rejection achieves the same safety with less complexity.
+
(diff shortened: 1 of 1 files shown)


features/2026_05_19_explicit_transaction_rejection.md
```diff
@@ -0,0 +1,409 @@
+# Feature Proposal: Explicit Transaction Rejection Response
+Component: `all`
+
+---
+
+## Overview
+
+Currently, when a validator determines a transaction is invalid, it silently drops it. The FROST signing ceremony times out with no on-chain record. From the user's and operator's perspective, "rejected" and "error/timeout" are indistinguishable.
+
+This feature adds an explicit on-chain rejection mechanism: a validator that determines a transaction is invalid submits a `rejectTransaction` (or `rejectOracleTransaction`) call to `Consensus.sol`, storing the rejection reason on-chain. The explorer surfaces this as a distinct "Rejected" status with the specific reason code.
```
rmeissner: > storing the rejection reason on-chain

We don't really need a reason, just an flag is enough.

features/2026_05_19_explicit_transaction_rejection.md
```diff
@@ -0,0 +1,409 @@
...
+
+**Phases (separate PRs):**
+1. **Contracts** — Add `rejectTransaction`, `rejectOracleTransaction`, `getTransactionRejection`, `getOracleTransactionRejection`, `TransactionRejected`, `OracleTransactionRejected` events, and `RejectionReason` enum.
```
rmeissner: Hmm this implies that the rejection is tracked on the consensus, I would have expected it on the coordinator. Can we outline the tradeoffs?

features/2026_05_19_explicit_transaction_rejection.md
```diff
@@ -0,0 +1,409 @@
...
+
+### Rejection is advisory — attestation is never blocked
+
+A rejection event provides fast feedback and observability, but **does not prevent a subsequent attestation**. `attestTransaction` is not modified. If an attestation is recorded for a previously rejected proposal, the explorer shows `ATTESTED` (attestation takes precedence).
+
+**Why**: The allowed-modules, allowed-guards, and allowed-fallback-handler lists are hardcoded per validator binary version (see `checks/config/modules.ts`, `guards.ts`, `fallback.ts`). During a rolling upgrade where some validators run a newer version with an expanded allowlist, a validator on the old version could reject a transaction that the majority considers valid. Making rejection block attestation would give a single outdated validator the power to permanently deny any transaction in that epoch, with no override mechanism. Keeping rejection advisory eliminates this DoS vector entirely.
+
+### First validator wins (race-to-reject)
+
+The first validator to call `rejectTransaction` establishes the on-chain record; subsequent calls revert with `AlreadyRejected`. This keeps storage and gas costs to a single slot and a single tx per proposal.
```
rmeissner: I would expect this to be a per validator tracking (on coordinator), otherwise you need to allow that a rejection can be overwritten with a valid attestation.

features/2026_05_19_explicit_transaction_rejection.md
```diff
@@ -0,0 +1,409 @@
...
+
+Both return `(address(0), RejectionReason.NotRejected)` when not rejected.
+
+#### `attestTransaction` — no change
+
+`attestTransaction` is **not** modified. Rejection is advisory and does not block attestation.
+
+#### Test cases
+
+- Valid tx: can be attested; can also be rejected after attestation — second call reverts with `AlreadyAttested`.
```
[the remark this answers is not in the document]
remedcu: Updated 👍🏾

features/2026_05_19_explicit_transaction_rejection.md
```diff
@@ -0,0 +1,409 @@
...
+#### Updated `BaseProtocol` (`base.ts`)
+
+- Add `case "consensus_reject_transaction"` and `case "consensus_reject_oracle_transaction"` to `performAction`.
+- Add abstract methods `rejectTransaction` and `rejectOracleTransaction`.
+
+#### Updated `OnchainProtocol` (`onchain.ts`)
+
+Implement both methods by calling the corresponding contract functions.
+
+#### `AlreadyRejected` handling (Issues 3 & 6)
```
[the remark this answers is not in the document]
remedcu: Updated 👍🏾

features/2026_05_19_explicit_transaction_rejection.md
remedcu: Updated 👍🏾

features/2026_05_19_explicit_transaction_rejection.md
remedcu: That is what was proposed initially, i.e. any validator can suggest rejection, but if there are enough FROST signers, then attestation has higher precedence.

But, changed it to be per validator for now.

features/2026_05_19_explicit_transaction_rejection.md
remedcu: Added a section specifying why rejection is stored on Consensus.

features/2026_05_19_explicit_transaction_rejection.md
```diff
@@ -0,0 +1,337 @@
...
+### Per-validator tracking
+
+Each validator's rejection is stored as an individual flag. This means:
+- Multiple `TransactionRejected` events can be emitted for the same proposal, one per rejecting validator.
+- Rejection and attestation coexist naturally — no "first wins" special-casing needed.
+- The explorer can show which validators rejected, providing richer observability.
+
+### Rejection stored on `Consensus`, not `FROSTCoordinator`
+
+Rejection is a consensus-layer concern (is this Safe transaction valid per the network's rules?) and belongs alongside attestation in `Consensus.sol`. The `FROSTCoordinator` manages low-level FROST signing mechanics and has no knowledge of Safe transactions or validator validity rules. Placing rejection there would introduce a semantic mismatch and couple the signing infrastructure to application-layer policy. Keeping both attestation and rejection in `Consensus` also maintains a consistent and complete picture of a proposal's lifecycle in one contract.
```
rmeissner: > Rejection is a consensus-layer concern

This is not true. A rejection is a explicit "I do not participate in the signing process". The signing process is managed in the coordinator, not in the consensus.

> belongs alongside attestation

Attestations are stored in the coordinator. The consensus only stored a pointer (the signature id).

remedcu: Updated 👍🏾

features/2026_05_19_explicit_transaction_rejection.md
remedcu: Closing as the architecture was updated.

features/2026_05_19_explicit_transaction_rejection.md
```diff
@@ -0,0 +1,809 @@
...
+
+## Overview
+
+Currently, when a validator determines a transaction is invalid, it silently drops it. The FROST signing ceremony times out with no on-chain record. From the user's and operator's perspective, "rejected" and "error/timeout" are indistinguishable.
+
+This feature adds an explicit **decline** to the signing ceremony: a validator that determines a transaction is invalid calls `signDeclineWithCallback` on `FROSTCoordinator`. Once enough validators decline (threshold: `count - threshold + 1`), the ceremony is definitively marked rejected on-chain — further `signShare` calls revert, and a callback triggers `onSignRejected` on `Consensus`, which emits a `TransactionRejected` event. The explorer surfaces this as a distinct "Rejected" status.
+
+**Phases (separate PRs):**
+1. **FROSTCoordinator** — Add `signDecline`, threshold stopping logic, and `SignDeclined`/`SignRejected` events. No callback yet.
+2. **Consensus callback** — Add `signDeclineWithCallback` to `FROSTCoordinator`, add `onSignRejected` to `IFROSTCoordinatorCallback`, implement on `Consensus`, emit `TransactionRejected`/`OracleTransactionRejected`. Completes the on-chain rejection story.
```
rmeissner: nit: this can happen after the validator implementation.

rmeissner (approved):

features/2026_05_19_explicit_transaction_rejection.md
remedcu: Done 👍🏾

PR #389, opened 2026-05-20
feat: FROSTCoordinator — decline tracking and threshold stopping

remedcu:
Adds `signDecline` to `FROSTCoordinator`, allowing validators to explicitly signal they will not participate in a signing ceremony. The function emits `SignDeclined` and has no other on-chain effect.

### Context and Motivation

Currently validators silently drop transactions they consider invalid, and the ceremony times out with no on-chain record — rejection is indistinguishable from a timeout or error. This PR adds `signDecline` so validators can explicitly signal they are opting out. The `SignDeclined` event allows the explorer to distinguish "declined" from "timeout" for observability.

### Decisions and Tradeoffs

- **`signDecline` is purely indicative — no on-chain state effect**: an earlier design tracked decline counts in the coordinator and blocked signing once enough validators declined. This was dropped because blockchain reorgs create an irreconcilable conflict between chain state and the cryptographic model: a `signRevealNonces` on one fork can be replaced by a `signDecline` on the reorged fork, leaving honest validators with stale selection roots and a potentially stuck ceremony. Trying to enforce participation-state invariants that cannot be guaranteed across reorgs risks locking the protocol in a state from which no recovery is possible. The coordinator acts as a communication channel; FROST cryptography is the sole source of truth — a valid group signature is an attestable fact on any chain regardless of on-chain decline state.

- **`SignDeclined` is for observability only**: the event lets the explorer surface explicit declines separately from timeouts. From the protocol's perspective the two are equivalent — in both cases the validator does not contribute to the ceremony.

### Testing

Ships one test (`test_SignDecline_EmitsSignDeclined`): a group member calling `signDecline` emits `SignDeclined`. Error paths and further coverage follow in stacked PRs.


contracts/src/FROSTCoordinator.sol
```diff
@@ -111,11 +111,15 @@ contract FROSTCoordinator {
...
+     * @custom:param declineCount The number of participants that have declined this ceremony.
      * @custom:param shares The accumulated signature shares.
      */
     struct Signature {
         bytes32 message;
         bytes32 signed;
+        bool rejected;
+        uint16 declineCount;
         FROSTSignatureShares.T shares;
     }
```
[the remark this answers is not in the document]
remedcu: Thanks for flagging this. The storage layout shift is intentional and safe here — Safenet is a pre-production contract with no live deployments and no persistent state to migrate. New fields were placed before `shares` deliberately to keep the `Signature` struct logically grouped (flags and counts together, then the bulk data). There is no backward-compatibility constraint to work around at this stage.

---

contracts/src/FROSTCoordinator.sol
```diff
@@ -590,6 +647,36 @@ contract FROSTCoordinator {
...
+     *      rejected: `signShare` will revert with `CeremonyRejected` and signature queries will
+     *      revert with `SignatureRejected`. Additional declines past the threshold are still
+     *      recorded (for observability) but `SignRejected` is not re-emitted.
+     */
+    function signDecline(FROSTSignatureId.T sid) public returns (bool rejected) {
+        FROSTGroupId.T gid = sid.group();
+        Group storage group = $groups[gid];
+        group.participants.getKey(msg.sender); // reverts InvalidParticipant for non-members
+        Signature storage signature = $signatures[sid];
+        require(signature.message != bytes32(0), NotSigning());
```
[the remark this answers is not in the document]
remedcu: The current ordering is deliberate: `signDecline` checks participant membership (`group.participants.getKey`) **before** checking ceremony existence (`signature.message != 0`). This means a non-participant probing an arbitrary `sid` gets `InvalidParticipant` rather than a ceremony-existence signal, which avoids leaking information about which ceremonies are active to addresses outside the group. Using `_signatureGroupAndMessage` would invert this — ceremony existence would be checked first — which is a subtle privacy regression. Keeping the explicit ordering.

---

contracts/src/FROSTCoordinator.sol
```diff
@@ -308,6 +325,31 @@ contract FROSTCoordinator {
      */
     error WrongSignature();
 
+    /**
+     * @notice Thrown when a participant attempts to decline a ceremony they already declined.
```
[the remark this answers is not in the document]
remedcu: Good catch — the NatSpec is too narrow. `AlreadyDeclined` is also the revert reason in `signShare` when a previously-declined participant attempts to submit a share. Updating the doc to cover both call sites.

---

contracts/test/FROSTCoordinatorDecline.t.sol
```diff
@@ -0,0 +1,915 @@
...
+    // HELPERS
+    // ============================================================
+
+    function _randomSortedAddresses(uint16 count) private view returns (address[] memory result) {
+        result = new address[](count);
+        for (uint256 i = 0; i < result.length; i++) {
+            result[i] = vm.randomAddress();
+        }
+        result.sort();
+    }
```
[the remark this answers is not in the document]
remedcu: Valid observation — the helpers (`_trustedKeyGen`, `_trustedSign`, `_lagrangeCoefficient`, etc.) are duplicated from `FROSTCoordinator.t.sol`. This is a pre-existing pattern in the test suite. Extracting a shared base contract would be a worthwhile refactoring but is out of scope for this Phase 1 PR; it's better handled as a dedicated follow-up once the stacked phases are merged and the full test surface is known.

---

contracts/src/FROSTCoordinator.sol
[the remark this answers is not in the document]
remedcu: Safenet is on beta, but these changes are expected to be breaking changes for the next version, so it is fine for now to not be backward compatible.

contracts/test/FROSTCoordinatorDecline.t.sol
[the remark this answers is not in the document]
remedcu: Created an issue to note this: https://github.com/safe-research/safenet/issues/398

contracts/src/FROSTCoordinator.sol
```diff
@@ -573,9 +628,12 @@ contract FROSTCoordinator {
         bytes32[] calldata proof
     ) public returns (bool signed) {
         (Group storage group, bytes32 message) = _signatureGroupAndMessage(sid);
+        Signature storage signature = $signatures[sid];
+        require(!signature.rejected, CeremonyRejected());
+        require(!$declined[sid][msg.sender], AlreadyDeclined());
         Secp256k1.Point memory key = group.key;
         FROST.verifyShare(key, selection.r, group.participants.getKey(msg.sender), share, message);
-        Signature storage signature = $signatures[sid];
+        $shared[sid][msg.sender] = true; // guards signDecline: prevents share-then-decline
```
rmeissner: question: why not use the signature.shares for this? this way we could avoid introducing another field.

contracts/src/FROSTCoordinator.sol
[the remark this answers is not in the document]
rmeissner: I don't understand this privacy concern? All information about all ceremonies are public. Could you elaborate the risks here and the exact regression?

contracts/src/FROSTCoordinator.sol
```diff
@@ -324,6 +367,18 @@ contract FROSTCoordinator {
     // forge-lint: disable-next-line(mixed-case-variable)
     mapping(FROSTSignatureId.T => Signature) private $signatures;
 
+    /**
+     * @notice Mapping from signature ID and participant to whether that participant has declined.
+     */
+    // forge-lint: disable-next-line(mixed-case-variable)
+    mapping(FROSTSignatureId.T sid => mapping(address participant => bool)) private $declined;
```
rmeissner: request change: For me this breaks our existing pattern. I would expect all information for a signature flow to be kept in the `$signatures` mapping, potentially this information could even be part of the `SignaturesShares` library

contracts/src/FROSTCoordinator.sol
rmeissner: this would also allow to move all the logic related to the checks into the register method

contracts/src/FROSTCoordinator.sol
```diff
@@ -573,9 +628,12 @@ contract FROSTCoordinator {
         bytes32[] calldata proof
     ) public returns (bool signed) {
         (Group storage group, bytes32 message) = _signatureGroupAndMessage(sid);
+        Signature storage signature = $signatures[sid];
+        require(!signature.rejected, CeremonyRejected());
```
rmeissner: nit: theoretically we do not need to stop submission of shares when it was rejected (and it is also not aligned with the `signDecline` logic, which allows to decline even after completion.

contracts/src/FROSTCoordinator.sol
```diff
@@ -694,11 +783,52 @@ contract FROSTCoordinator {
...
+    }
+
+    /**
+     * @notice Returns the message associated with a signing ceremony.
+     * @dev Returns `bytes32(0)` if no ceremony has been started for this `sid`. Unlike most other query functions,
+     *      this does NOT revert for an unknown `sid`. Callers must treat a zero return as "not found."
+     * @param sid The signature ID.
+     * @return The message being signed, or `bytes32(0)` if the ceremony does not exist.
+     */
+    function signatureMessage(FROSTSignatureId.T sid) external view returns (bytes32) {
```
rmeissner: request change: This seems unrelated to the PR, why was this added?

contracts/src/FROSTCoordinator.sol
```diff
@@ -694,11 +783,52 @@ contract FROSTCoordinator {
...
         return $signatures[sid].shares.groupSignature(signed);
     }
 
+    /**
+     * @notice Returns whether a participant has declined a specific signing ceremony.
+     * @param sid The signature ID.
+     * @param participant The participant address.
+     * @return True if the participant has declined.
+     */
+    function isSignDeclined(FROSTSignatureId.T sid, address participant) external view returns (bool) {
```
rmeissner: nit: Imo all the read methods here are not relevant to the PR and I would remove them to keep the PR small

contracts/test/FROSTCoordinatorDecline.t.sol
```diff
@@ -0,0 +1,915 @@
...
+        vm.prank(participants.addr(0));
+        bool rejected = coordinator.signDecline(sid);
+
+        assertTrue(rejected);
+        assertTrue(coordinator.isSignRejected(sid));
+    }
+}
+
+/// @dev threshold == 2 (4-of-2): DECLINE_THRESHOLD = 3 — need count-threshold+1 = 3 declines to reject.
+contract FROSTCoordinatorDeclineMinThresholdTest is Test {
```
rmeissner: request change: Why is this a separate test contract, the whole setup is duplicated because of this

contracts/test/FROSTCoordinatorDecline.t.sol
```diff
@@ -0,0 +1,915 @@
...
+import {ForgeSecp256k1} from "@test/util/ForgeSecp256k1.sol";
+import {ParticipantMerkleTree} from "@test/util/ParticipantMerkleTree.sol";
+import {FROSTCoordinator} from "@/FROSTCoordinator.sol";
+import {FROST} from "@/libraries/FROST.sol";
+import {FROSTGroupId} from "@/libraries/FROSTGroupId.sol";
+import {FROSTParticipantMap} from "@/libraries/FROSTParticipantMap.sol";
+import {FROSTSignatureId} from "@/libraries/FROSTSignatureId.sol";
+import {Secp256k1} from "@/libraries/Secp256k1.sol";
+
+contract FROSTCoordinatorDeclineTest is Test {
```
rmeissner: nit: to be honest reviewing 900 lines of test code is really hard. I would recommend to cover the happy path first and then add the tests in smaller sections.

rmeissner (changes requested):
For me the tests are hard to review (so far I have not reviewed them in detail). I would really like if we chunk the tests into smaller parts (note this is not really a blocker, but would make the review easier).

Change requests have been added in the comments.

contracts/src/FROSTCoordinator.sol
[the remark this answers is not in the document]
remedcu: You are right — the privacy argument was wrong. All on-chain state is public; the check ordering provides no real privacy benefit. `signDecline` has been refactored to use `_signatureGroupAndMessage` for consistency with `signRevealNonces` and `signShare`.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — `anyShared` has been added to `FROSTSignatureShares.T` and set inside `register` after a successful share. `signDecline` now checks `signature.shares.isShared(msg.sender)` to enforce mutual exclusion. The top-level `$shared` mapping has been removed entirely.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — `declined` is now a `mapping(address participant => bool)` field directly inside the `Signature` struct, keeping all per-ceremony state in `$signatures[sid]`. The top-level `$declined` mapping has been removed.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — the `CeremonyRejected` guard has been removed from `signShare`. This is symmetric with `signDecline` being callable past the threshold: both are still accepted after the ceremony is definitively rejected (for observability), while `signatureVerify`/`signatureValue` gate on `SignatureRejected` for callers that need a definitive answer.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — `isSignDeclined`, `isSignShared`, `isSignRejected`, and `signatureMessage` have all been removed. They are not used by any external contract or off-chain consumer in Phase 1 and can be reintroduced in Phase 2 when `Consensus.rejectTransaction` needs them.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — all four read methods (`isSignDeclined`, `isSignShared`, `isSignRejected`, `signatureMessage`) have been removed from this PR.

---

contracts/test/FROSTCoordinatorDecline.t.sol
remedcu: The test file has been stripped to a single happy-path test (`test_SignDecline_ThresholdReached_EmitsSignRejected`) with a minimal `_keyGen` helper, so the setup duplication is now minimal. The rest of the test coverage will be added via four stacked PRs on top of this branch (threshold boundaries, mutual exclusion, edge cases, post-rejection behaviour), each reviewable independently.

---

contracts/test/FROSTCoordinatorDecline.t.sol
remedcu: Done — the test file is now ~80 lines with a single test covering the happy path (threshold reached → `SignRejected` emitted, return value correct). The remaining coverage is split across four stacked PRs: threshold boundaries, mutual exclusion (share↔decline), edge case configurations, and post-rejection/post-completion behaviour. Each stacked PR will be small and independently reviewable.

---

contracts/src/FROSTCoordinator.sol
```diff
@@ -308,6 +327,27 @@ contract FROSTCoordinator {
...
+    /**
+     * @notice Thrown when a participant attempts to decline a ceremony they already shared a signature for.
+     */
+    error AlreadyShared();
+
+    /**
+     * @notice Thrown when a participant attempts to decline a ceremony that has already been signed.
+     */
+    error SigningComplete();
+
```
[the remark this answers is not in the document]
remedcu: `CeremonyRejected` is intentionally not added. Once the rejection threshold is crossed, the remaining non-declined participants are fewer than `threshold`, so the ceremony can never accumulate enough shares to complete regardless — the guard would be mathematically redundant. Keeping `signShare` open after rejection is also symmetric with `signDecline` being callable past threshold for observability reasons. The reviewer who flagged the original `CeremonyRejected` guard in `signShare` explicitly requested its removal. The updated spec and PR description both explain this decision in the Architecture Decision section.

---

contracts/src/FROSTCoordinator.sol
```diff
@@ -573,9 +613,10 @@ contract FROSTCoordinator {
         bytes32[] calldata proof
     ) public returns (bool signed) {
         (Group storage group, bytes32 message) = _signatureGroupAndMessage(sid);
+        Signature storage signature = $signatures[sid];
+        require(!signature.declined[msg.sender], AlreadyDeclined());
```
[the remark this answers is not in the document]
remedcu: The rejection guard is intentionally absent from `signShare`. After the rejection threshold is crossed, the remaining non-declined participants are fewer than `threshold`, so a complete signing set can never be assembled — no guard is needed to prevent ceremony completion. Omitting it is also symmetric with `signDecline` being callable past the threshold (both remain open for observability). The `CeremonyRejected` guard was explicitly removed per reviewer request. See the Architecture Decision section in the updated spec and PR description for the full rationale.

---

contracts/test/FROSTCoordinatorDecline.t.sol
```diff
@@ -0,0 +1,80 @@
...
+            vm.prank(participants.addr(i));
+            assertFalse(coordinator.signDecline(sid));
+        }
+
+        vm.expectEmit();
+        emit FROSTCoordinator.SignRejected(sid);
+        vm.prank(participants.addr(DECLINE_THRESHOLD - 1));
+        bool rejected = coordinator.signDecline(sid);
+        assertTrue(rejected);
+    }
```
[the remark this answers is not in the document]
remedcu: Intentional. This PR ships one focused happy-path test per reviewer feedback to keep the PR small and reviewable. The remaining cases (threshold boundary, mutual exclusion, error paths, post-rejection/post-completion) are split across four stacked PRs that target this branch, each independently reviewable. The stacked PR plan is documented in the updated spec and PR description under "Test cases (Phase 1)".

---

contracts/src/FROSTCoordinator.sol
```diff
@@ -573,9 +613,10 @@ contract FROSTCoordinator {
...
         Secp256k1.Point memory key = group.key;
         FROST.verifyShare(key, selection.r, group.participants.getKey(msg.sender), share, message);
```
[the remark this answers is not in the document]
remedcu: The scenario described is mathematically impossible. The rejection threshold is `count - threshold + 1`. Once that many participants have declined, the number of remaining non-declined participants is `count - (count - threshold + 1) = threshold - 1`, which is strictly fewer than `threshold`. A signing ceremony requires `threshold` shares; it can never complete. There is no guard needed in `signShare` because the precondition for `SignCompleted` (enough shares from non-declined participants) can never be satisfied. `signatureVerify` and `signatureValue` additionally gate on `signature.rejected` with `SignatureRejected`, giving callers a clear error if they query a rejected SID. The `CeremonyRejected` guard on `signShare` was deliberately removed by request.

---

contracts/src/FROSTCoordinator.sol
```diff
@@ -590,6 +631,33 @@ contract FROSTCoordinator {
...
+     * @param sid The signature ID.
+     * @return rejected True if the decline threshold has been crossed for the first time.
+     * @dev Once `count - threshold + 1` participants have declined, the ceremony is definitively
+     *      rejected: signature queries revert with `SignatureRejected`. Additional declines past
+     *      the threshold are still recorded (for observability) but `SignRejected` is not re-emitted.
+     */
+    function signDecline(FROSTSignatureId.T sid) public returns (bool rejected) {
+        (Group storage group,) = _signatureGroupAndMessage(sid);
+        group.participants.getKey(msg.sender); // reverts InvalidParticipant for non-members
+        Signature storage signature = $signatures[sid];
```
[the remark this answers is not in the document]
remedcu: Intentionally deferred to Phase 2. `isSignRejected` and `signatureMessage` are only needed by `Consensus.rejectTransaction`, which doesn't exist until Phase 2. Adding them in Phase 1 would introduce dead external API surface with no caller. They will be added in the same commit as `signDeclineWithCallback` and `IFROSTCoordinatorCallback.onSignRejected`, which is the natural point where the interface becomes useful. The updated spec and PR description document this decision explicitly.

---

features/2026_05_19_explicit_transaction_rejection.md
```diff
@@ -111,8 +111,11 @@ struct Signature {
 
 ```solidity
 mapping(FROSTSignatureId.T sid => mapping(address participant => bool)) private $declined;
+mapping(FROSTSignatureId.T sid => mapping(address participant => bool)) private $shared;
 ```
 
+`$declined` records which participants have explicitly declined a ceremony. `$shared` records which participants have successfully submitted a signature share. Both are needed to enforce mutual exclusion (see Updated `signShare` and `signDecline` below).
+
```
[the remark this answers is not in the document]
remedcu: The spec has been updated in the latest commit to this branch. It now correctly reflects the actual storage layout: `declined` as a `mapping(address => bool)` inside the `Signature` struct, `anyShared` inside `FROSTSignatureShares.T`, and no top-level `$declined`/`$shared` mappings.

---

contracts/test/FROSTCoordinatorDecline.t.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    // ============================================================
+    // HELPERS
+    // ============================================================
+
```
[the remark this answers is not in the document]
remedcu: Intentional. This PR ships one focused happy-path test per reviewer feedback to keep the diff small and reviewable. The cases you list (non-participant revert, double-decline, `AlreadyShared`, `AlreadyDeclined`, `SignatureRejected` on `signatureVerify`/`signatureValue`) are all covered in four stacked PRs targeting this branch: threshold boundary, mutual exclusion, error paths, and post-rejection/post-completion behaviour. The plan is documented in the updated spec and PR description.

---

nlordell:
First of all:

> Once enough validators decline (`count - threshold + 1`), the ceremony is definitively marked rejected

This is quite cleaver, and avoids us fixing a `treshold > count / 2`.

nlordell:
question: Why is the declining process happening after revealing nonces? I would expect a validator to just say "I don't want to sign" and not reveal its nonce (incrementing the sequence in the nonces tree, thus preventing the validator from participating in signature sharing if it were to decline).

**Edit**: To clarify a bit, right now, it is possible for a validator to reveal their nonces without an intent to participate in signing (i.e. they plan on declining). This causes issues as a validator may signature share to the incorrect selection root (one that includes the validator that wants to decline signing). Additionally, keeping track of declines becomes very awkward because it needs to be done across selection roots. The more natural place for me is with `NonceCommitmentSet`, where you can either `verify` a nonce for revealing, or `burn` the nonce by setting some mapping on the sequence ID to say "I am not allowed to reveal a nonce with this sequence anymore. Nonces are already per sequence (i.e. signature ID), which is the same "keys", unlike signing shares (which are per signature ID and selection root tuples), making it awkward.

request change: move the tracking of who did and did not decline to `NOncesCommitmentsSet` and have the invariant either: timeout, nonces are revealed exactly once, signing is declined exactly once.

contracts/src/FROSTCoordinator.sol
```diff
@@ -111,11 +111,17 @@ contract FROSTCoordinator {
...
+     * @custom:param declineCount The number of participants that have declined this ceremony.
+     * @custom:param declined Per-participant decline flags.
      * @custom:param shares The accumulated signature shares.
      */
     struct Signature {
         bytes32 message;
         bytes32 signed;
+        bool rejected;
+        uint16 declineCount;
+        mapping(address participant => bool) declined;
```
nlordell: request change: We have duplicated mappings: `anyShared` and `declined` here can carry enough information without storing additional state to the contract. It also is creates assumptions/invariants on how these two fields relate (they are exclusive) in a confusing way. We should use a single mapping - it can store 256 boolean flags if needed (although I think just one flag is generally sufficient, if this is part of the nonces commitments).

nlordell (changes requested):

contracts/src/FROSTCoordinator.sol
remedcu: Updated 👍🏾

contracts/src/libraries/FROSTNonceCommitmentSet.sol
```diff
@@ -120,7 +148,9 @@ library FROSTNonceCommitmentSet {
         Secp256k1.Point memory e,
         uint64 sequence,
         bytes32[] calldata proof
-    ) internal view {
+    ) internal {
+        require(self.commitments[participant].nonces[sequence] != SequenceStatus.Burned, NoncesAlreadyBurned());
```
rmeissner: Should we also disallow that it is revealed multiple times?

remedcu: Updated 👍🏾

contracts/src/libraries/FROSTNonceCommitmentSet.sol
```diff
@@ -39,6 +41,19 @@ library FROSTNonceCommitmentSet {
...
+     * @dev A nonce slot starts as `None`, transitions to `Revealed` when the participant calls
+     *      `signRevealNonces`, or to `Burned` when the participant calls `signDecline`. These
+     *      transitions are mutually exclusive: a revealed nonce cannot be burned and a burned slot
+     *      cannot be revealed.
+     */
+    enum SequenceStatus {
+        None,
+        Revealed,
+        Burned
+    }
```
nlordell: request change: this differentiation is not used at all by our code (except for error differentiation, which has very limited value: _these are not user-facing functions_), it is a premature abstraction IMO and would remove it. Enums in SOlidity generate really bad code.

contracts/src/libraries/FROSTNonceCommitmentSet.sol
```diff
@@ -48,6 +63,16 @@ library FROSTNonceCommitmentSet {
...
+    /**
+     * @notice Thrown when attempting to burn a nonce slot that has already been revealed.
+     */
+    error NoncesAlreadyRevealed();
+
+    /**
+     * @notice Thrown when attempting to reveal or burn a nonce slot that has already been burned.
+     */
+    error NoncesAlreadyBurned();
+
```
nlordell: comment: Given the requested change above, this can be collapsed to:


```suggestion
    error NoncesAlreadyUsed();
```

contracts/src/FROSTCoordinator.sol
```diff
@@ -259,6 +267,19 @@ contract FROSTCoordinator {
      */
     event SignCompleted(FROSTSignatureId.T indexed sid, bytes32 indexed selectionRoot, FROST.Signature signature);
 
+    /**
+     * @notice Emitted when a participant declines a signing ceremony.
+     * @param sid The signature ID.
+     * @param participant The participant address that declined.
+     */
+    event SignDeclined(FROSTSignatureId.T indexed sid, address indexed participant);
```
nlordell: nit: to keep the events ordered, I would put this under `SignNoncesRevelaed` event: they happen at the "same time" in the signing ceremony.

contracts/src/FROSTCoordinator.sol
```diff
@@ -308,6 +329,27 @@ contract FROSTCoordinator {
...
+    /**
+     * @notice Thrown when a participant attempts to submit a signature share without first revealing
+     *         their nonces via `signRevealNonces`.
+     */
+    error NoncesNotRevealed();
+
+    /**
+     * @notice Thrown when a participant attempts to decline a ceremony that has already been signed.
+     */
+    error SigningComplete();
```
nlordell: question: I'm not sure why we maintain this invariant. We can reveal nonces for signatures that are complete, why not decline? I would personally relax this (there is no harm in declining late).

contracts/src/FROSTCoordinator.sol
```diff
@@ -308,6 +329,27 @@ contract FROSTCoordinator {
...
+    /**
+     * @notice Thrown when a previously-declined participant attempts to submit a signature share.
+     */
+    error AlreadyDeclined();
+
+    /**
+     * @notice Thrown when a participant attempts to submit a signature share without first revealing
+     *         their nonces via `signRevealNonces`.
+     */
+    error NoncesNotRevealed();
```
nlordell: request change: this error was intentionally omitted in the original version. We already have a `selectionRoot` that would prevent malicious validators for trying to sign to a selection of honest validators.

contracts/src/FROSTCoordinator.sol
```diff
@@ -308,6 +329,27 @@ contract FROSTCoordinator {
      */
     error WrongSignature();
 
+    /**
+     * @notice Thrown when a previously-declined participant attempts to submit a signature share.
+     */
+    error AlreadyDeclined();
```
nlordell: request change: idem, I don't think this is needed.

contracts/src/FROSTCoordinator.sol
```diff
@@ -308,6 +329,27 @@ contract FROSTCoordinator {
...
+
+    /**
+     * @notice Thrown when a participant attempts to decline a ceremony that has already been signed.
+     */
+    error SigningComplete();
+
+    /**
+     * @notice Thrown when querying a signature for a ceremony that was rejected.
+     */
+    error SignatureRejected();
```
nlordell: request change: error differentiation is not useful IMO. _these aren't user facing functions_. `NotSigned` is enough.

contracts/src/FROSTCoordinator.sol
```diff
@@ -111,11 +111,19 @@ contract FROSTCoordinator {
...
+     * @dev Per-participant decline and reveal state is tracked in `FROSTNonceCommitmentSet` via the
+     *      `SequenceStatus` enum, which enforces that each participant either reveals nonces (at most
+     *      once) or declines (at most once) for a given sequence — the two transitions are mutually
+     *      exclusive.
      */
     struct Signature {
         bytes32 message;
         bytes32 signed;
+        bool rejected;
+        uint16 declineCount;
```
nlordell: request change: `rejected` is duplicate storage information: Given a signature ID, you can always compute:

```solidity
bool rejected = signature.declineCount > (group.count - group.threshold)
```

It doesn't need to be stored in state (especially not for error differentiation on a view function).

contracts/src/FROSTCoordinator.sol
```diff
@@ -590,6 +640,31 @@ contract FROSTCoordinator {
...
+    /**
+     * @notice Records an explicit decline by the sender for a signing ceremony.
+     * @param sid The signature ID.
+     * @return rejected True if the decline threshold has been crossed for the first time.
+     * @dev Once `count - threshold + 1` participants have declined, the ceremony is definitively
+     *      rejected: signature queries revert with `SignatureRejected`. Additional declines past
+     *      the threshold are still recorded (for observability) but `SignRejected` is not re-emitted.
+     */
+    function signDecline(FROSTSignatureId.T sid) public returns (bool rejected) {
+        (Group storage group,) = _signatureGroupAndMessage(sid);
```
nlordell: request change: Don't read `message` from storage if you don't plan on using it.

contracts/src/FROSTCoordinator.sol
```diff
@@ -590,6 +640,31 @@ contract FROSTCoordinator {
...
+        group.participants.getKey(msg.sender); // reverts InvalidParticipant for non-members
+        Signature storage signature = $signatures[sid];
+        require(signature.signed == bytes32(0), SigningComplete());
+        group.nonces.burn(msg.sender, sid.sequence());
+        signature.declineCount++;
+        emit SignDeclined(sid, msg.sender);
+        GroupState memory state = group.state;
+        if (signature.declineCount >= state.count - state.threshold + 1 && !signature.rejected) {
+            signature.rejected = true;
```
nlordell: comment: AFAICT, this is just used here and for emitting "nicer errors" on some view methods. I don't think this is a good use of gas.

contracts/test/FROSTCoordinatorDecline.t.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: GPL-3.0-only
```
nlordell: request change: please put the small test function in `FROSTCoordinator.t.sol` and do a refactor in a follow-up (if it makes sense). This just adds a lot of noise to the PR.

nlordell (changes requested):
This is much closer to how I think it should work.

There are a few issues: my biggest issue is the additional use of gas for the sole purpose of reverting with better errors. I don't think we should do this.

Additionally, I don't think the check that a validator is not allowed to decline if the signature is already complete is needed. It also is inconsistent with what we do with nonce reveals for completed signatures, so I would suggest removing it. Protection against submitting shares without revealing nonces is implicit with the `selectionRoot` and enforced by honest validators.

contracts/src/libraries/FROSTNonceCommitmentSet.sol
remedcu: Done — the `SequenceStatus` enum has been removed entirely. Both `verify()` and `burn()` now operate on a single `used` boolean per `(participant, sequence)`, with a single `NoncesAlreadyUsed` error covering both paths. No enum code is generated.

---

contracts/src/libraries/FROSTNonceCommitmentSet.sol
remedcu: Done — collapsed to a single `NoncesAlreadyUsed` error as suggested.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — `SignDeclined` is now positioned immediately after `SignRevealedNonces` in the events section, reflecting that both occur at the same stage of the signing ceremony.

---

contracts/src/FROSTCoordinator.sol
remedcu: The `used` boolean only prevents a participant from **both** revealing nonces **and** declining for the same sequence — there is no check on whether the ceremony has completed. A participant who has not yet interacted can decline at any time, including after the ceremony completes or is rejected. The restriction is purely: once `signRevealNonces` has been called (`verify()` sets `used = true`), that participant cannot also call `signDecline` for that sequence (and vice versa). Late declines from participants who never revealed are fully accepted.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — the `AlreadyDeclined` guard has been removed from `signShare`. The `selectionRoot` Merkle commitment is the authoritative gate: a participant who has declined (or never revealed nonces) cannot appear in a valid signing selection.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — removed.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — `SignatureRejected` has been removed. `NotSigned` covers both the "still in progress" and "rejected" cases for `signatureVerify` and `signatureValue`.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — `bool rejected` has been removed from `Signature`. Rejection is computable as `signature.declineCount > group.state.count - group.state.threshold` and does not need to be stored.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — `signDecline` no longer calls `_signatureGroupAndMessage`. Ceremony existence is checked with `$signatures[sid].message != bytes32(0)` (a single minimal read for the check itself), and `message` is not loaded into a local variable.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — the redundant rejection check in view functions has been removed. `signatureVerify` and `signatureValue` rely solely on `NotSigned` (checking `signed == bytes32(0)`), which covers both in-progress and rejected ceremonies without additional storage reads or error variants.

---

contracts/test/FROSTCoordinatorDecline.t.sol
remedcu: Done — `test_SignDecline_ThresholdReached_EmitsSignRejected` has been moved into `FROSTCoordinator.t.sol` and the separate `FROSTCoordinatorDecline.t.sol` file has been deleted.

---

nlordell:
request change: Can you please update the PR description, there is a lot of outdated information (for example the `rejected` flag). I would also ask that it be trimmed down, there is a lot of noise in the description (changes summary is just listing what can be seen in the code: for example "new errors" are obvious from the code, and unless there is something noteworthy about a non-obvious choice, please do not include it in the description). **Note that you are not expected to fill out all the sections all the time**, only what is relevant to the reviewer. Again, a rule of thumb is: read through the description and see if is succinct and informative, and includes **rationale** for non-obvious decisions (and not just mention that they were made).

contracts/src/FROSTCoordinator.sol
```diff
@@ -111,11 +111,18 @@ contract FROSTCoordinator {
      * @notice Tracks a signing ceremony state.
      * @custom:param message The message being signed.
      * @custom:param signed The Merkle root of the signature shares.
+     * @custom:param declineCount The number of participants that have declined this ceremony.
      * @custom:param shares The accumulated signature shares.
+     * @dev Per-participant decline and reveal state is tracked in `FROSTNonceCommitmentSet` via a single
+     *      `used` boolean per sequence. The first call to either `signRevealNonces` or `signDecline`
+     *      consumes the slot; any subsequent call reverts with `NoncesAlreadyUsed`.
```
nlordell: nit: I find this comment a bit confusing, I wouldn't add it here.

contracts/src/FROSTCoordinator.sol
```diff
@@ -573,9 +598,10 @@ contract FROSTCoordinator {
         bytes32[] calldata proof
     ) public returns (bool signed) {
         (Group storage group, bytes32 message) = _signatureGroupAndMessage(sid);
-        Secp256k1.Point memory key = group.key;
-        FROST.verifyShare(key, selection.r, group.participants.getKey(msg.sender), share, message);
         Signature storage signature = $signatures[sid];
+        Secp256k1.Point memory senderKey = group.participants.getKey(msg.sender);
```
nlordell: nit:

```suggestion
        Secp256k1.Point memory participantKey = group.participants.getKey(msg.sender);
```

remedcu: There is a function named `participantKey(...)` already. How about `_participantKey`?

contracts/src/FROSTCoordinator.sol
```diff
@@ -573,9 +598,10 @@ contract FROSTCoordinator {
...
+        Secp256k1.Point memory key = group.key;
+        FROST.verifyShare(key, selection.r, senderKey, share, message);
```
nlordell: request change: I just noticed, this function does not need to be changed (including the added `@dev` NatSpec section). Can we revert these changes to this function as they are additional noise in the PR.

contracts/src/FROSTCoordinator.sol
```diff
@@ -590,6 +616,31 @@ contract FROSTCoordinator {
...
+    function signDecline(FROSTSignatureId.T sid) public returns (bool rejected) {
+        require($signatures[sid].message != bytes32(0), NotSigning());
+        Group storage group = $groups[sid.group()];
+        group.participants.getKey(msg.sender); // reverts InvalidParticipant for non-members
+        Signature storage signature = $signatures[sid];
+        group.nonces.burn(msg.sender, sid.sequence());
+        signature.declineCount++;
+        emit SignDeclined(sid, msg.sender);
+        GroupState memory state = group.state;
+        if (signature.declineCount == state.count - state.threshold + 1) {
```
nlordell: request change: This is double-reading the storage slot. 
```suggestion
        uint16 declineCount = ++signature.declineCount
        emit SignDeclined(sid, msg.sender);
        GroupState memory state = group.state;
        if (declineCount == state.count - state.threshold + 1) {
```

contracts/src/FROSTCoordinator.sol
```diff
@@ -693,8 +744,7 @@ contract FROSTCoordinator {
      * @return result The FROST signature.
      */
     function signatureValue(FROSTSignatureId.T sid) external view returns (FROST.Signature memory result) {
-        Signature storage signature = $signatures[sid];
-        bytes32 signed = signature.signed;
+        bytes32 signed = $signatures[sid].signed;
         require(signed != bytes32(0), NotSigned());
```
nlordell, line 717: nit: Another example of unnecessary noise that should be reverted in the PR.

nlordell (changes requested):
Some more minor changes requested.

remedcu:
Done — decline tracking has been moved to `FROSTNonceCommitmentSet`. Both `signRevealNonces` (via `verify()`) and `signDecline` (via `burn()`) set a single `used` boolean per `(participant, sequence)`, giving the invariant: a participant either times out, reveals nonces exactly once, or declines exactly once.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — removed.

---

contracts/src/FROSTCoordinator.sol
remedcu: Moot — the entire `signShare` function has been reverted to its original form (no variable rename), per the request below.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — reverted. The `@dev` NatSpec block and the variable reorder are both removed; `signShare` is back to its original form.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — applied the suggestion: `uint16 declineCount = ++signature.declineCount` caches the incremented value so the storage slot is only read once.

---

contracts/src/FROSTCoordinator.sol
remedcu: Done — reverted. `signatureValue` is back to using `Signature storage signature = $signatures[sid]`.

---

contracts/src/FROSTCoordinator.sol
```diff
@@ -590,6 +605,31 @@ contract FROSTCoordinator {
...
+     *      but `SignRejected` is not re-emitted. Late declines after signing completes are also
+     *      accepted — they cannot affect the outcome.
+     */
+    function signDecline(FROSTSignatureId.T sid) public returns (bool rejected) {
+        require($signatures[sid].message != bytes32(0), NotSigning());
+        Group storage group = $groups[sid.group()];
+        group.participants.getKey(msg.sender); // reverts InvalidParticipant for non-members
+        Signature storage signature = $signatures[sid];
+        group.nonces.burn(msg.sender, sid.sequence());
+        uint16 declineCount = ++signature.declineCount;
```
rmeissner: ubernit: I was expecting this variable to be used for something else then the check since it is assigned before the `emit` instead before the `if`

rmeissner (approved):
Code changes look good.

Note: I did not review the feature spec update

contracts/src/FROSTCoordinator.sol
remedcu: Updated 👍🏾

nlordell:
From the description:

> A single used boolean per slot replaces the former dual-mapping approach (declined in Signature, anyShared in FROSTSignatureShares.T), giving one source of truth with no cross-struct invariant to maintain.

This does not belong in the description that will get merged: we never had an `anyShared` to replace.

> The selectionRoot Merkle commitment additionally prevents participants who have not revealed their nonces from appearing in a valid signing selection.

This is not correct: the selection root prevents participants which influencing the share collection of other participant selections where they are excluded. Assuming honest participants require onchain nonce reveals, then they cannot contribute to the signing selection for honest participants.

---

One thing I noticed is that there are edge cases WRT reorgs:
- Malicious validator A includes a `signNonceReveal` in block X at height n
- Honest validator B includes a `signShare` in block Y at height n+1
- Malicious validator A includes a `signDecline` in reorged block X' at height n
- Honest validator B's `signShare` gets included at reorged block Y' at n+1 including A in the selection (despite them having declined that selection)
- A signature can simultaneously be signed and declined.

IMO: we need to think of a way to handle this. It may be that we do need to introduce additional accounting to prevent this :thinking:

rmeissner:
> IMO: we need to think of a way to handle this. It may be that we do need to introduce additional accounting to prevent this 🤔

Is this in scope for this PR tho? Also imo this is not a contract issue but if at all a validator/ general protocol design issue, therefore I would handle this separately.

contracts/src/FROSTCoordinator.sol
```diff
@@ -111,6 +111,7 @@ contract FROSTCoordinator {
      * @notice Tracks a signing ceremony state.
      * @custom:param message The message being signed.
      * @custom:param signed The Merkle root of the signature shares.
+     * @custom:param declineCount The number of participants that have declined this ceremony.
```
nlordell:
```suggestion
```

remedcu: Updated 👍🏾

nlordell:
request change: The PR description is outdated,it would be good to capture some of our discussion in the rationale as to why this was chosen to be perfectly indicative. We all know why, but our future selves would really appreciate if we wrote it down

rmeissner (approved):

nlordell (approved):
> `signDeclineWithCallback` (the Consensus callback) is deferred to Phase 2.

This is no longer true right? Please remove (since we do not know when to call the callback).

Optimistically approving.

PR #394, opened 2026-05-21
Explorer: default to Gnosis Chain RPC

tschubotz:
Switch the explorer's default RPC to the official Gnosis Chain public endpoint.

Context and Motivation

- The previous default (`1rpc.io`) is a third-party aggregator and became unreliable. `rpc.gnosischain.com` is the official Gnosis Chain public RPC and a more appropriate default for a Gnosis Chain explorer.


tschubotz:
@rmeissner that build on cloudflare pages fails due to node version mismatch - before I change settings there: Was the version recently upped from 22 to 24 for Safenet? Do I see correctly that this needs a config fix on cloudflare rather than a code fix on my end?

features/2026_05_21_rpc_url_param_and_default.md
```diff
@@ -0,0 +1,103 @@
+# Feature Proposal: DAppSpec RPC URL Param + Default RPC Update
```
rmeissner: Feature spec should be removed (or not merged) if the feature is completed.

features/2026_05_21_rpc_url_param_and_default.md
```diff
@@ -0,0 +1,103 @@
...
+Component: `explorer`
+
+---
+
+## Overview
+
+Two small related changes to the explorer's RPC configuration:
```
rmeissner: nit: under our current PR policy (#smallerisbetter) I would have defaulted to have it as 2 PRs

features/2026_05_21_rpc_url_param_and_default.md
```diff
@@ -0,0 +1,103 @@
...
+
+1. Change the default RPC endpoint from `https://1rpc.io/gnosis` to `https://rpc.gnosischain.com/`.
+2. Support the [DAppSpec](https://ethereum-magicians.org/t/new-erc-best-practices-for-dapps-dappspec/24407) `ds-rpc-{chainId}` URL query parameter so users and integrators can override the RPC endpoint via URL. For this Gnosis Chain explorer the relevant parameter is `ds-rpc-100`.
```
rmeissner: Hmm I understand that the standard requires to specify the chainId, but for us this makes no sense, as we retrieve the chainid from the rpc in our code. 

How adopted is this standard (is there even a final ERC)? Otherwise it might make sense to consider using just a custom `rpc-override` query string, to avoid this weird limiting to a hardcoded chain id.

explorer/src/lib/settings.ts
```diff
@@ -3,6 +3,19 @@ import { z } from "zod";
 import { checkedAddressSchema } from "@/lib/schemas";
 
 const STORAGE_KEY_SETTINGS = "localStorage.settings.object.v1";
+
+// DAppSpec (ds-rpc-{chainId}) — Gnosis Chain only
+const GNOSIS_CHAIN_ID = 100;
+
+const _rpcUrlParam = (() => {
```
rmeissner: why the `_` prefix?

request change: Also imo this should not be a lib function but rather a hook OR even use the tanstack `useSearch` hook

rmeissner (changes requested):

features/2026_05_21_rpc_url_param_and_default.md
tschubotz: Removed

features/2026_05_21_rpc_url_param_and_default.md
tschubotz: query param feature removed from PR

explorer/src/lib/settings.ts
tschubotz: query param feature removed from PR

features/2026_05_21_rpc_url_param_and_default.md
tschubotz: query param feature removed from PR

tschubotz:
@rmeissner I've removed the query param feature from PR. That leaves only the rpc update

rmeissner (approved):

PR #395, opened 2026-05-21
fix: add og:image and twitter:image meta tags to explorer

### LINKED ISSUE #382: explorer: missing og:image / twitter:image breaks link previews on X and Slack
## Problem

Sharing https://explorer.safenet-beta.eth.limo on X (Twitter), Slack, Discord, LinkedIn, or iMessage produces a blank link preview card with no image, and on X the card fails to render entirely.

`explorer/index.html` declares `<meta property="twitter:card" content="summary_large_image" />` and the standard `og:title` / `og:description` tags, but does not declare `og:image`, `og:url`, or `twitter:image`. Most social crawlers require at least `og:image` or `twitter:image` to render a card; without them the card is skipped or rendered blank.

Verified against the live deploy:

```sh
curl -sL "https://explorer.safenet-beta.eth.limo" -H "User-Agent: facebookexternalhit/1.1" | grep -iE "og:image|twitter:image"
# (returns nothing)
```

## Suggested diff

In `explorer/index.html`, add the missing image and URL tags to the existing Open Graph and Twitter sections:

```diff
     <!-- Open Graph / Facebook -->
     <meta property="og:type" content="website" />
+    <meta property="og:url" content="https://explorer.safenet-beta.eth.limo" />
     <meta property="og:title" content="Safenet Explorer" />
     <meta property="og:description" content="..." />
+    <meta property="og:image" content="https://explorer.safenet-beta.eth.limo/logo512.png" />
+    <meta property="og:image:width" content="512" />
+    <meta property="og:image:height" content="512" />
+    <meta property="og:image:alt" content="Safenet Explorer" />
     <meta property="og:site_name" content="Safenet Explorer" />
     <meta property="og:locale" content="en_US" />

     <!-- Twitter -->
     <meta property="twitter:card" content="summary_large_image" />
     <meta property="twitter:title" content="Safenet Explorer" />
     <meta property="twitter:description" content="..." />
+    <meta property="twitter:image" content="https://explorer.safenet-beta.eth.limo/logo512.png" />
+    <meta property="twitter:image:alt" content="Safenet Explorer" />
```

Absolute URLs are intentional so the `.eth.limo` IPFS gateway and any alternate deploy host resolve the asset correctly.

## Note on the image asset

`/logo512.png` is currently the only image in `explorer/public/`. It works but is a 512×512 square logo, so `summary_large_image` cards will render as smaller square cards rather than full-width heros. For optimal rendering, a 1200×630 PNG/JPG added to `explorer/public/og-image.png` would replace `/logo512.png` in the meta tags above.

## Verification after deploy

- X Card Validator: https://cards-dev.twitter.com/validator
- Facebook Sharing Debugger: https://developers.facebook.com/tools/debug/
- Or:
  ```sh
  curl -sL "https://explorer.safenet-beta.eth.limo" -H "User-Agent: facebookexternalhit/1.1" | grep -i og:image
  ```

tschubotz:
Sharing the explorer URL on X, Slack, Discord, and iMessage produced a
blank preview card because og:image and twitter:image were missing.

### Context and Motivation

- Social crawlers require at least og:image to render a link preview
  card. Without it the card is skipped entirely (verified via curl with
  facebookexternalhit UA). Fixes #382.

### Decisions and Tradeoffs

- The image URL must be absolute — relative paths break on IPFS path
  gateways where the document origin differs from the asset root, and
  the OG spec requires absolute URLs regardless.
- The deploy URL is provided via `VITE_APP_URL`. Vite's built-in
  `%VITE_*%` HTML substitution only fires when the var is present in
  the env, so a `transformIndexHtml` plugin ensures `%VITE_APP_URL%`
  always resolves — falling back to Cloudflare's built-in `CF_PAGES_URL`
  for preview deployments, then to `""` (which crawlers silently ignore).
- The GitHub Pages deployment hardcodes `VITE_APP_URL=https://safe.dev/safenet`
  in the workflow.
- The preview image lives in `explorer/public/` so Vite serves it at a
  stable, unversioned path that matches the filename in the meta tags.

### Testing

- Set `VITE_APP_URL` locally and confirmed substituted URLs appear in
  page source via `npm run dev`.
- After deploy: `curl -sL <url> -H "User-Agent: facebookexternalhit/1.1" | grep -i og:image`
- X Card Validator: https://cards-dev.twitter.com/validator
- Facebook Sharing Debugger: https://developers.facebook.com/tools/debug/


explorer/.env.sample
```diff
@@ -29,6 +29,10 @@
 # Default signing timeout in blocks.
 # VITE_DEFAULT_SIGNING_TIMEOUT=12
 
+# Canonical public URL of the app (no trailing slash). Used in og:url and og:image meta tags
+# so social crawlers resolve the preview image correctly from any deployment host (e.g. IPFS gateways).
+# VITE_APP_URL=
```
rmeissner, line 35: nit: currently we have no required parameters. Therefore I would assume this one is also not required. That being said, without this being set, it will not work right? Maybe something to mention in the comment.

rmeissner (approved):

rmeissner:
Should we set the `VITE_APP_URL` for the preview (so staging) build?

explorer/.env.sample
tschubotz: It's now just defaulting to empty if not set. and therefore not required anymore

tschubotz:
> Should we set the `VITE_APP_URL` for the preview (so staging) build?

yes, I've set it now for the preview and for the build to safe.dev/safenet

PR #399, opened 2026-05-25
feat(contracts/test): create `FROSTMath` library and refactor `FROSTCoordinator.t.sol`

remedcu:
Introduces `FROSTMath`, a stateless test utility library for FROST math, and refactors `FROSTCoordinator.t.sol`.

### Context and Motivation

- Pure math helpers (polynomial evaluation, Lagrange coefficients, ECDH) have no dependency on test state and belong in a reusable library rather than an abstract base contract.

### Decisions and Tradeoffs

- `FROSTMath` functions are parameterised by participant `address` values, not `ParticipantMerkleTree` indices, so they can be reused by any test contract without coupling to a specific test's state.

### Testing

- `forge test --match-contract FROSTCoordinatorTest` - all tests pass.
- `forge fmt --check` and `forge lint --deny notes` pass.

### DIFF
--- a/contracts/test/FROSTCoordinator.t.sol
+++ b/contracts/test/FROSTCoordinator.t.sol
@@ -4,9 +4,9 @@ pragma solidity ^0.8.30;
 import {Test, Vm} from "@forge-std/Test.sol";
 import {Arrays} from "@oz/utils/Arrays.sol";
 import {MerkleProof} from "@oz/utils/cryptography/MerkleProof.sol";
-import {Math} from "@oz/utils/math/Math.sol";
 import {CommitmentShareMerkleTree} from "@test/util/CommitmentShareMerkleTree.sol";
 import {ForgeSecp256k1} from "@test/util/ForgeSecp256k1.sol";
+import {FROSTMath} from "@test/util/FROSTMath.sol";
 import {ParticipantMerkleTree} from "@test/util/ParticipantMerkleTree.sol";
 import {FROSTCoordinator} from "@/FROSTCoordinator.sol";
 import {FROST} from "@/libraries/FROST.sol";
@@ -16,7 +16,6 @@ import {Secp256k1} from "@/libraries/Secp256k1.sol";
 
 contract FROSTCoordinatorTest is Test {
     using Arrays for address[];
-    using Arrays for uint256[];
     using ForgeSecp256k1 for ForgeSecp256k1.P;
 
     struct Nonces {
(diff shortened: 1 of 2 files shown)


contracts/test/util/FROSTCoordinatorTestBase.sol
```diff
@@ -0,0 +1,287 @@
...
+        {
+            bytes32[] memory commitments = new bytes32[](COUNT);
+            for (uint256 i = 0; i < COUNT; i++) {
+                Nonces memory n = nonces[i];
+                uint256 d = FROST.nonce(bytes32(vm.randomUint()), s[i]);
+                n.d = ForgeSecp256k1.g(d);
+                uint256 e = FROST.nonce(bytes32(vm.randomUint()), s[i]);
+                n.e = ForgeSecp256k1.g(e);
+                // forge-lint: disable-next-line(asm-keccak256)
+                bytes32 leaf = keccak256(abi.encode(0, n.d.x(), n.d.y(), n.e.x(), n.e.y()));
```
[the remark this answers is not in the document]
remedcu: Good catch. Fixed in 4b75724 — `offset` is now derived from `coordinator.groupSignCount(gid) & 0x3ff` before calling `sign()`, matching the `sequence & _OFFSETMASK` the coordinator uses in `signRevealNonces`.

---

contracts/test/util/FROSTCoordinatorTestBase.sol
```diff
@@ -0,0 +1,290 @@
...
+        participants = new ParticipantMerkleTree(_randomSortedAddresses(COUNT));
+    }
+
+    /// @dev Trusted dealer key generation. Returns the group ID, per-participant
+    /// secret keys, and the group private key scalar (for debugging).
+    function _trustedKeyGen(bytes32 context) internal returns (FROSTGroupId.T gid, uint256[] memory s, uint256 gs) {
+        s = new uint256[](COUNT);
+
+        uint256[] memory a = new uint256[](THRESHOLD);
+        for (uint256 j = 0; j < THRESHOLD; j++) {
```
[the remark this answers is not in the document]
remedcu: Fixed in 30fbda7 — `a[0]` now samples from `[1, N-1]`. The other polynomial coefficients remain unrestricted since they don't affect the group key directly.

---

contracts/test/util/FROSTCoordinatorTestBase.sol
```diff
@@ -0,0 +1,314 @@
...
+import {CommitmentShareMerkleTree} from "@test/util/CommitmentShareMerkleTree.sol";
+import {ForgeSecp256k1} from "@test/util/ForgeSecp256k1.sol";
+import {ParticipantMerkleTree} from "@test/util/ParticipantMerkleTree.sol";
+import {FROSTCoordinator} from "@/FROSTCoordinator.sol";
+import {FROST} from "@/libraries/FROST.sol";
+import {FROSTGroupId} from "@/libraries/FROSTGroupId.sol";
+import {FROSTSignatureId} from "@/libraries/FROSTSignatureId.sol";
+import {Secp256k1} from "@/libraries/Secp256k1.sol";
+
+abstract contract FROSTCoordinatorTestBase is Test {
```
rmeissner: nit: if we follow the code patterns from the prod contracts, then we should use libraries for this.

rmeissner (commented):
The branch that you are reviewing against is not fully merged. I would request that either the target branch is merged first, or that it is rebased towards a merged branch (i.e. `main` in this case), before marking it as ready for review.

If you want to stack the changes first you can use `stack/...` as a base branch, but imo this not really required here.

nlordell (commented):
question: Up until now, we've roughly kept 1 .sol file == 1 .t.sol file. I don't understand why we split the files this way, instead of using a single test file for the whole Coordinator contract. If this was a past decision that was merged and I missed, can someone please point me to the discussion around this decision?

request change: I find the PR description to be very noisy. The motivation behind the change is quite short (deduplicate test code), but there is a BUNCH of extra (IMO) unnecessary context that just makes the whole thing longer than I believe it should be.

remedcu:
> I don't understand why we split the files this way, instead of using a single test file for the whole Coordinator contract.

The only reason to split it was to make the review easier. The decision was only to split the PR (not to split into different files). Subjectively, after having a shared base, writing different files for different sections looked cleaner. Don't have a very strong preference to either though.

nlordell:
> The only reason to split it was to make the review easier.

Adding a new file with duplicated code is a lot more noise to review IMO. I would assume the typical path is:
1. Refactor test code to make it more reusable for upcoming tests
2. Add new new tests with the refactored code

Doing it the other way around add a lot more review pressure.

remedcu:
> Adding a new file with duplicated code is a lot more noise to review IMO. I would assume the typical path is:
> 
> 1. Refactor test code to make it more reusable for upcoming tests
> 2. Add new new tests with the refactored code

Exactly, using the changes in this PR I will be creating the next PRs: #400 #401 #402 and #403 where you can see they only check the new part, the rest is shared among these test with this base.

The only duplicated code is the import, which IMO is fine.

nlordell:
> The only duplicated code is the import, which IMO is fine.

`contracts/test/FROSTCoordinatorDecline.t.sol` already exists for this PR - I was arguing it shouldn't. There exists duplicated imports, and dummy key generation.

> The only reason to split it was to make the review easier.

I would argue that having the following added to `FROSTCoordinator.t.sol` is much easier to review than adding a whole new test file...

```solidity
    function test_SignDecline_ThresholdReached_EmitsSignRejected() public {
        FROSTGroupId.T gid = _keyGen();
        FROSTSignatureId.T sid = coordinator.sign(gid, keccak256("msg"));

        for (uint256 i = 0; i < DECLINE_THRESHOLD - 1; i++) {
            vm.prank(participants.addr(i));
            assertFalse(coordinator.signDecline(sid));
        }

        vm.expectEmit();
        emit FROSTCoordinator.SignRejected(sid);
        vm.prank(participants.addr(DECLINE_THRESHOLD - 1));
        bool rejected = coordinator.signDecline(sid);
        assertTrue(rejected);
    }
```

contracts/test/util/FROSTCoordinatorTestBase.sol
remedcu: The library pattern (`struct T` + `using Foo for Foo.T`) is the right choice when a piece of code is stateless and pure, like `ForgeSecp256k1`. `FROSTCoordinatorTestBase` is different in three ways that make that pattern impractical here:

1. **It holds state.** `coordinator` and `participants` are storage fields initialized in `setUp()`. A library would need them passed as a `T storage self` parameter to every helper call, which is significantly more verbose with no readability gain.
2. **It uses `vm` via `Test` inheritance.** `vm` is not a global — it is exposed through `StdCheats`, which is part of the `Test` inheritance chain. A library workaround using `Vm constant _VM = Vm(0x7109...)` is technically possible but non-idiomatic, and every `vm.prank` / `vm.randomUint` / `vm.shuffle` would need to go through that constant.
3. **It calls `assertEq`.** Assertion functions live in `StdAssertions`, which a library cannot inherit. `_trustedKeyGen` uses `assertEq` to verify the derived group public key matches the expected value.

`MerkleTreeBase` (the other test utility in this directory) follows the same abstract contract pattern for the same reasons. The library pattern is reserved for stateless helpers that don't need the `Test` infrastructure.

---

contracts/test/FROSTCoordinatorDecline.t.sol
```diff
@@ -1,80 +0,0 @@
-// SPDX-License-Identifier: GPL-3.0-only
```
rmeissner: request change: Please use `main` as the target branch, then this file doesn't have to be deleted, as it doesn't exist

contracts/test/FROSTCoordinator.t.sol
```diff
@@ -4,42 +4,287 @@ pragma solidity ^0.8.30;
...
-
         coordinator = new FROSTCoordinator();
         participants = new ParticipantMerkleTree(_randomSortedAddresses(COUNT));
     }
 
+    // ========================= Trusted Setup Helpers =========================
+
+    /// @dev Trusted dealer key generation. Returns the group ID, per-participant
+    /// secret keys, and the group private key scalar (for debugging).
+    function _trustedKeyGen(bytes32 context) internal returns (FROSTGroupId.T gid, uint256[] memory s, uint256 gs) {
```
rmeissner, line 44: I might have missed some of @nlordell comments, but I am not sure anymore what the change here is? Where methods reordered? 

My expectation would have been that if the primary purpose is to introduce FROSTMath, that we see one new file and a couple line edits where this new file is used

nlordell: request change: In many of our files (Staking.sol, FROSTCoordinator.sol, Consensus.sol, libraries/*.sol), we put internal/private helpers at the bottom of the file. Here, it was moved to the top without justification, adding noise to the PR. I think it should go back (unless there is a good reason for moving the function around).

nlordell (changes requested):
Math helpers LGTM!

Request that code reorganization doesn't happen here (at least without a reason and some justification which is missing from the PR description).

remedcu:
Closing as the initial idea was to create a base and share among different test files, but later it was decided not to have different test files for a single `.sol` file.

nlordell:
I mean, splitting the math to its own file was not a bad idea...

PR #423, opened 2026-05-28
feat(validator): call signDecline when transaction is rejected

remedcu:
Validators that reject a transaction now call `signDecline` on-chain instead of silently skipping the signing round.

### Context and Motivation

The coordinator had no way to distinguish a validator that was offline from one that explicitly refused to sign, making threshold tracking incomplete and leaving operators blind to explicit rejections.

### Decisions and Tradeoffs

- A `waiting_to_decline` state is introduced rather than calling `signDecline` immediately on rejection, because the `signDecline` call requires the signing session ID, which only arrives with the coordinator's `Sign` event.
- Oracle transactions that fail verification still drop silently — their reorg/timing semantics differ enough that the same treatment isn't obviously correct. This should be decided before the feature is considered complete.

---

## CLA signature

---
_Generated by [Claude Code](https://claude.ai/code/session_01NQP8LGdd3sws86RWJiuWqg)_


validator/src/machine/consensus/transactionProposed.ts
```diff
@@ -38,9 +39,18 @@ export const handleTransactionProposed = async (
...
+			signing: [
+				message,
+				{
+					id: "waiting_to_decline",
+					packet,
+					deadline: event.block + machineConfig.signingTimeout,
+				},
+			],
+		};
 	}
```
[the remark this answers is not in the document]
remedcu: The packet reaching this branch has already been ABI-decoded and passed through Zod schema validation, so all fields are well-typed before `safeTxPacketHash` is called. `hashTypedData` does not throw on well-typed inputs. Adding a catch-all here would mask real bugs and goes against the codebase guideline of not adding error handling for scenarios that cannot happen at an already-validated internal boundary.

---

validator/src/machine/consensus/transactionProposed.ts
```diff
@@ -38,9 +39,18 @@ export const handleTransactionProposed = async (
 	const result = await verificationEngine.verify(packet);
 	const span = { epoch: event.epoch, safeTxHash: event.safeTxHash };
 	if (result.status === "invalid") {
-		// Invalid packet, don't update state
+		const message = safeTxPacketHash(packet);
 		logger?.info?.(`Invalid transaction packet: ${result.error.message}`, span);
```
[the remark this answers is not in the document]
remedcu: The packet reaching this branch has already been ABI-decoded and passed through Zod schema validation, so all fields are well-typed before `safeTxPacketHash` is called. `hashTypedData` does not throw on well-typed inputs. Adding a catch-all here would mask real bugs and goes against the codebase guideline of not adding error handling for scenarios that cannot happen at an already-validated internal boundary.

---

validator/src/machine/storage/schemas.ts
```diff
@@ -167,13 +167,19 @@ const waitingForAttestationSchema = z.object({
...
+const waitingToDeclineSchema = z.object({
+	id: z.literal("waiting_to_decline"),
+	deadline: coercedBigIntSchema,
+});
+
 export const signingStateSchema = z.intersection(
 	baseSigningStateSchema,
 	z.discriminatedUnion("id", [
 		waitingForRequestSchema,
 		collectNonceCommitmentsSchema,
```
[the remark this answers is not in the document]
remedcu: Good catch — this is a pre-existing gap unrelated to this PR. Tracked in #425 to keep this PR focused.

---

nlordell (approved):
LGTM, but I would wait for @rmeissner's review first.

validator/src/machine/consensus/transactionProposed.ts
```diff
@@ -38,9 +39,18 @@ export const handleTransactionProposed = async (
 	const result = await verificationEngine.verify(packet);
 	const span = { epoch: event.epoch, safeTxHash: event.safeTxHash };
 	if (result.status === "invalid") {
-		// Invalid packet, don't update state
+		const message = safeTxPacketHash(packet);
```
rmeissner: request change: I don't think this should be calculated here, rather incoperate this in the verification engine response

validator/src/machine/signing/sign.ts
```diff
@@ -21,6 +21,13 @@ export const handleSign = async (
 		logger?.debug?.(`Not part of signing group ${event.gid}!`);
 		return {};
 	}
+	const status = machineStates.signing[event.message];
+	if (status?.id === "waiting_to_decline") {
```
rmeissner: request change: this should happen after `checkAvailableNonces` as we always want to check (at least that is the current implementation)

validator/src/machine/signing/timeouts.ts
```diff
@@ -177,6 +177,11 @@ const checkSigningRequestTimeout = (
 				signing: [message, undefined],
 			};
 		}
+		case "waiting_to_decline": {
```
rmeissner, line 175: nit: I would combine this with the entry above

rmeissner (changes requested):

validator/src/machine/consensus/transactionProposed.ts
remedcu: Addressed — split `PacketHandler.hashAndVerify` into `hash` (pure, sync, returns `Hex`) and `verify` (async, throws if invalid). The engine now calls `handler.hash(packet)` before the try/catch and includes `packetId` in both valid and invalid `PacketVerificationResult`. `transactionProposed` drops the `safeTxPacketHash` import and reads `result.packetId` directly.

---

validator/src/machine/signing/sign.ts
remedcu: Addressed — `checkAvailableNonces` now runs before the `waiting_to_decline` check, and its diff is spread into the decline return value so nonce replenishment always happens when a Sign event arrives.

---

validator/src/machine/signing/timeouts.ts
remedcu: Done — combined into a single fall-through case.

---

validator/src/machine/signing/sign.ts
```diff
@@ -29,7 +30,13 @@ export const handleSign = async (
 		event.sequence,
 		logger,
 	);
-	const status = machineStates.signing[event.message];
```
rmeissner, line 32: nit: why was this line moved?

validator/src/consensus/verify/safeTx/handler.ts
```diff
@@ -23,6 +28,5 @@ export class SafeTransactionHandler implements PacketHandler<SafeTransactionPack
 			this.metrics?.transactionChecks.labels({ result: label }).inc();
 			throw error;
```
rmeissner, line 29: nit: hmmm I feel like I am too opinionated here, but personally I would have found it cleaner to convert this method to return the hash and an error, rather than splitting the method.

Because now we have to parse the packet twice, which feels unnecessary. But I would keep this on the level of a nit and not block the PR.

rmeissner (approved):

validator/src/machine/signing/sign.ts
remedcu: Reverted 👍🏾

PR #424, opened 2026-05-28
feat(explorer): show declined validators in attestation status

remedcu:
The attestation breakdown now shows which validators explicitly declined to sign a transaction.

### Context and Motivation

Operators had no visibility into explicit rejections — a validator refusing to sign looked identical to one that was simply offline. The "Declined" row makes this distinction visible at a glance.

### Decisions and Tradeoffs

- A participant who has a `SignShared` event takes priority over any `SignDeclined` from the same address, since a later successful sign supersedes an earlier decline. Repeated decline events from the same participant are also deduplicated to avoid inflating the count.

---

## CLA signature

---
_Generated by [Claude Code](https://claude.ai/code/session_01NQP8LGdd3sws86RWJiuWqg)_

### DIFF
--- a/explorer/src/components/transaction/SafeTxAttestationStatus.tsx
+++ b/explorer/src/components/transaction/SafeTxAttestationStatus.tsx
@@ -49,10 +49,10 @@ export function SafeTxAttestationStatus({ proposal }: { proposal: TransactionPro
 							<p className={"ml-4"}>Declined:</p>
 							<p>
 								<ValidatorList
-									all={allValidatorIds}
+									all={status.data.declined.map((s) => s.address)}
 									active={status.data.declined.map((s) => s.address)}
-									mapInfo={mapInfo}
-									completed={false}
+									mapInfo={(suffix) => mapInfo(suffix === "✅" ? "❌" : suffix)}
+									completed={true}
 								/>
 							</p>
 						</div>
--- a/explorer/src/components/transaction/SafeTxAttestationStatus.tsx
+++ b/explorer/src/components/transaction/SafeTxAttestationStatus.tsx
@@ -44,6 +44,19 @@ export function SafeTxAttestationStatus({ proposal }: { proposal: TransactionPro
 							/>
 						</p>
 					</div>
+					{status.data.declined.length > 0 && (
+						<div className={"md:flex md:justify-between"}>
+							<p className={"ml-4"}>Declined:</p>
+							<p>
+								<ValidatorList
+									all={allValidatorIds}
+									active={status.data.declined.map((s) => s.address)}
+									mapInfo={mapInfo}
+									completed={false}
+								/>
+							</p>
+						</div>
+					)}
 				</div>
 			)}
 		</>
--- a/explorer/src/lib/coordinator/signing.test.ts
+++ b/explorer/src/lib/coordinator/signing.test.ts
@@ -9,11 +9,11 @@ import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 //
 // Each test resets the module so the module-level address cache is cleared.
 
-const CONSENSUS = "0x1111111111111111111111111111111111111111" as Address;
-const COORDINATOR_FROM_GETTER = "0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" as Address;
+const CONSENSUS: Address = "0x1111111111111111111111111111111111111111";
+const COORDINATOR_FROM_GETTER: Address = "0xAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA";
 
 // Valid 32-byte hex (required by safeTxProposalHash's hashTypedData)
-const SAFE_TX_HASH = `0x${"ab".repeat(32)}` as `0x${string}`;
+const SAFE_TX_HASH: `0x${string}` = `0x${"ab".repeat(32)}`;
 
 type LoadLatestAttestationStatus = typeof import("./signing").loadLatestAttestationStatus;
 
@@ -28,14 +28,19 @@ const loadModule = async () => {
 // the coordinator address is resolved.
 const makeProvider = ({
 	readContractImpl,
+	getLogsImpl,
+	requestImpl,
 }: {
 	readContractImpl?: (args: { functionName: string }) => unknown;
-}): PublicClient =>
+	getLogsImpl?: (args: unknown) => unknown;
+	requestImpl?: (args: unknown) => unknown;
+} = {}): PublicClient =>
 	({
 		getBlockNumber: vi.fn().mockResolvedValue(10000n),
 		getChainId: vi.fn().mockResolvedValue(1),
-		getLogs: vi.fn().mockResolvedValue([]),
+		getLogs: getLogsImpl ? vi.fn(getLogsImpl) : vi.fn().mockResolvedValue([]),
 		readContract: readContractImpl ? vi.fn(readContractImpl) : vi.fn(),
+		request: requestImpl ? vi.fn(requestImpl) : vi.fn().mockResolvedValue([]),
 	}) as unknown as PublicClient;
 
 const baseArgs = {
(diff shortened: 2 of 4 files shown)


explorer/src/components/transaction/SafeTxAttestationStatus.tsx
```diff
@@ -44,6 +44,19 @@ export function SafeTxAttestationStatus({ proposal }: { proposal: TransactionPro
...
+					{status.data.declined.length > 0 && (
+						<div className={"md:flex md:justify-between"}>
+							<p className={"ml-4"}>Declined:</p>
+							<p>
+								<ValidatorList
+									all={allValidatorIds}
+									active={status.data.declined.map((s) => s.address)}
+									mapInfo={mapInfo}
+									completed={false}
+								/>
```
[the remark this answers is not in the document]
rmeissner: I agree that it would be cleaner to combine all of the states into one visual item. I.e. committed gets 📝 , attested gets ✅, declined gets ❌ and pending is ❔ (or something similar).

That being said this will change the scope of this PR so maybe it is worth only doing the logic part here and not the visual representation.

explorer/src/lib/coordinator/signing.test.ts
```diff
@@ -111,3 +112,159 @@ describe("loadCoordinator (via loadLatestAttestationStatus)", () => {
...
 	});
 });
+
+// Selectors for coordinator signing progress events
+const SIGN_SELECTOR = "0xb48d242879f9f3df555c800db966f65cba128c7213198748fa202ed54e092691";
+const SIGN_DECLINED_SELECTOR = "0xe6d872ab6c2f6512d506498a50ba8ba1bbc2bb3c19439ad7c6ff3d74465277d7";
+const SIGN_SHARED_SELECTOR = "0x25a4d6e8d11a9fdc20ffdd826473485ae4cdd453271726c072a16836c1882e7c";
+
+const SID = `0x${"aa".repeat(32)}` as `0x${string}`;
+const SID_PADDED = SID; // bytes32 is already padded
```
rmeissner: request change: why do we have to const with the same value?

explorer/src/lib/coordinator/signing.test.ts
```diff
@@ -111,3 +112,159 @@ describe("loadCoordinator (via loadLatestAttestationStatus)", () => {
 		expect(provider.getLogs).toHaveBeenCalledWith(expect.objectContaining({ address: COORDINATOR_FROM_GETTER }));
 	});
 });
+
+// Selectors for coordinator signing progress events
+const SIGN_SELECTOR = "0xb48d242879f9f3df555c800db966f65cba128c7213198748fa202ed54e092691";
+const SIGN_DECLINED_SELECTOR = "0xe6d872ab6c2f6512d506498a50ba8ba1bbc2bb3c19439ad7c6ff3d74465277d7";
+const SIGN_SHARED_SELECTOR = "0x25a4d6e8d11a9fdc20ffdd826473485ae4cdd453271726c072a16836c1882e7c";
+
+const SID = `0x${"aa".repeat(32)}` as `0x${string}`;
```
rmeissner: request change: please use type declaration instead of type overriding

Note: this should be a general rule, as this was already commented on another PR. So far we did not document this explicitly because for me this is best practice when working with types. Type overrides defeat the purpose of type checking as they **override** the type, therefore they should be used sparingly.

explorer/src/lib/coordinator/signing.test.ts
```diff
@@ -111,3 +112,159 @@ describe("loadCoordinator (via loadLatestAttestationStatus)", () => {
...
+const SID_PADDED = SID; // bytes32 is already padded
+const SELECTION_ROOT = `0x${"bb".repeat(32)}` as `0x${string}`;
+const PARTICIPANT_A = "0xaAaAaAaaAaAaAaaAaAAAAAAAAaaaAaAaAaaAaaAa" as Address;
+const PARTICIPANT_B = "0xbBbBBBBbbBBBbbbBbbBbbbbBBbBbbbbBbBbbBBbB" as Address;
+// Address padded to 32 bytes (for use as topic)
+const padAddress = (addr: string) => `0x${"00".repeat(12)}${addr.slice(2).toLowerCase()}`;
+
+// Builds a Sign-initiated raw log
+const makeSignLog = (_sid: string) => ({
+	address: COORDINATOR_FROM_GETTER,
```
rmeissner, line 136: nit: maybe a base `makeLog` method would be helpful

explorer/src/lib/coordinator/signing.test.ts
```diff
@@ -111,3 +112,159 @@ describe("loadCoordinator (via loadLatestAttestationStatus)", () => {
...
+	removed: false,
+	topics: [SIGN_SHARED_SELECTOR, sid, SELECTION_ROOT],
+	// participant is non-indexed, encoded in data alongside z — but viem's parseEventLogs
+	// reads from data, so we pack participant (padded to 32) + z (uint256 = 0)
+	transactionHash: `0x${"dd".repeat(32)}`,
+	transactionIndex: "0x0",
+});
+
+// Full mock provider for declined-behavior tests (Sign event present)
+const makeFullProvider = ({ progressLogs = [] }: { progressLogs?: unknown[] }): PublicClient => {
```
rmeissner, line 183: request change: we already have a makeProvider, why not extend that instead of defining a new function (i.e. by allowing to pass in specific override mocks)

explorer/src/lib/coordinator/signing.test.ts
```diff
@@ -111,3 +112,159 @@ describe("loadCoordinator (via loadLatestAttestationStatus)", () => {
...
+					gid: `0x${"00".repeat(32)}`,
+					message: SAFE_TX_HASH,
+					sid: SID,
+					sequence: 0n,
+				},
+				eventName: "Sign",
+				logIndex: 0,
+			},
+		]),
+		// provider.request handles eth_getLogs for progress events
```
rmeissner: nit: this comments is confusing. It because for more fine grained control we use request to fetch logs in our code, therefore it is necessary to mock that.

explorer/src/lib/coordinator/signing.ts
```diff
@@ -268,6 +279,24 @@ export const loadGroupPublicKey = async (
 	}
 };
 
+const getDeclined = (status: StatusAggregation | undefined): AttestationParticipation[] => {
+	if (status === undefined || status.declined.length === 0) return [];
+	// Participants who signed (any selection) take priority over decline events
+	const signedAddresses = new Set([
```
rmeissner: request change: committed should also have priority over declined

rmeissner: nit: it might be easier to build this "signed set" when aggregating the events

rmeissner (changes requested):
Note: for the request change related to the test we can discuss how blocking these are

explorer/src/components/transaction/SafeTxAttestationStatus.tsx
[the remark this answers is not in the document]
remedcu: Fixed — the `ValidatorList` for declined validators now passes only the declined addresses as both `all` and `active`, and maps `✅` → `❌` so declined validators render with the correct icon without listing non-declined validators.

---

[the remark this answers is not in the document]
remedcu: Agreed — combining all states into a single list with per-state icons would be a cleaner UX. Keeping it out of this PR to stay focused on the decline tracking feature. Happy to follow up in a separate PR.

---

explorer/src/lib/coordinator/signing.test.ts
remedcu: Removed — `SID` is used directly everywhere now.

---

explorer/src/lib/coordinator/signing.test.ts
remedcu: Changed to type declarations throughout the new constants (`SID`, `SELECTION_ROOT`, `PARTICIPANT_A`, `PARTICIPANT_B`).

---

explorer/src/lib/coordinator/signing.test.ts
remedcu: Added — `makeLog(topics, data, blockNumber)` holds the shared fields; each builder spreads from it.

---

explorer/src/lib/coordinator/signing.test.ts
remedcu: `makeProvider` now accepts optional `getLogsImpl` and `requestImpl` overrides; `makeFullProvider` is a thin wrapper around it.

---

explorer/src/lib/coordinator/signing.test.ts
remedcu: Updated — the comment now explains that `provider.request` is mocked because the production code calls it directly with `eth_getLogs` for fine-grained topic filtering.

---

explorer/src/lib/coordinator/signing.ts
remedcu: Both addressed. `priorityAddresses: Set<Address>` is now populated during aggregation as `SignRevealedNonces` and `SignShared` events are processed, and `getDeclined` uses it directly — eliminating the post-processing reconstruction. A new test covers the committed-takes-priority case.

---

remedcu: Addressed — the set is now built incrementally during aggregation rather than reconstructed in `getDeclined`.

---

rmeissner (approved):

PR #429, opened 2026-05-29
fix(validator): poll for epoch staged logs instead of asserting immediately

remedcu:
### Context and Motivation

`waitForBlock` synchronises with the chain clock but not with the validator's off-chain processing. Two integration tests queried `CONSENSUS_EPOCH_STAGED` logs immediately after the block target was reached and asserted on the count. Because the validator emits those events asynchronously, the assertion could fire before the staging transaction landed — causing an intermittent `expected 1 to be 2` failure in CI (run [78498612298](https://github.com/safe-research/safenet/actions/runs/26636662248/job/78498612298), passed on retry [78499193497](https://github.com/safe-research/safenet/actions/runs/26636662248/job/78499193497)).

### Decisions and Tradeoffs

- Added `waitForLogs` to `src/__tests__/utils.ts`: a `vi.waitFor` wrapper that polls `getLogs` until at least `minCount` matching entries appear, using the same 20 s timeout as `waitForBlock`.
- Replaced the two racy `getLogs` + `expect(length).toBe(...)` patterns for `CONSENSUS_EPOCH_STAGED` with `waitForLogs`. The `CONSENSUS_ORACLE_TRANSACTION_PROPOSED` pattern is emitted by the test's own on-chain call so it is not racy and left unchanged.

---

## CLA signature

---
_Generated by [Claude Code](https://claude.ai/code/session_01PWUS6v3bbjsYH1XFuX2c6u)_

### DIFF
--- a/validator/src/consensus/integration.test.ts
+++ b/validator/src/consensus/integration.test.ts
@@ -96,12 +96,12 @@ describe("integration", () => {
 		rotateOutEpoch,
 		oracleTimeout,
 	}: {
-			blocksPerEpoch?: bigint;
-			timeout?: bigint;
-			blockTimeMs?: number;
-			rotateOutEpoch?: bigint;
-			oracleTimeout?: bigint;
-		} = {}) => {
+		blocksPerEpoch?: bigint;
+		timeout?: bigint;
+		blockTimeMs?: number;
+		rotateOutEpoch?: bigint;
+		oracleTimeout?: bigint;
+	} = {}) => {
 		// Check deployment information is available
 		const deploymentInfoFile = path.join(
 			process.cwd(),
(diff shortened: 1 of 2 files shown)


nlordell (changes requested):
One concern I have is that waiting can hide bugs (where the tests run for a whole other epoch instead of failing as expected).

Staging _must_ happen before rollover, without a staging log means that one was missed and the test should fail.

validator/src/consensus/integration.test.ts
```diff
@@ -291,8 +291,9 @@ describe("integration", () => {
 				clients[2].service.stop();
 			},
 		});
-		// Wait for end of epoch
-		await waitForBlock(testClient, 40n);
+		// Wait a couple of blocks past the epoch boundary so the validator's
+		// final staging transaction is fully indexed before we assert
+		await waitForBlock(testClient, 42n);
```
nlordell, line 296: I don't think this works... The staging event has to come in before block 40, otherwise there will be no staging event.

remedcu:
Not sure what might be a possible solution for this right now, so closing.

PR #436, opened 2026-06-02
SentinelOracle - Phase 2.5: Adjust Fee Payment

rmeissner:
### Context and Motivation

Task in Phase 2.5 of the Sentinel Oracle ([Feature Spec](https://github.com/safe-research/safenet/blob/stack/sentinels_game/features/2026_05_05_transaction_checker_oracle_v1.md#phase-25--fee-payment-fix-pr-25-depends-on-phase-1)).

So far the fee was pulled from the Consensus contract. This pushed the fee tracking responsibility to that contract. As the Oracle already tracks payments, the decision was made to trust the Consensus and pull the fee from the proposer specified by it.

Also to allow future passing of Oracle specific data a `bytes oracleData` was added. This data is passed to the Oracle and can be decoded there to handle additional parameters (i.e. reward token or different reward amounts).

---

## CLA signature

### DIFF
--- a/contracts/src/SentinelOracle.sol
+++ b/contracts/src/SentinelOracle.sol
@@ -61,8 +61,6 @@ contract SentinelOracle is IOracle {
     error InvalidAddress();
     error ZeroFee();
     error SentinelNotActive();
-    error InvalidRewardToken();
-    error InvalidRewardAmount();
 
     // ============================================================
     // MODIFIERS
@@ -107,13 +105,8 @@ contract SentinelOracle is IOracle {
     // IOracle IMPLEMENTATION
     // ============================================================
 
-    function postRequest(bytes32 requestId, address proposer, address rewardToken, uint256 rewardAmount)
-        external
-        override(IOracle)
-    {
+    function postRequest(bytes32 requestId, address proposer, bytes calldata) external override(IOracle) {
         require(msg.sender == CONSENSUS, NotConsensus());
-        require(rewardToken == address(FEE_TOKEN), InvalidRewardToken());
-        require(rewardAmount == REQUEST_FEE, InvalidRewardAmount());
         uint256 fee = REQUEST_FEE;
         uint256 bondTarget = fee * $bondConfig.currentMultiplier();
         uint256 deadline = block.number + VOTING_WINDOW;
--- a/contracts/src/SentinelOracle.sol
+++ b/contracts/src/SentinelOracle.sol
@@ -61,6 +61,8 @@ contract SentinelOracle is IOracle {
     error InvalidAddress();
     error ZeroFee();
     error SentinelNotActive();
+    error InvalidRewardToken();
+    error InvalidRewardAmount();
 
     // ============================================================
     // MODIFIERS
@@ -105,13 +107,18 @@ contract SentinelOracle is IOracle {
     // IOracle IMPLEMENTATION
     // ============================================================
 
-    function postRequest(bytes32 requestId) external override(IOracle) {
+    function postRequest(bytes32 requestId, address proposer, address rewardToken, uint256 rewardAmount)
+        external
+        override(IOracle)
+    {
         require(msg.sender == CONSENSUS, NotConsensus());
+        require(rewardToken == address(FEE_TOKEN), InvalidRewardToken());
+        require(rewardAmount == REQUEST_FEE, InvalidRewardAmount());
         uint256 fee = REQUEST_FEE;
         uint256 bondTarget = fee * $bondConfig.currentMultiplier();
         uint256 deadline = block.number + VOTING_WINDOW;
-        $requests.create(requestId, msg.sender, fee, bondTarget, deadline);
-        FEE_TOKEN.safeTransferFrom(msg.sender, address(this), fee);
+        $requests.create(requestId, proposer, fee, bondTarget, deadline);
+        FEE_TOKEN.safeTransferFrom(proposer, address(this), fee);
     }
 
     // ============================================================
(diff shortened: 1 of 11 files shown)


rmeissner:
Response to the Gemini review:
- `getFee` is only informative and therefore has been left out of this PR to make it easier to review
- SimpleOracle on purpose uses the msg.sender for tracking who submitted the request. This implementation will be removed in a follow up PR
- afaik `z.url()` is a valid zod method and we use it in multiple places already

contracts/src/SentinelOracle.sol
```diff
@@ -105,13 +107,18 @@ contract SentinelOracle is IOracle {
...
     // ============================================================
 
-    function postRequest(bytes32 requestId) external override(IOracle) {
+    function postRequest(bytes32 requestId, address proposer, address rewardToken, uint256 rewardAmount)
+        external
+        override(IOracle)
+    {
         require(msg.sender == CONSENSUS, NotConsensus());
+        require(rewardToken == address(FEE_TOKEN), InvalidRewardToken());
+        require(rewardAmount == REQUEST_FEE, InvalidRewardAmount());
```
nlordell: This is unclear to me: why are we passing in parameters that are required to be some fixed values?

nlordell: As in, based on the PR description, I would have expected the function signature to change to:

```
function postRequest(bytes32 requestId, address proposer)
```

To further clarify based on our PR etiquette:

request change: I would avoid passing in parameters that are fixed values, it doesn't make sense to me. Additionally, the PR description suggested that the only change should be passing in `msg.sender` in order to `transferFrom` the fees.

contracts/src/SimpleOracle.sol
```diff
@@ -65,7 +65,7 @@ contract SimpleOracle is IOracle {
     /**
      * @inheritdoc IOracle
      */
-    function postRequest(bytes32 requestId) external {
+    function postRequest(bytes32 requestId, address, address, uint256) external {
         require($proposers[requestId] == address(0), RequestAlreadyPending());
         $proposers[requestId] = msg.sender;
```
nlordell: ~~request change~~ nit:
```suggestion
        $proposers[requestId] = proposer;
```

**Edit**: this does mean that the contract now technically requires authentication for correctness (something like `require(msg.sender == CONSENSUS)`) which might not be worth the effort given that this contract should be removed.

nlordell (changes requested):

contracts/src/SimpleOracle.sol
rmeissner: that was mentioned above: https://github.com/safe-research/safenet/pull/436#issuecomment-4610277631

Can you say why you think this has to be adjusted? In that specific case the address is just a marker, and as mentione once the SentinelOracle is complete we should remove that SimpleOracle implementation

contracts/src/SentinelOracle.sol
rmeissner: These values are fixed in this initial implementation, but they are neither fixed for other oracles, nor do I see that the fee is fixed in an iteration of the sentinel oracle. Adjusting the the interface in the consensus contract each time will create quite some noise, which I would like to avoid.

contracts/src/SimpleOracle.sol
nlordell: I missed the comment. If the implementation would be removed, then it really doesn't matter.

That being said, it also feels super low effort to set this to the proposer, and it helps ensure that the contract matches the original intention of tracking the actual proposer. It also helps validate my mental model of "this should be the proposer that is passed in, and not the caller".

Since the contract will be removed, **I'm fine downgrading this to a `nit`.** I understand that there is an issue that, for correctness, you would need to add authentication to the `SimpleOracle` which does feel like too much additional work for something that should be removed.

contracts/src/SentinelOracle.sol
rmeissner: For documentation, after a call we decided to replace the fee and token parameter with a "oracleData" parameter that can encode the parameters required by a specific oracle. In this case for the current implementation this would be empty as the data is not used, but it can easily be extended without breaking the Oracle interface (therefore doesn't require adjustments to the Consensus contract).

contracts/src/SentinelOracle.sol
```diff
@@ -172,7 +179,10 @@ contract SentinelOracle is IOracle {
         address proposer = req.proposer;
         uint256 slashed = req.resolveDispute(approveWins);
         SentinelOracleRequest.State outcome = req.state;
-        FEE_TOKEN.safeTransfer(ARBITRATOR, slashed);
+        uint256 refundFee = req.fee;
+        req.fee = 0;
+        FEE_TOKEN.safeTransfer(proposer, refundFee);
+        FEE_TOKEN.safeTransfer(ARBITRATOR, slashed - refundFee);
         emit DisputeResolved(requestId, outcome, slashed);
```
remedcu, line 179: Shouldn't this be now?

```suggestion
        emit DisputeResolved(requestId, outcome, slashed - refundFee);
```

rmeissner: Hmm I would argue the slashed amount is still all of this. It is just that the distribution of the slashed amount is different (i.e. part goes to the user and part to the arbitrator).

contracts/src/interfaces/IOracle.sol
```diff
@@ -26,10 +26,12 @@ interface IOracle {
     /**
      * @notice Post a signing request to the oracle for evaluation.
      * @param requestId The EIP-712 hash of the OracleTransactionProposal message.
```
nlordell: comment: Does this need to be enforced by the oracle? Otherwise, this is more of a "proposer-specified" request identifier no?

**NOte**: this is unrelated to the PR, just something I noted in passing.

contracts/test/AlwaysApproveOracle.t.sol
```diff
@@ -21,14 +21,14 @@ contract AlwaysApproveOracleTest is Test {
         emit IOracle.OracleResult(REQUEST_ID, requester, "", true);
 
         vm.prank(requester);
```
nlordell: nit: pranks aren't needed anymore.

contracts/test/AlwaysApproveOracle.t.sol
```diff
@@ -21,14 +21,14 @@ contract AlwaysApproveOracleTest is Test {
...
-        oracle.postRequest(REQUEST_ID);
+        oracle.postRequest(REQUEST_ID, requester, "");
     }
 
     function test_PostRequest_EmitsApprovedTrue() public {
         vm.recordLogs();
 
         vm.prank(requester);
```
nlordell: nit: here as well.

nlordell (approved):
LGTM, some extremely minor nits. One comment (but unrelated to the PR).

remedcu (approved):

PR #443, opened 2026-06-05
Introduce Shared Rust Crate

nlordell:
Since we plan on porting the Safenet services to Rust, the first step is to introduce some shared infrastructure that can be used by both the `validator` and the `sentinel` services.

### Decisions and Tradeoffs

For how, this is just an empty crate with some documentation. Additionally, we will only have _one_ shared crate for now. In the future, we can consider splitting things into additional logical parts to allow for better code reuse (for example, a crate with Safenet Consensus specific types and hashing -- which can be used by projects outside of this monorepo; one for event indexing; one for observability for Safenet services; etc.)

I decided to put the CI into a separate workflow. This is because it is still unclear to me how we will tie this in to a complete "monorepo" build in CI.


crates/core/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "safenet-core"
+version = "0.2.0"
+edition = "2024"
```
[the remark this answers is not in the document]
nlordell: 2024 is stable, and the current default.

Cargo.toml
```diff
@@ -0,0 +1,3 @@
+[workspace]
+resolver = "3"
```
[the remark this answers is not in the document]
nlordell: resolver V3 is the current default.

.github/workflows/rust.yml
```diff
@@ -0,0 +1,20 @@
...
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
```
rmeissner, line 13: nit: I know we don't do it in other places yet, but maybe we could ping the checkout action version by using the concrete commit hash

.gitignore
```diff
@@ -1,14 +1,12 @@
-_build/
```
rmeissner, line 1: why was this one removed? (tbf I don't know if we ever required it 🙈 )

rmeissner (approved):
nit: I would expect that we also add something to the project readme that mentions some details about this shared crate (i.e. how to compile it and that we have 2 workspacing tools: `npm` and `cargo`). But this can also happen in a later PR

Maybe to provide an ongoing overview of all related PR you could create a milestone to collect the PRs for the shared crate.

.gitignore
nlordell: Sorry, should have left a comment:

`_build` is the output folder for `dune` build system that we used with the OCaml spec. It is no longer needed since the spec was removed.

nlordell:
> I would expect that we also add something to the project readme

Ah, good point and nice catch. Will add something.

> Maybe to provide an ongoing overview of all related PR you could create a milestone to collect the PRs for the shared crate.

I will create a ~~tag~~ label :+1:.

**Edit**: here is the label: https://github.com/safe-research/safenet/pulls?q=is%3Apr+label%3Aoxidation

.github/workflows/rust.yml
nlordell: I will update all our workflows to commit hashes separately.

nlordell:
**Update**: moved to milestones as they are a better fit for this kind of classification: https://github.com/safe-research/safenet/milestone/3

PR #444, opened 2026-06-09
Update Agents File

nlordell:
This PR moves CLAUDE.md -> AGENTS.md so that it is read by more agent harnesses than just claude (Codex, Zed, etc.)

### Decisions and Tradeoffs

Additionally, I made some small corrections that will hopefully improve our results with it:

- Include information about the shared Rust crate
- Include missing information about the `examples` NPM workspace package
- De-duplicate information about documentation
- Remove feature planning section in favour of skill in `safe-research/agents` repository
  - The `features/` directory was also removed (see #445).
- Update to our current branch naming policy

### Testing

Unfortunately, this is not considered a "move and edit" in the GitHub diff view. You can view the diff by:

```
diff -u --color <(curl -sL https://raw.githubusercontent.com/safe-research/safenet/refs/heads/main/CLAUDE.md) AGENTS.md
```

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://docs.safefoundation.org/cla).

PR #445, opened 2026-06-09
Remove Old Features

nlordell:
#444 updated our `AGENTS.md` to prefer the `/plan-epic` skill from the `safe-research/agents` repository. As such, we no longer need the `features/` directory (which is now the `epics/` directory from the skill).

### Decisions and Tradeoffs

I also removed two existing feature specs that were in the `features/` directory as they are both (AFAICT) completed.

---

## CLA signature


remedcu (approved):

rmeissner (approved):
Assuming that #444 gets merged I am fine with this change.

Nit: I would have expected the template to be removed in #444, as this is where the instructions referencing the files are removed.

nlordell:
> Nit: I would have expected the template to be removed in https://github.com/safe-research/safenet/pull/444, as this is where the instructions referencing the files are removed.

I will merge this onto #444 so that it gets squashed into 1 commit.

PR #450, opened 2026-06-10
Add Bloom Filter Helper

nlordell:
This PR ports the bloom filter inclusion helper used by the event indexer to check if a block contains events from a set of watched contracts.

### Context and Motivation

While `alloy` provides primitives for dealing with bloom filters, this helper specifically handles the case of `one_of(addresses) && one_of(topics)` which is needed by the event indexer.

### Testing

The PR may seem large at first glance but over 400 lines is just test data from an actual Gnosis Chain block to ensure that the implementation is sound. The test data is the same from the TypeScript bloom filter tests (to make sure the implementations are equivalent).

---

## CLA signature

### DIFF
--- a/crates/core/src/index/bloom.rs
+++ b/crates/core/src/index/bloom.rs
@@ -18,13 +18,17 @@ use alloy::primitives::{Address, B256, Bloom, BloomInput};
 /// a watched address and a different log supplies a watched topic still returns
 /// `true`, matching the conservative behaviour of the reference implementation.
 pub fn may_contain_log(bloom: &Bloom, addresses: &[Address], topics: &[B256]) -> bool {
-    let contains_address = addresses
-        .iter()
-        .any(|address| bloom.contains_input(BloomInput::Raw(address.as_slice())));
-    let contains_topics = topics
-        .iter()
-        .any(|topic| bloom.contains_input(BloomInput::Raw(topic.as_slice())));
-    contains_address && contains_topics
+    let contains_at_least_one_address = || {
+        addresses
+            .iter()
+            .any(|address| bloom.contains_input(BloomInput::Raw(address.as_slice())))
+    };
+    let contains_at_least_one_topic = || {
+        topics
+            .iter()
+            .any(|topic| bloom.contains_input(BloomInput::Raw(topic.as_slice())))
+    };
+    contains_at_least_one_address() && contains_at_least_one_topic()
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


crates/core/src/index/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! Indexing of onchain blocks and events: following the chain head, fetching
+//! event logs in order, and handling chain reorgs.
+
+#[allow(dead_code)]
```
nlordell, line 4: We leave this in for now to silence clippy (the linter) errors, we will remove it once it is no longer needed in a future PR.

crates/core/src/index/mod.rs
```diff
@@ -0,0 +1,5 @@
...
+mod bloom;
```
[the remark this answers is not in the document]
nlordell: This is intentional, as we only plan on using internally in the indexing logic.

crates/core/src/index/bloom.rs
```diff
@@ -0,0 +1,511 @@
...
+/// a watched address and a different log supplies a watched topic still returns
+/// `true`, matching the conservative behaviour of the reference implementation.
+pub fn may_contain_log(bloom: &Bloom, addresses: &[Address], topics: &[B256]) -> bool {
+    let contains_address = addresses
+        .iter()
+        .any(|address| bloom.contains_input(BloomInput::Raw(address.as_slice())));
+    let contains_topics = topics
+        .iter()
+        .any(|topic| bloom.contains_input(BloomInput::Raw(topic.as_slice())));
+    contains_address && contains_topics
```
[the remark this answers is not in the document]
nlordell: Nice small optimization. I changed the implementation to use thunks so it is more descriptive.

crates/core/src/index/bloom.rs
```diff
@@ -0,0 +1,515 @@
...
+    use alloy::{
+        hex,
+        primitives::{address, b256},
+    };
+
+    #[test]
+    fn checks_inclusion() {
+        let bloom = Bloom(
+            hex!(
+                "0000000000000000000000000000000000000000100000000000000000000000
```
rmeissner, line 46: question: Did we reuse existing test data here

rmeissner (approved):

crates/core/src/index/bloom.rs
nlordell: Yes! Added to the PR description.

PR #453, opened 2026-06-11
Add Clock Utility

nlordell:
Implement a wall clock used by the block indexer for waiting in-between blocks.

### Context and Motivation

When indexing blocks, we need to guess when the next block will be produced (based on the block time) and wait for it (to avoid busy polling the RPC node and wasting a bunch of credits).

This PR adds a small `Clock` utility that uses the `tokio` time facilities in order to keep track of "now" as a Unix epoch timestamp. This allows two things:

- We have an idea of what the current system time as a Unix epoch is so we know how much we need to wait in between querying blocks (noting that Ethereum timestamps are all Unix timestamps)
- We can leverage `tokio`'s builtin time testing utilities instead of having to roll our own abstractions.

### Testing

We added a couple unit tests that double check that, indeed, the `Clock` abstraction works with the Tokio test utilities.

---

## CLA signature

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -16,3 +16,6 @@ tokio = { version = "1", features = ["full"] }
 tracing = "0.1"
 tracing-subscriber = { version = "0.3", features = ["env-filter", "fmt", "json"] }
 url = { version = "2", features = ["serde"] }
+
+[dev-dependencies]
+tokio = { version = "1", features = ["test-util"] }
--- a/crates/core/src/index/clock.rs
+++ b/crates/core/src/index/clock.rs
@@ -1,0 +1,83 @@
+//! A wall clock implementation that produces Unix epoch timestamps and is
+//! compatible with `tokio::time::{advance, pause}` in tests.
+
+#![allow(dead_code)]
+
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
+
+/// Estimates the current wall-clock time from elapsed monotonic time.
+#[derive(Clone, Debug)]
+pub struct Clock {
+    base_ms: u64,
+    anchor: tokio::time::Instant,
+}
+
+impl Clock {
+    /// Anchors "now" to the current system time.
+    pub fn start() -> Self {
+        let base_ms = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or_default()
+            .as_millis() as u64;
+        Self {
+            base_ms,
+            anchor: tokio::time::Instant::now(),
+        }
+    }
+
+    /// The estimated current wall-clock time, in milliseconds.
+    pub fn now_ms(&self) -> u64 {
+        self.base_ms + self.anchor.elapsed().as_millis() as u64
+    }
+
+    /// Sleeps until a target Unix epoch time, in milliseconds.
+    ///
+    /// If `target` is in the past, returns immediately.
+    pub async fn sleep_until(&self, target: u64) {
+        let now = self.now_ms();
+        if target > now {
+            tokio::time::sleep(Duration::from_millis(target - now)).await;
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test(start_paused = true)]
+    async fn tokio_test_util_compatibility() {
+        let clock = Clock::start();
+        let start = clock.now_ms();
+        let target = start + 1_000;
+
+        let sleeper = tokio::spawn({
+            let clock = clock.clone();
+            async move {
+                clock.sleep_until(target).await;
+            }
+        });
+
+        tokio::task::yield_now().await;
+        assert!(!sleeper.is_finished());
+
+        tokio::time::advance(Duration::from_millis(999)).await;
+        assert_eq!(clock.now_ms(), start + 999);
+        assert!(!sleeper.is_finished());
+
+        tokio::time::advance(Duration::from_millis(1)).await;
+        sleeper.await.unwrap();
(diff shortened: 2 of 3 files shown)


rmeissner (approved):
The implementation looks good, and I totally understand the testing advantage, but I would love to understand why this pattern is necessary. Something to talk about next week.

Edit: after talking with Gemini about it, there seems to be a potential edge case that a SystemClock that is not 100% synced when you initialized this Clock, will result in a Clock that never syncs (while the SystemClock might at some point sync over NTP).

nlordell:
It is a little complicated, but there are 3 "get `now`" APIs that we have access to:

- [`std::time::SystemTime`](https://doc.rust-lang.org/std/time/struct.SystemTime.html): as the name suggests, this reads the system clock. There is an "epoch" that we can use to calculate the time elapsed since Unix epoch in order to convert to and from Ethereum timestamps. This is what we used to seed our clock.
- [`std::time::Instant`](https://doc.rust-lang.org/std/time/struct.Instant.html): this is a completely opaque concept of "now" and does not have a stable representation. As such, you can only use it to comparatively get durations in between measurements, but cannot create a value that corresponds to some historic point in time such as the Unix epoch.
- [`tokio::time::Instant`](https://docs.rs/tokio/latest/tokio/time/struct.Instant.html): the Tokio version of the exact same API, with support for testing utilities.

Now, there are two ways we can implement a wall clock:
- just use `SystemTime` and convert measurements to durations from the Unix epoch, this is the more straightforward way
- anchor a clock at some instant, where we have an approximation of the timestamp of that anchor, and compute deltas from the anchor with new measurements: this is what I implemented 

Now, the second way is more complicated. The _only_ advantage is that it allows us to use the Tokio test utilities for controlling time which allows us to do fine grained testing on how the block watcher does timing things.

As per Gemini's finding, it is a good point that we should consider; we need a mechanism to deal with system time drift. Alternatively we can use the Tokio clock just for testing. I will propose something in a follow up.

pinky-hands:
# GitHub PR Review: Add Clock Utility (#453)

## Overview
- **PR**: #453 "Add Clock Utility"
- **Author**: nlordell
- **Status**: Open (22 days old)
- **Branch**: pr/core-crate-phase-C2b → pr/core-crate-phase-C2
- **Labels**: validator, Oxidation milestone
- **Changes**: +85 lines, 0 deletions (3 files)
- **Workflow Status**: CI ✅, Docker ✅, Integration ✅, Check ❌ (Rust), Markdown ✅

## Summary

This PR adds a Clock utility to track wall-clock time using elapsed monotonic time, designed for use in the block indexer to avoid busy-polling RPC nodes. The implementation is compatible with Tokio's virtual time testing utilities.

## Code Review

### ✅ Strengths

1. **Clean Abstraction**: The Clock struct provides a well-encapsulated abstraction for time that's compatible with both production and test environments.

2. **Good Documentation**: The code includes clear documentation explaining the purpose and behavior of each method.

3. **Comprehensive Tests**: Two unit tests verify compatibility with Tokio's test utilities (start_paused = true) and correct behavior when sleeping until a past target.

4. **Proper Async Design**: The sleep_until method correctly handles the case where the target time is in the past.

### ⚠️ Medium Priority Findings

#### 1. Potential Timing Drift in sleep_until

**Issue**: The current implementation calculates a relative duration between self.now_ms() (wall-clock time) and the target, then sleeps for that duration. This approach can introduce timing inaccuracies because:

- There's a time gap between calling now_ms() and sleep()
- If the task is preempted during this gap, the sleep duration becomes stale
- System clock adjustments can cause misalignment

**Recommendation**: Use tokio::time::sleep_until with an absolute Instant instead of a relative duration:

```rust
pub async fn sleep_until(&self, target: u64) {
    if target > self.base_ms {
        let deadline = self.anchor + Duration::from_millis(target - self.base_ms);
        tokio::time::sleep_until(deadline).await;
    }
}
```

This approach is more idiomatic and precise.

**Status**: Gemini Code Assist already raised this issue (medium priority) on line 39.

#### 2. System Clock Synchronization

**Issue**: As noted by reviewer rmeissner, there's a potential edge case with system clock drift:

- The Clock is anchored to the system time at start()
- If the system clock is not perfectly synced with NTP when anchoring, the base_ms will be offset
- This offset will persist for the lifetime of the Clock

**Mitigation**: Consider one of these approaches:
- Add a method to periodically resynchronize the clock with the system time
- Use SystemTime::now() directly for production instead of the Clock abstraction
- Accept this as a known limitation for testing purposes

### 📋 Suggested Improvements

1. **Add Clock Resynchronization**: Consider adding a method to resync the clock.

2. **Document Time Source Assumptions**: Add documentation explaining the expected precision of the clock and how it handles clock adjustments.

3. **Consider Using Instant::now() for Simple Cases**: If the block indexer doesn't need precise Unix timestamps, consider using tokio::time::Instant directly for simpler timing operations.

## Testing Analysis

The tests are well-structured and properly use Tokio's testing features. Consider adding tests for sleep duration accuracy and overflow handling.

## Decision: APPROVED WITH COMMENTS

The implementation is solid and addresses the stated requirements well. I recommend:

1. Address the sleep_until timing drift (medium priority)
2. Consider adding clock resynchronization for production robustness
3. Document the clock's limitations clearly for future maintainers

## Next Steps

1. Update the sleep_until implementation to use sleep_until with absolute Instant
2. Add a comment explaining the timing behavior
3. Consider adding a resync method for long-running processes

---

**Review Confidence**: HIGH
The code is well-written, properly tested, and addresses the requirements. The identified issues are minor improvements rather than critical bugs.

PR #456, opened 2026-06-11
Event Decoding Trait

nlordell:
This PR introduces an event decoding trait. This trait describes how logs can be decoded for a particular set of contract event types.

### Context and Motivation

Both validators and sentinels index different events. This means that the event indexer that we use needs to be generic on the events to index.

Fortunately, `alloy::sol!` gets us most of the way there:
- It generates event types matching Solidity interfaces
- It groups those events into enums and generates code for enumerating all selectors

There are a couple of small differences that our `Events` trait needs to reconcile:
- The `SELECTORS` associated constant is not part of any trait, meaning we can't use it in a generic way
- It does not have a way to group all events if we want to index multiple contracts.

### Decisions and Tradeoffs

We additionally introduce a macro to generate the `Events` implementation (since there is no use in doing it by hand as it can be trivially derived from the types created by the `sol!` macro).

### Testing

Added some tests to check the decoding, sorting, and macro logic on some test events.

---

## CLA signature

### DIFF
--- a/crates/core/src/index/events.rs
+++ b/crates/core/src/index/events.rs
@@ -1,0 +1,290 @@
+//! Event watcher.
+//!
+//! Reliably produces a stream of decoded, typed EVM logs. Consumers describe the
+//! events they care about with the [`watcher_events!`](crate::watcher_events)
+//! macro over `alloy` `sol!`-generated `*Events` enums, yielding a type that
+//! implements [`Events`] which the watcher decodes raw logs into.
+
+use alloy::{
+    primitives::{Address, B256},
+    providers::Provider,
+    rpc::types::Log,
+    transports::TransportError,
+};
+use std::marker::PhantomData;
+
+/// A typed set of EVM events that raw logs can be decoded into.
+///
+/// Implemented by the [`watcher_events!`](crate::watcher_events) macro over one
+/// or more `alloy` `sol!`-generated `*Events` enums.
+pub trait Events: Sized {
+    /// The event signature hashes (topic0) of every event in the set.
+    ///
+    /// These are used both to build the `eth_getLogs` query filter and for the
+    /// bloom-filter skip check, so they must cover exactly the events that
+    /// [`decode_log`](Events::decode_log) can produce.
+    fn topics() -> Vec<B256>;
+
+    /// Decodes a raw log, given its `topics` and `data`, into the event set;
+    /// returns `None` when the log is not one of the events in the set.
+    fn decode_log(topics: &[B256], data: &[u8]) -> Option<Self>;
+}
+
+/// Error produced by the event watcher.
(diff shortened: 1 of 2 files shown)


crates/core/src/index/events.rs
```diff
@@ -0,0 +1,290 @@
...
+
+/// Watches for the logs of the events `E` emitted by a set of addresses.
+pub struct EventWatcher<P, E> {
+    provider: P,
+    addresses: Vec<Address>,
+    topics: Vec<B256>,
+    _events: PhantomData<fn() -> E>,
+}
+
+impl<P, E> EventWatcher<P, E>
```
rmeissner, line 52: nit: feels like this is not related to the decoding trait

crates/core/src/index/events.rs
```diff
@@ -0,0 +1,290 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloy::{primitives::Address, sol, sol_types::SolEvent, uint};
+    use std::assert_matches;
+
+    sol! {
+        #[derive(Debug, Default, Eq, PartialEq)]
+        contract Erc20 {
```
rmeissner, line 157: nit: maybe test with abis that we use for Safenet

rmeissner (approved):

crates/core/src/index/events.rs
nlordell: I wanted to setup the generics of the decoding trait already in the `EventWatcher` to make sure that it works. In particular, there is a small detail here that I didn't comment on which is the `PhantomData` type (which is also intrinsically tied to the event watcher.

In particular, the `Events` trait only has methods for constructing instances of the event and we do not hold any "E" type in the event watcher itself. Without the phantom data, we would get a compiler error:

```
type parameter `E` is never used
consider removing `E`, referring to it in a field, or using a marker such as `PhantomData`
if you intended `E` to be a const parameter, use `const E: /* Type */` instead
```

Furthermore, we have a few options here:
- `PhantomData<E>`, the slightly more obvious choice
- `PhantomData<fn() -> E>`, the slightly less obvious choice, but that is slightly more explicit in the [variance](https://doc.rust-lang.org/reference/subtyping.html) with E. In particular, we want to specify that we are explicitly covariant with the type of E, and share no variances with the lifetime of `E`.

I wanted to keep all the generic magic into a single PR, but you are right that defining the type is not strictly needed here.

crates/core/src/index/events.rs
nlordell: I wanted to keep the tests simpler. We will have tests for Safenet ABIs in the validator and sentinel, and I don't want to test `sol!` macro's implementation. It would also make the tests more verbose (because of the larger types) which I was trying to avoid.

PR #458, opened 2026-06-12
Log Fetching Client-Side Error Detection

nlordell:
This ports the client-side error detection strategies from the TypeScript logs fetcher.

### Context and Motivation

We can detect node `eth_getLogs` bad behaviour in some cases and produce errors in the event watcher. Unfortunately, this happens often enough that it is worth porting this logic over to Rust in order to maintain a reliable event stream.

### Decisions and Tradeoffs

I reworked some of the errors to be more helpful. In particular, we error with the block and log index when we fail to decode (making it easier to pinpoint the log in a block explorer for example).

### Testing

Tested the new error conditions, the log fetching logic should now match the TypeScript implementation: https://github.com/safe-research/safenet/blob/86d5d0fada3a19d2c6a3f3b5b6ba585702a29206/validator/src/watcher/events.ts#L111-L218

---

## CLA signature

### DIFF
--- a/crates/core/src/index/events.rs
+++ b/crates/core/src/index/events.rs
@@ -5,6 +5,7 @@
 //! macro over `alloy` `sol!`-generated `*Events` enums, yielding a type that
 //! implements [`Events`] which the watcher decodes raw logs into.
 
+use super::bloom;
 use alloy::{
     primitives::{Address, B256, Bloom},
     providers::Provider,
@@ -168,7 +169,7 @@ where
 
                 // Verify the node served a complete set of logs for the block by
                 // recomputing the bloom filter over every returned log.
-                if compute_logs_bloom(&logs) != logs_bloom {
+                if bloom::compute_logs_bloom(&logs) != logs_bloom {
                     return Err(Error::IncompleteLogs { block_hash });
                 }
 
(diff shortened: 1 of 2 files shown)


crates/core/src/index/events.rs
```diff
@@ -30,15 +31,35 @@ pub trait Events: Sized {
...
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
     /// An RPC request failed.
     #[error(transparent)]
     Rpc(#[from] TransportError),
     /// An error decoding a log into an event.
-    #[error("failed to decode log with topic0 {0:?}")]
-    DecodeLog(Option<B256>),
+    #[error("failed to decode log index {log_index} on block {block_hash:?}")]
+    DecodeLog { block_hash: B256, log_index: u64 },
```
rmeissner, line 54: question: the topic is not useful anymore, because we can pull it from the block?

rmeissner (approved):

crates/core/src/index/events.rs
nlordell: The topic wasn't useful for debugging - it doesn't help us figure out _which_ log was fetched and had the issue decoding (in case there were multiple logs with that topic in the block).

PR #459, opened 2026-06-12
Add Event Warping Implementation

nlordell:
Add implementation for fetching events over large block ranges where reorgs are no longer possible.

### Decisions and Tradeoffs

Like with the TypeScript implementation, if we get to a page size of 1 and still fail, then we split the `getLogs` query into multiple queries. This allows us to support blocks where too many events happen for a node to return with a single getLogs query.

### Testing

Warping logic is verified. Matches TypeScript implementation: https://github.com/safe-research/safenet/blob/86d5d0fada3a19d2c6a3f3b5b6ba585702a29206/validator/src/watcher/events.ts#L220-L252

---

## CLA signature

### DIFF
--- a/crates/core/src/index/events.rs
+++ b/crates/core/src/index/events.rs
@@ -5,14 +5,18 @@
 //! macro over `alloy` `sol!`-generated `*Events` enums, yielding a type that
 //! implements [`Events`] which the watcher decodes raw logs into.
 
-use super::bloom;
+use super::{blocks::BlockUpdate, bloom};
 use alloy::{
     primitives::{Address, B256, Bloom},
     providers::Provider,
     rpc::types::{Filter, Log},
     transports::TransportError,
 };
-use std::{collections::BTreeSet, marker::PhantomData, num::NonZeroUsize};
+use std::{
+    collections::BTreeSet,
+    marker::PhantomData,
+    num::{NonZeroU64, NonZeroUsize},
+};
 
 /// A typed set of EVM events that raw logs can be decoded into.
 ///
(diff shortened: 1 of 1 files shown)


crates/core/src/index/events.rs
```diff
@@ -772,4 +901,123 @@ mod tests {
...
+                block_page_size: NonZeroU64::new(5).unwrap(),
+                ..Default::default()
+            },
+        );
+        events
+            .on_block_update(BlockUpdate::Warp { from: 1, to: 10 })
+            .unwrap();
+
+        // The first page queries blocks 1..=5, the second 6..=10.
+        asserter.push_success(&vec![log(
```
rmeissner, line 920: comment: In this example I find the tests not meaningful enough. I know we decided not to test internals but the test doesn't check that actually the correct pages are queries, it just checks that 2 pages where queried, if the implementation queries 2 times page 1, then the tests also succeeds.

rmeissner (approved):

crates/core/src/index/events.rs
nlordell: :100: agree, this bothered me as well (and in other tests too). Unfortunately, the mocking facilities in `alloy` only allow pushing responses and not asserting on requests.

I created an issue, as it will be a bit more work to get it working, and this is something that we can propose upstream in `alloy` itself.

https://github.com/safe-research/safenet/issues/465

nlordell: I still like this test as is (despite asserting on inputs being an obvious improvement), as it documents _how many_ requests we make (which we try very hard to reduce as much as possible in the indexing logic).

PR #462, opened 2026-06-12
Implement Watcher

nlordell:
This PR hooks up the block and event watchers to make the top level watcher used by the services.

### Decisions and Tradeoffs

Unlike in TypeScript, we don't need the "start" function, the background worker, or subscriptions. We can trivially turn the watcher into a `std::futures::Stream` of updates using `tokio` helper methods, that we can chain with the handler. This simplifies the code, and makes it easier to compose :tada:.

### Testing

Added tests to verify the block invalidation logic. The logic should match the TypeScript implementation: https://github.com/safe-research/safenet/blob/86d5d0fada3a19d2c6a3f3b5b6ba585702a29206/validator/src/watcher/index.ts#L60-L108

---

## CLA signature

### DIFF
--- a/crates/core/src/index/mod.rs
+++ b/crates/core/src/index/mod.rs
@@ -1,10 +1,356 @@
 //! Indexing of onchain blocks and events: following the chain head, fetching
 //! event logs in order, and handling chain reorgs.
 
-#[allow(dead_code)]
-mod blocks;
+pub mod blocks;
 #[allow(dead_code)]
 mod bloom;
 mod clock;
-#[allow(dead_code)]
-mod events;
+pub mod events;
+
+use alloy::{primitives::Address, providers::Provider, rpc::types::error::EthRpcErrorCode};
+use blocks::BlockWatcher;
+use events::{EventWatcher, Events};
+
+pub use blocks::BlockUpdate;
+
+/// Watcher configuration, aggregating the block and event watcher configs.
+#[derive(Clone, Debug)]
+pub struct Config {
+    /// Block watcher configuration.
+    pub blocks: blocks::Config,
+    /// Event watcher configuration.
+    pub events: events::Config,
+}
+
+/// Error produced by the [`Watcher`].
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    /// An error from the block watcher.
+    #[error(transparent)]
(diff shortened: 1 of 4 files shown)


crates/core/src/index/mod.rs
```diff
@@ -1,10 +1,356 @@
...
+    /// `addresses`, resuming from `last_indexed_block` (see
+    /// [`BlockWatcher::create`]).
+    pub async fn create(
+        provider: P,
+        config: Config,
+        addresses: Vec<Address>,
+        last_indexed_block: Option<u64>,
+    ) -> Result<Self, Error> {
+        let blocks =
+            BlockWatcher::create(provider.clone(), config.blocks, last_indexed_block).await?;
```
rmeissner: nit: for blocks we use `create` for events we use `new`, probably would make sense to use the same in both cases.

crates/core/src/index/mod.rs
```diff
@@ -1,10 +1,356 @@
...
+    ///
+    /// Returns the next batch of logs or blocks and waits for a new block to
+    /// be mined.
+    pub async fn next(&mut self) -> Result<Update<E>, Error> {
+        loop {
+            let update = match self.next_logs().await? {
+                // The query produced nothing for us; keep draining.
+                Some(logs) if logs.is_empty() => continue,
+                Some(logs) => Update::Logs(logs),
+                None => {
```
rmeissner, line 85: uber nit: personally I find it easier to understand if the cases would be ordered differently:
1. fetch new block (new None case)
2. handle new logs (the non empty logs case)
3. skip empty logs
as this is how the flow happens in my head :D

rmeissner (approved):

crates/core/src/index/mod.rs
nlordell: Ah yeah - nice catch.

I ported the `create` function directly from TypeScript and didn't think about the naming, but definitely prefer consistent `new` naming for component initialization (Rust typically uses `new` for creating values).

crates/core/src/index/mod.rs
nlordell: Continue at the bottom makes sense: its kind of "if all else fails" which happens at the end.

PR #466, opened 2026-06-22
Database Storage Layer

nlordell:
This PR implements a simple filesystem storage for recording state updates, in a way that allows for rollbacks in case of reorgs.

### Decisions and Tradeoffs

For absolute simplicity and trivial correctness, we implemented a system with snapshots, where we store the full state for each block. Older blocks can be pruned in order to prevent unbounded growth of the state database.

---

## CLA signature

### DIFF
--- a/crates/core/src/state/storage.rs
+++ b/crates/core/src/state/storage.rs
@@ -36,7 +36,7 @@ pub enum Error {
 /// Generic over a serializable state value `S`, stored as JSON.
 pub struct SnapshotStore<S> {
     pool: SqlitePool,
-    _state: PhantomData<fn() -> S>,
+    _state: PhantomData<S>,
 }
 
 impl<S> SnapshotStore<S>
@@ -214,6 +214,8 @@ mod tests {
         ));
         // The failed rollback did not delete the snapshots above the target.
         assert_eq!(store.current().await.unwrap(), Some((3, state(30))));
+        // Block 2 is still in the store.
+        assert_eq!(store.reorg(3).await.unwrap(), (2, state(20)));
     }
 
     #[tokio::test]
--- a/crates/core/src/state/storage.rs
+++ b/crates/core/src/state/storage.rs
@@ -67,12 +67,14 @@ where
     /// On startup this is the resume point: the block number is the indexer's
     /// last indexed block.
     pub async fn current(&self) -> Result<Option<(u64, S)>, Error> {
-        sqlx::query_as::<_, (u64, String)>(
+        sqlx::query_as::<_, (i64, String)>(
             "SELECT block_number, state FROM snapshots ORDER BY block_number DESC LIMIT 1",
         )
         .fetch_optional(&self.pool)
         .await?
-        .map(|(block_number, state)| Ok((block_number, serde_json::from_str(&state)?)))
+        .map(|(block_number, state)| {
+            Ok((u64::try_from(block_number)?, serde_json::from_str(&state)?))
+        })
         .transpose()
     }
 
@@ -84,7 +86,7 @@ where
             "INSERT INTO snapshots (block_number, state) VALUES (?, ?)
              ON CONFLICT (block_number) DO UPDATE SET state = excluded.state",
         )
-        .bind(block_number as i64)
+        .bind(i64::try_from(block_number)?)
         .bind(state)
         .execute(&self.pool)
         .await?;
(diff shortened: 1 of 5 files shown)


crates/core/src/state/storage.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    /// Rolls the store back, discarding `uncle` and every snapshot above it, and
+    /// returns the parent block number and its restored state.
+    ///
+    /// Used to recover from a reorg, where `uncle` was removed from the canonical
+    /// chain. Errors with [`Error::MissingSnapshot`] if there is no snapshot at
+    /// the parent of `uncle`, leaving the store unchanged.
+    pub async fn reorg(&self, uncle: u64) -> Result<(u64, S), Error> {
+        let parent = uncle.checked_sub(1).ok_or(Error::BlockNumberOverflow)?;
```
rmeissner, line 103: nit: isn't this an underflow? :P

crates/core/src/state/storage.rs
```diff
@@ -0,0 +1,235 @@
...
+    #[error("no snapshot found at block {0}")]
+    MissingSnapshot(u64),
+}
+
+/// A reorg-aware store of per-block state snapshots, backed by SQLite.
+///
+/// Generic over a serializable state value `S`, stored as JSON.
+pub struct SnapshotStore<S> {
+    pool: SqlitePool,
+    _state: PhantomData<fn() -> S>,
```
rmeissner: note: this one is a pattern I didn't know before and worth sharing in the tech team ;)

rmeissner: question: why is it `fn() -> S` and not just `S`?

crates/core/src/state/storage.rs
```diff
@@ -0,0 +1,235 @@
...
+            _state: PhantomData,
+        })
+    }
+
+    /// Returns the tip snapshot (the one with the highest block number) and its
+    /// block number, or `None` when the store is empty.
+    ///
+    /// On startup this is the resume point: the block number is the indexer's
+    /// last indexed block.
+    pub async fn current(&self) -> Result<Option<(u64, S)>, Error> {
```
rmeissner, line 69: thought/ nit: this requires that we always load (read and serialize) all state from the snapshots (i.e. the state of all signing requests). Not a blocker for this initial version, but maybe there is a way to handle this nicely.

crates/core/src/state/storage.rs
```diff
@@ -0,0 +1,235 @@
...
+        let store = store().await;
+        store.commit(2, &state(20)).await.unwrap();
+        store.commit(3, &state(30)).await.unwrap();
+
+        assert!(matches!(
+            store.reorg(2).await,
+            Err(Error::MissingSnapshot(1))
+        ));
+        // The failed rollback did not delete the snapshots above the target.
+        assert_eq!(store.current().await.unwrap(), Some((3, state(30))));
```
rmeissner, line 216: uber nit: from the test it is not clear if the state for block 2 is now still in the db or not. For this test I don't think having 2 block entries is necessary, right? So if only one is in store before reorg this can be tested.

rmeissner (approved):

crates/core/src/state/storage.rs
nlordell: I discuss this pattern here: https://github.com/safe-research/safenet/pull/456#discussion_r3442544365

> question: why is it `fn() -> S` and not just `S`?

Copy-pasta from the event decoder, it should be `S` here.

crates/core/src/state/storage.rs
nlordell: Correct. This is just the storage layer, the state management layer (future PR) will be responsible for calling this once on startup, but keeping an in-memory representation of the data to avoid reading from disk for every event.

crates/core/src/state/storage.rs
nlordell: Good point, I will make it clear (add an assertion showing that block 2 is still in there).

crates/core/src/state/storage.rs
nlordell: It is an underflow. I didn't want to differentiate between the two (too much ceremony for figuring out under/over-flow, especially with casts).

crates/core/src/state/storage.rs
rmeissner: it was not only about reading from disk, but also about having to keep the whole state in-memory. I would have preferred to only have the state in memory that is required for a specific state transition. 

That being said, this is probably a completely unnecessary optimization at this point, as our memory consumption on the typescript implementation is not a bottleneck (at least as far as I can tell)

nlordell: Ahhh, that would be a bit more challenging as it would also require the storage layer to know about different state slices.

I do agree it is a nice optimization, but then the storage becomes a bit more complicated (and closer to how I implemented the SQLite storage, where I modeled the completed stored data relationally).

rmeissner: Imo we should somewhere document why we think it is fair to load all state into memory. I assume that this is possible because we don't assume too many transaction checks running in parallel.

PR #467, opened 2026-06-22
Additional Information to Subscribers

nlordell:
Add some additional information to the indexer updates for the storage layer.

### Context and Motivation

In order to fully handle reorgs and pruning at the storage layer, we need some additional context:

- The safe block that will never reorg, to decide what can be pruned
- The event block ranges, so we know which blocks to commit states to and whether or not events can be skipped in case of a failed reorg

### Decisions and Tradeoffs

This PR does just that in an unintrucive way:

- The block listener has a method for retrieving the current safe block
- Event updates now include block ranges

---

## CLA signature

### DIFF
--- a/crates/core/src/index/blocks.rs
+++ b/crates/core/src/index/blocks.rs
@@ -76,6 +76,9 @@ pub enum BlockUpdate {
         number: u64,
         hash: B256,
         logs_bloom: Bloom,
+        /// The highest block that can no longer reorg. State at or below it is
+        /// final, so older snapshots may be pruned.
+        safe: u64,
     },
 }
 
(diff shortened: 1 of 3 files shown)


nlordell:
~~Moved to draft for now, I think we also need block numbers on the events themselves.~~

Edit: Not needed, this can be expressed with the `Events` trait (since the full log is given there).

crates/core/src/index/blocks.rs
```diff
@@ -259,6 +259,16 @@ where
         Ok(())
     }
 
+    /// Gets the highest block that can no longer be reorged. State at or below
+    /// it is considered final by the block indexer.
+    pub fn safe_block(&self) -> u64 {
+        self.recent
```
rmeissner: question: why make this a read function instead of adding it to the event info (i.e. add it as a field in BlockUpdate)?

rmeissner (approved):

crates/core/src/index/blocks.rs
nlordell: It is a nice suggestion. I had tried it out, but it annoyed me how I was passing the safe block to the event watcher.

In retrospect, having the safe block there makes a lot of sense.

nlordell: Implemented in 512b2818735a24f3b2419bf9d81d81db585f2de0

PR #469, opened 2026-06-23
Add State Machine Implementation

nlordell:
Implement a state machine helper that handles indexer updates and applies state transitions for the updates.

### Context and Motivation

In order to not duplicate storage snapshot handling code (as well as some basic indexer update muxing) in each service, implement a `StateMachine` type that does a lot of the heavy lifting for you.

### Decisions and Tradeoffs

In order to ensure that state transitions are sound, we always move the state into the state transition implementation, and expect it to return an updated value. The proposed interface ensures a couple of things:

- We are efficient with memory and don't duplicate the whole state, we can apply incremental changes to it. I had researched some persistent data structures (like [`rpds`](https://docs.rs/rpds/latest/rpds/)) that would allow for very cheap clones, but decided against it. Also, conceptually, the state machine should never error out and with the current interface is forced to gracefully handle all errors.
- The `transitions` is a mutable, allowing it to perform side-effects outside of the state transition. In particular this is required for nonce usage (nonces need to be burned on usage).

This choice has some implications on how we store the current state in the state machine struct. In particular, we need it to be an `Option` (so we can move the value out of the statemachine and into the state transition for updating). Additionally, we wrap in a mutex to ensure no concurrent updates (which is required by the semantics of the state machine). We intend to use this in a single fiber anyway, so this should not have a negative impact on its usability.

Finally, we implement the state machine in such a way that it cannot recover from errors, and becomes poisoned on failed state transitions, out of order updates, or database failures. This is because this potentially puts us in a bad state with no clear way to recover - in which case we should fail loud.

---

## CLA signature

### DIFF
--- a/crates/core/src/state/mod.rs
+++ b/crates/core/src/state/mod.rs
@@ -9,7 +9,7 @@ pub mod storage;
 use self::storage::SnapshotStore;
 use crate::index::{BlockUpdate, EventUpdate, Update};
 use serde::{Serialize, de::DeserializeOwned};
-use std::mem;
+use std::{mem, range::RangeInclusive};
 use tokio::sync::Mutex;
 
 /// Error produced by the [`StateMachine`].
@@ -18,6 +18,13 @@ pub enum Error {
     /// A snapshot storage error.
     #[error(transparent)]
     Storage(#[from] storage::Error),
+    /// We have reached the end of the block chain and cannot continue handling
+    /// updates.
+    #[error("end of chain")]
+    EndOfChain,
+    /// Received an update that is either out-of-order or has unexpected data.
+    #[error("out-of-order update")]
+    BadUpdate,
     /// The state machine is in a poisoned state, where a previous state
     /// transition failed in a non-recoverable way.
     #[error("poisoned state machine")]
@@ -54,9 +61,15 @@ pub struct StateMachine<S, T> {
 }
 
 struct Inner<S> {
-    block: u64,
     state: S,
-    warping: bool,
+    status: Status,
+}
+
+enum Status {
+    Initialized,
+    BlockPending { pending: u64 },
+    BlockEvents { latest: u64 },
+    WarpEvents { range: RangeInclusive<u64> },
 }
 
 impl<S, T, E> StateMachine<S, T>
@@ -79,12 +92,16 @@ where
         snapshots: SnapshotStore<S>,
         init: impl FnOnce() -> S,
     ) -> Result<Self, Error> {
-        let (block, state) = snapshots.current().await?.unwrap_or_else(|| (0, init()));
-        let inner = Mutex::new(Some(Inner {
-            block,
-            state,
-            warping: false,
-        }));
+        let (state, status) = snapshots
+            .current()
+            .await?
+            .map(|(latest, state)| -> Result<_, Error> {
+                let pending = latest.checked_add(1).ok_or(Error::EndOfChain)?;
+                Ok((state, Status::BlockPending { pending }))
+            })
+            .transpose()?
+            .unwrap_or_else(|| (init(), Status::Initialized));
+        let inner = Mutex::new(Some(Inner { state, status }));
 
         Ok(Self {
             inner,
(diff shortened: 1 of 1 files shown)


crates/core/src/state/mod.rs
```diff
@@ -5,3 +5,433 @@
...
+                if inner.warping {
+                    self.snapshots.prune(blocks.last).await?;
+                }
+
+                actions
+            }
+            _ => return Err(Error::OutOfOrderUpdate),
+        };
+        *lock = Some(inner);
+        Ok(actions)
```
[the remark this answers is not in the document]
nlordell: This in intentional.

crates/core/src/state/mod.rs
```diff
@@ -5,3 +5,433 @@
...
+/// case of unexpected events, the state transition must gracefully recover.
+pub trait StateTransition<S>
+where
+    S: Sized,
+{
+    type Event;
+    type Action;
+
+    /// Perform the state transition for when a new block is observed.
+    fn new_block(&mut self, state: S, block: u64) -> impl Future<Output = (S, Vec<Self::Action>)>;
```
rmeissner, line 46: why a future? In typescript we moved away from async functions for state transitions.

crates/core/src/state/mod.rs
```diff
@@ -5,3 +5,433 @@
...
+        &mut self,
+        state: S,
+        event: Self::Event,
+    ) -> impl Future<Output = (S, Vec<Self::Action>)>;
+}
+
+/// A service state machine.
+pub struct StateMachine<S, T> {
+    inner: Mutex<Option<Inner<S>>>,
+    transition: T,
```
rmeissner, line 59: question: on the implementation we enforce T to be a StateTransition. For me it would feel more intuitive to do this here already. Is this a Rust pattern to limit the state type a late as possible?

crates/core/src/state/mod.rs
```diff
@@ -5,3 +5,433 @@
...
+
+    /// Creates a new state machine with the given state transition and an
+    /// initial value constructor.
+    pub async fn with_init(
+        transition: T,
+        snapshots: SnapshotStore<S>,
+        init: impl FnOnce() -> S,
+    ) -> Result<Self, Error> {
+        let (block, state) = snapshots.current().await?.unwrap_or_else(|| (0, init()));
+        let inner = Mutex::new(Some(Inner {
```
rmeissner: nit: for me a more descriptive name would be "current" as this is the current state.

crates/core/src/state/mod.rs
```diff
@@ -5,3 +5,433 @@
...
+    /// Handle an indexer update.
+    ///
+    /// The state machine halts if it returns an error, as it can no longer
+    /// correctly progress.
+    pub async fn handle_update(&mut self, update: Update<E>) -> Result<Vec<T::Action>, Error> {
+        let mut lock = self.inner.lock().await;
+        let mut inner = mem::take(&mut *lock).ok_or(Error::Poisoned)?;
+        let actions = match update {
+            Update::Block(BlockUpdate::Warp { from, to })
+                if Some(from) == inner.block.checked_add(1) && to >= from =>
```
rmeissner: `to >= from` - is there any way this is not true?

Also I think it would be nice to comment that we expect no "holes" in the block numbers (so we ensure that we have "seen" every block).

crates/core/src/state/mod.rs
```diff
@@ -5,3 +5,433 @@
...
+                    warping: false,
+                };
+                actions
+            }
+            Update::Logs(EventUpdate { blocks, logs })
+                if !blocks.is_empty()
+                    && (inner.is_latest_block(blocks) || inner.is_historic_range(blocks)) =>
+            {
+                let mut actions = vec![];
+                for log in logs {
```
rmeissner, line 153: question/ nit: we do not enforce log ordering here, right? we assume them to come in order. This is maybe an uber nit, but it feels weird that we enforce block ordering, but not log ordering.

crates/core/src/state/mod.rs
```diff
@@ -5,3 +5,433 @@
...
+                    let (new_state, new_actions) = self.transition.event(inner.state, log).await;
+                    inner.state = new_state;
+                    actions.extend(new_actions);
+                }
+                self.snapshots.commit(blocks.last, &inner.state).await?;
```
rmeissner: question: why only commit state on logs and not on a block? In case there is a block without events we still want to commit the state, right?

crates/core/src/state/mod.rs
nlordell: Great question! I hinted at this in the PR description:

> The transitions is a mutable, allowing it to perform side-effects outside of the state transition. In particular this is required for nonce usage (nonces need to be burned on usage).

The big difference is that SQLite is `async` for us, but not in TypeScript.

crates/core/src/state/mod.rs
nlordell: I was being overly defensive here, we do not expect this ever.

crates/core/src/state/mod.rs
```diff
@@ -5,3 +5,433 @@
...
+                // In case we are warping, we can prune intermediate state from
+                // storage for the event pages.
+                if inner.warping {
+                    self.snapshots.prune(blocks.last).await?;
```
rmeissner, line 174: question: do we expect pruning to happen only in this case? Shouldn't we prune on each new block to the latest safe block?

rmeissner (dismissed):

crates/core/src/state/mod.rs
nlordell: :facepalm: We definitely should, it was a mistake when rebasing from https://github.com/safe-research/safenet/pull/467#discussion_r3453113220. Great catch!

crates/core/src/state/mod.rs
rmeissner: Kind of missed this, but actually make me a little sad 😢 because I was hoping for side effect free functions. An alternative would be to have nonce burning as an "action". note: not a blocker, rather a uber nit

crates/core/src/state/mod.rs
nlordell: We emit `Update::Events { blocks: n..=n, logs: [] }` in that case, which will trigger the commit, introduced in #467 

https://github.com/safe-research/safenet/blob/d08d6934558e8b249e7fe34005d8652c37dfdc76/crates/core/src/index/mod.rs#L87-L89

crates/core/src/state/mod.rs
nlordell: We expect `Update` ordering, not block ordering; it is just that updates tend to be block-centric.

Definitely worth discussing (we can talk shortly about it tomorrow).

crates/core/src/state/mod.rs
nlordell: `inner` is just a common Rust idiom when wrapping some internal state that is not part of the public type. Will rename :+1:

rmeissner (approved):
besides the small prune improvement, no blocker only questions

crates/core/src/state/mod.rs
nlordell: It is, we try to restrict traits where they are needed (typically, this is on the methods, because when we implement the method then you need additional typing information).

One exception is with associated types, where you sometimes need to restrict a type in the struct declaration.

crates/core/src/state/mod.rs
rmeissner: ahhh ok, if this is common in Rust you can also keep it ;)

crates/core/src/state/mod.rs
nlordell: I remember we discussed this in the past. My concern was that it might make the state machine harder to reason about (you have more "waiting" states).

I am not against the idea... I will think about it and maybe we come up with a cleaver way to represent this (basically, we want to find a way to model [algebraic effects](https://overreacted.io/algebraic-effects-for-the-rest-of-us/) in Rust, which AFAIK is not super straight forward at the moment).

rmeissner: don't spent too much time, as said it is an uber nit ;) (premature optimization)

crates/core/src/state/mod.rs
nlordell: I ended up making stricter update ordering checks in the update handler, just to make sure we catch any issues with the indexing logic.

crates/core/src/state/mod.rs
nlordell: Added the pruning.

PR #471, opened 2026-06-23
Optimize Block State Transition

nlordell:
Process new block state transitions earlier, allowing us to include actions that would have been triggered on the pending block to be included in the pending block itself. **This is an optimisation and not a required change**.

### Context and Motivation

The validator processes timeouts when new blocks are observed. Some timeouts may actually cause actions, some examples:

- In case some validators timed out their nonce reveal or signature sharing for signing, the last participant is required to restart a signing process to that they can recover with a smaller selection
- For sentinel voting, once the voting period ends validators can immediately start attesting

With that in mind, once we're in a state where we expect that the next update will be a `BlockUpdate::New`, then we can optimistically apply the state transition. This will allow us to know the actions that the validator wants to do after the start of the next pending block, and submit them to the mempool one block earlier, such that the pending block actions are already included in the pending block (instead of having to wait for the pending block to mine and only have the action be submitted the following block).

This saves us a whole block of waiting for all actions related to new block transitions.

### Decisions and Tradeoffs

This is related to comments in #469, but we are quite a bit strict on the ordering of block numbers from updates. We remain this way in this optimisation as this ordering is important for soundness of the state with the optimistic transition application.

---

## CLA signature

### DIFF
--- a/crates/core/src/state/mod.rs
+++ b/crates/core/src/state/mod.rs
@@ -135,7 +135,7 @@ where
                         .snapshots
                         .current()
                         .await?
-                        .expect("applied block state implies a storage snapshot");
+                        .ok_or(storage::Error::MissingSnapshot(pending.saturating_sub(1)))?;
                     debug_assert_eq!(block + 1, pending);
                     inner.state = state;
                 }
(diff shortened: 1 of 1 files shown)


nlordell:
Converting to draft, I'm still not entirely sure we need this optimization quite yet. We can introduce it later if needed.

PR #472, opened 2026-06-24
Validator 7702 Account

remedcu:
Add a minimal, stateless EIP-7702 account that lets the validator EOA batch multiple calls into a single transaction.

### Context and Motivation

The validator currently operates as a plain EOA, so it can only submit one call per transaction (one propose/attest at a time). By delegating the validator EOA to this account via EIP-7702, the validator can batch multiple calls — typically across the Consensus and FROST coordinator contracts — into a single transaction.

This PR intentionally contains **only the contract**, to get an early review of the design and security model from the team. The Foundry tests and the deployment script will follow in a separate PR once the contract is agreed upon.

Authorization relies entirely on EIP-7702: when the delegating EOA sends a transaction to its own address, `msg.sender == address(this)`, which `execute` requires. Replay protection comes from the EOA's transaction nonce, so no additional signature or nonce handling is implemented.

### Decisions and Tradeoffs

- **Per-call target, no hardcoding.** Each `Call` carries its own `to`, so a batch can route to different contracts (Consensus, coordinator, …) rather than a single fixed address. The contract holds no state — no immutable, no constructor.
- **Per-call gas limit.** Each call forwards at most `gasLimit` gas (`.call{gas: ...}`). This bounds each call so a single runaway/expensive call cannot consume the gas needed by the rest of the batch. No per-call `value` field — the targeted functions are non-payable; easy to add later if needed.
- **Best-effort, not atomic.** A failing call emits `CallFailed(index, result)` and the batch continues, rather than reverting the whole transaction. Failures are observable via `CallFailed` logs in the receipt; successful calls via the target contracts' own events.
- **Deliberately minimal, not ERC-4337.** No entry point, no signature validation, no fallback — only `execute` (self-only) and a payable `receive` so the account can be funded with gas.

### DIFF
--- a/contracts/src/Validator7702Account.sol
+++ b/contracts/src/Validator7702Account.sol
@@ -3,12 +3,13 @@ pragma solidity ^0.8.30;
 
 /**
  * @title Validator 7702 Account
- * @notice A minimal EIP-7702 account implementation that lets the validator EOA batch multiple calls to the
- *         Safenet {Consensus} contract into a single transaction.
+ * @notice A minimal EIP-7702 account implementation that lets the validator EOA batch multiple calls into a
+ *         single transaction.
  * @dev This contract is intended to be set as the EIP-7702 delegation target of the validator's EOA. Once the
  *      EOA has delegated to this implementation, the validator can submit a single transaction to its own
- *      address that invokes {execute}, proposing (or attesting to) multiple Safe transactions at once instead
- *      of being limited to a single consensus call per transaction as a plain EOA.
+ *      address that invokes {execute}, performing multiple calls (for example to the Safenet Consensus and
+ *      FROST coordinator contracts) at once instead of being limited to one call per transaction as a plain
+ *      EOA.
  *
  *      Authorization is provided entirely by EIP-7702. When the validator EOA initiates a transaction to its
  *      own address, the EVM sets `msg.sender == address(this)`, which {execute} requires. Producing such a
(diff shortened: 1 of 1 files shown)


nlordell (changes requested):
LGTM, we cannot fix the consensus address since we send to both coordinator and consensus contract.

Additionally, for non-atomic transactions I would suggest individual gas limits per call to make debugging gas parameter issues easier.

Love the event for failed calls, and just requesting changes on the fixed `caller` (the gas parameter is a nit).

contracts/src/Validator7702Account.sol
```diff
@@ -0,0 +1,95 @@
...
+ *      EOA.
+ *
+ *      Authorization is provided entirely by EIP-7702. When the validator EOA initiates a transaction to its
+ *      own address, the EVM sets `msg.sender == address(this)`, which {execute} requires. Producing such a
+ *      call requires the EOA's private key, and replay protection is provided by the EOA's transaction nonce,
+ *      so no additional signature or nonce handling is implemented here.
+ *
+ *      The account is intentionally minimal and is NOT ERC-4337 compatible: it has no entry point, no
+ *      signature validation, and no functionality beyond batching calls and receiving the native token used
+ *      to fund gas.
```
nlordell, line 26: Can you add a prominent warning that this is not suitable as a general purpose account?

contracts/src/Validator7702Account.sol
```diff
@@ -0,0 +1,100 @@
...
+    // CONSTRUCTOR
+    // ============================================================
+
+    /**
+     * @notice Constructs the account implementation.
+     * @param consensus The address of the Safenet Consensus contract that this account may call.
+     */
+    constructor(address consensus) {
+        CONSENSUS = consensus;
+    }
```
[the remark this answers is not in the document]
remedcu: Not required anymore.

contracts/src/Validator7702Account.sol
```diff
@@ -0,0 +1,95 @@
...
+     *      execution continues with the remaining calls, so one failing call does not prevent the others from
+     *      running. Bounding each call's gas also prevents a single call from consuming the gas needed by the
+     *      rest of the batch. The only condition that reverts the whole transaction is the self-call guard (see
+     *      {OnlySelf}).
+     * @param calls The calls to execute, in order.
+     */
+    function execute(Call[] calldata calls) external {
+        require(msg.sender == address(this), OnlySelf());
+        for (uint256 i = 0; i < calls.length; ++i) {
+            (bool success, bytes memory result) = calls[i].to.call{gas: calls[i].gasLimit}(calls[i].data);
```
nlordell, line 85: comment: Only interesting thing that I can see with Solidity would be that it actually allocates for `calls[i].data` here instead of just writing temporary memory past the free memory pointer.

**NOT FOR THIS PR**, but we should probably investigate (I think this can be checked by reading the free memory pointer and seeing if it grows more than what is needed for the results bytes).

nlordell (approved):

rmeissner (dismissed):
I am note sure where this code is planned to be used and therefore I am also not sure if this should be part of our code base right now

contracts/src/Validator7702Account.sol
remedcu: Updated 👍🏾

PR #473, opened 2026-06-24
Implement Signer Module

nlordell:
### Context and Motivation

We need to be able to sign transactions for submitting actions on-chain. This PR adds a simple abstraction that wraps Alloy types to do just that.

### Decisions and Tradeoffs

We wrapped the types for a couple of reasons:
- Alloy interfaces are complicated, and there are many of them (it usually trips up coding agents). With this, we have a simple abstraction that the agent can leverage
- encapsulates signing code into a module, so calls to transaction signing and getting the transaction hash are easier.

---

## CLA signature


rmeissner (approved):

crates/core/src/tx/account.rs
```diff
@@ -0,0 +1,77 @@
+//! The local account used to sign transactions for submitting onchain.
+
+use alloy::{
+    consensus::{SignableTransaction as _, TxEip1559},
+    eips::Encodable2718 as _,
+    network::TxSignerSync as _,
+    primitives::{Address, B256, keccak256},
+    signers::k256,
+    signers::local::PrivateKeySigner,
```
remedcu: nit:
```suggestion
    signers::{k256, local::PrivateKeySigner},
```

crates/core/src/tx/account.rs
```diff
@@ -0,0 +1,77 @@
...
+};
+
+/// An error ECDSA signing a transaction.
```
remedcu, line 15: nit: 
```suggestion
/// An error from ECDSA signing a transaction.
```

crates/core/src/tx/account.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Debug, thiserror::Error)]
+#[error("an error occurred signing an Ethereum transaction")]
+pub struct SigningError;
```
remedcu: Question: Why here only the `SigningError` was considered and not unlike other files a custom `enum Error` which can include both signing as well as the `k256::ecdsa::Error` type?

crates/core/src/tx/account.rs
```diff
@@ -0,0 +1,77 @@
...
+            .sign_transaction_sync(&mut tx)
+            .map_err(|_| SigningError)?;
+        let raw_tx = tx.into_signed(signature).encoded_2718();
+        Ok(SignedTransaction(raw_tx))
+    }
+}
+
+impl SignedTransaction {
+    /// Compute the hash of the signed transaction.
+    pub fn hash(&self) -> B256 {
```
remedcu:
```suggestion
    pub fn hash(&self) -> TxHash {
```

Can we use [TxHash](https://docs.rs/alloy-core/latest/alloy_core/primitives/aliases/type.TxHash.html) here?

crates/core/src/tx/account.rs
```diff
@@ -0,0 +1,77 @@
...
+pub struct Account {
+    signer: PrivateKeySigner,
+}
+
+/// A raw signed transaction.
+pub struct SignedTransaction(Vec<u8>);
+
+impl Account {
+    /// Creates an account for the given `private_key`.
+    pub fn new(private_key: B256) -> Result<Self, k256::ecdsa::Error> {
```
remedcu:
```suggestion
    pub fn new(private_key: Zeroizing<B256>) -> Result<Self, k256::ecdsa::Error> {
```

Question: Should this be zeroed out from local memory?

remedcu (approved):
Not a fan of the PR heading, as the name is a bit misleading when considering AA from Ethereum 😅 

Rest mainly questions and nits.

crates/core/src/tx/account.rs
nlordell: Its an alias - it means they are the same type to the compiler. We can, and should, use the alias (but it doesn't really provide any additional guarantees).

crates/core/src/tx/account.rs
nlordell: Great question! This is because of error typing:

- At account creation, we get a very specific ECDSA error type.
- For signing, this goes over a generic trait (like an interface) which returns a very broad error type with things that do not apply to us (it even includes RPC errors as a variant). This is because the signing trait is generic over all kinds of signers (including remote ones like AWS KMS, or signers on an RPC node itself). The signing error types narrows the error to an ECDSA error which is the only thing that can happen in our case.

crates/core/src/tx/account.rs
nlordell: Nice suggestion!

crates/core/src/tx/account.rs
nlordell: Fixed.

crates/core/src/tx/account.rs
nlordell: We now use zeroizing types (in particular, `SigningKey` zeroizes on Drop). Deserialization zeroizes the temporary raw private key.

PR #474, opened 2026-06-24
Epoch Rollover Test

remedcu:
Add the Foundry unit test suite for the `EpochRollover` library merged in #468.

### Context and Motivation

[#468](https://github.com/safe-research/safenet/pull/468) merged the `EpochRollover` library but deferred its tests to a follow-up. This PR adds that suite, bringing the library under coverage before any guard adopts it, so its storage and verification behavior is pinned.

### Decisions and Tradeoffs

- **Real signing, no mocks or precomputed vectors.** Signatures are produced in-test with the `ForgeSecp256k1` helper plus a single-signer Schnorr `_sign`, so the tests exercise the actual `FROST.verify` / `Secp256k1` math rather than stubbing it. A single key suffices because `FROST.verify` only checks the group-key Schnorr equation. This follows the existing pattern in `SafenetGuardA.t.sol`.
- **Thin external wrappers (`callRollover`, `callInitialize`) for revert tests only.** The library functions are `internal`, so a direct call inlines into the test frame and reverts at the cheatcode's own depth, which `vm.expectRevert` rejects. The wrappers give those assertions a sub-call to catch; non-revert tests call the library directly.

### Testing

The change is the test suite itself: `forge test --match-path test/libraries/EpochRollover.t.sol` → **13 passing**; `forge fmt --check` and `forge lint` clean. Coverage:
- `initialize` — records the genesis pair and emits `EpochInitialized`; reverts on a zero key.
- `rollover` — happy path records the child and emits `EpochRolledOver`; chains past genesis (a recorded key serves as a parent); accepts branches (two keys at one epoch, and the same key at two epochs); and reverts on each guard: not-advancing (`==` and `<`), unknown parent, a known key at the wrong epoch, a zero new key, and a signature over a mismatched message. Re-submitting an identical rollover is a no-op (no event, state unchanged).
- `isKnown` — membership is exact on the `(key, epoch)` pair.

### DIFF
--- a/contracts/test/libraries/EpochRollover.t.sol
+++ b/contracts/test/libraries/EpochRollover.t.sol
@@ -1,0 +1,234 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+import {Test} from "@forge-std/Test.sol";
+import {Vm} from "@forge-std/Vm.sol";
+import {ConsensusMessages} from "@/libraries/ConsensusMessages.sol";
+import {EpochRollover} from "@/libraries/EpochRollover.sol";
+import {FROST} from "@/libraries/FROST.sol";
+import {Secp256k1} from "@/libraries/Secp256k1.sol";
+import {ForgeSecp256k1} from "@test/util/ForgeSecp256k1.sol";
+
+contract EpochRolloverTest is Test {
+    using EpochRollover for EpochRollover.T;
+    using ForgeSecp256k1 for ForgeSecp256k1.P;
+
+    EpochRollover.T internal state;
+
+    uint256 public constant CONSENSUS_CHAIN_ID = 100;
+    address public constant CONSENSUS_ADDR = address(0xC01115E1115115);
+
+    // Genesis group: secret key 1 → public key G, active at epoch 1.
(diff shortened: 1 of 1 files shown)


contracts/test/libraries/EpochRollover.t.sol
```diff
@@ -0,0 +1,235 @@
...
+        // Two valid rollovers from the same parent, distinct keys, SAME proposed epoch.
+        _doRollover(GENESIS_SK, GENESIS_NK, GENESIS_EPOCH, 2, 100, 5);
+        _doRollover(GENESIS_SK, 3, GENESIS_EPOCH, 2, 100, 6);
+
+        assertTrue(state.isKnown(_key(5), 2));
+        assertTrue(state.isKnown(_key(6), 2));
+    }
+
+    function test_rollover_sameKeyTwoEpochs() public {
+        _seedGenesis();
```
nlordell, line 151: nit: might be worth adding a note that this is _never_ expected to happen and included for completeness only.

nlordell (approved):

contracts/test/libraries/EpochRollover.t.sol
remedcu: Done 👍🏾

PR #480, opened 2026-06-25
[E 2/6]: Add Fee Helpers

nlordell:

### Context and Motivation

This PR contains some fee helpers used when building transactions for signing. It implements priority fee capping and fee bumping (to ensure that nodes do not reject fresh transactions).

### Decisions and Tradeoffs

We made an effort (especially in fee bumping) to ensure that there are no weird overflow edge-cases. This is mostly for defensive programming and should not actually occur in real life.

---

## CLA signature

### DIFF
--- a/crates/core/src/tx/fees.rs
+++ b/crates/core/src/tx/fees.rs
@@ -1,6 +1,6 @@
 //! EIP-1559 fee calculations for reliable transaction submission.
 
-use alloy::{eips::eip1559::Eip1559Estimation, primitives::U256, uint};
+use alloy::eips::eip1559::Eip1559Estimation;
 
 /// Caps the priority fee of `fees` so it is at most `cap_percentage` percent of
 /// the total max fee per gas, leaving the base-fee component unchanged.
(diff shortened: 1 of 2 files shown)


rmeissner (approved):

crates/core/src/tx/fees.rs
```diff
@@ -0,0 +1,94 @@
...
+        Eip1559Estimation {
+            max_fee_per_gas,
+            max_priority_fee_per_gas,
+        }
+    }
+
+    #[test]
+    fn caps_the_priority_fee_at_a_percentage_of_the_max_fee() {
+        // A priority fee already below the cap is unchanged.
+        assert_eq!(cap_priority_fee(fees(100, 10), 50.0), fees(100, 10));
```
rmeissner, line 72: nit: test for 0 or negative should be added

rmeissner (approved):

crates/core/src/tx/fees.rs
nlordell: Added.

PR #481, opened 2026-06-25
[E 3/6]: Add Specialized Transaction Types

nlordell:
This is part of a larger stack, see https://github.com/safe-research/safenet/pull/479.

### Context and Motivation

Alloy transaction request types have a lot of `Option`s. This is because it is designed for use with an account, which would make RPC calls to fill in certain values that are missing (such as nonce, gas fees, etc.).

In our case, we don't want this - we want _very_ strict control over what fields are required as input, and what fields are not. To model this, we introduce a specialized `Transaction` type, along with an `*WithNonce` variant that can be built into an EIP-1559 transaction for signing with the `Signer`.

---

## CLA signature

### DIFF
--- a/crates/core/src/tx/types.rs
+++ b/crates/core/src/tx/types.rs
@@ -1,0 +1,87 @@
+//! Transaction types for the queue.
+
+use crate::tx::fees;
+use alloy::{
+    consensus::TxEip1559,
+    eips::eip1559::Eip1559Estimation,
+    primitives::{Address, Bytes, TxKind, U256},
+    rpc::types::AccessList,
+};
+use serde::{Deserialize, Serialize};
+
+/// A transaction to submit onchain.
+///
+/// Analogous to alloy's [`TransactionRequest`], carrying the fields the queue
+/// requires to build an EIP-1559 transaction.
+///
+/// [`TransactionRequest`]: alloy::rpc::types::TransactionRequest
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Transaction {
+    /// The destination of the transaction.
+    pub to: Address,
+    /// The transaction value.
+    pub value: U256,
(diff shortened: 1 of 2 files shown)


crates/core/src/tx/types.rs
```diff
@@ -0,0 +1,87 @@
...
+/// [`TransactionRequest`]: alloy::rpc::types::TransactionRequest
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct Transaction {
+    /// The destination of the transaction.
+    pub to: Address,
+    /// The transaction value.
+    pub value: U256,
+    /// The transaction calldata.
+    pub input: Bytes,
```
rmeissner: nit: why not `data`?

crates/core/src/tx/types.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+}
+
+/// A [`Transaction`] with a nonce that is ready for submission.
+///
+/// It may contain fees from a previous submission.
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TransactionWithNonce {
```
rmeissner: uber nit: For me the `Transaction` struct is more the `TransactionData` (or content) and this `TransactionWithNonce` is the `Transaction`. 

Another alternative would be to call this the envelope. 

Why I am mentioning this nit, is because this is not just a transaction with nonce, but also with gas params ;)  

Edit: after reading the impl I am not sure about the my proposed naming, but neither with the current naming ... too many "transaction" types.

rmeissner (approved):

crates/core/src/tx/types.rs
nlordell: As discussed in person

- Transaction stays the same
- TransactinoWithNonce becomes AllocatedTransaction

PR #482, opened 2026-06-25
[E 4/6]: Transaction Storage Implementation

nlordell:

### Context and Motivation

Transactions need to be persisted in storage in order to survive restarts.

### Decisions and Tradeoffs

The storage implementation in the Rust port is different from the TypeScript one. In particular, it assigns stable "IDs" to transactions that is different from their nonce. This allows us a few things:

- We can store transactions without their nonce in the database, and remove them if the action were to expire before a signed transaction is broadcast. In particular, this can happen with initial transactions from actions for event state transitions while warping (signing ceremonies from a long gone past). This optimization allows us to skip these transactions and save on gas money
- We support reorgs by keeping execution blocks around until a safe block is reached. This allows us to handle some edge cases where a reorg causes a transaction not to be included, there is a gas increase preventing it from being included in future blocks, and gives us a chance to bump the transaction fee (there are some other edge cases around reorgs that are also solved)

This PR contains all the methods that are needed by the transaction queue - refer to the reference PR to see how it is used.

---

## CLA signature

### DIFF
--- a/crates/core/src/tx/storage.rs
+++ b/crates/core/src/tx/storage.rs
@@ -4,7 +4,7 @@
 //! serialized [`Transaction`] alongside the bookkeeping the queue needs: when it
 //! expires, its allocated nonce, when it was submitted and executed.
 
-use crate::tx::types::{Transaction, TransactionWithNonce};
+use super::types::{AllocatedTransaction, Transaction};
 use alloy::eips::eip1559::Eip1559Estimation;
 use sqlx::sqlite::SqlitePool;
 use std::num::TryFromIntError;
@@ -58,6 +58,13 @@ impl TransactionStorage {
     /// Creates a store backed by `pool`, creating the transactions table if it
     /// does not already exist.
     pub async fn new(pool: SqlitePool) -> Result<Self, Error> {
+        // Note that we store the `nonce` in a separate column from the
+        // transaction request JSON data. This allows us to work more naturally
+        // with the `nonce` column (for things like `MAX` to determine the next
+        // nonce), which would be more verbose if it were part of the request
+        // data directly (as we would need JSON extractors to use the column
+        // and would have to potentially deal with hexadecimal encoding, to
+        // match other numerical values are serialized).
         sqlx::query(
             "CREATE TABLE IF NOT EXISTS transactions (
                  id           INTEGER PRIMARY KEY,
@@ -101,6 +108,11 @@ impl TransactionStorage {
     /// Selects the oldest queued transaction that has not expired and assigns
     /// it a nonce. Returns `None` when nothing is queued.
     ///
+    /// The `status.block` is used as the current latest block number in order
+    /// to determine whether or not a transaction is expired. This is needed
+    /// since we keep  expired transactions around until they are older than the
+    /// `safe` block in order to be robust to reorg edge cases.
+    ///
     /// The nonce is the first free nonce at or above `status.nonce` (the
     /// account's current onchain transaction count, passed in so nonces
     /// consumed by transactions submitted outside the queue are respected),
@@ -109,7 +121,16 @@ impl TransactionStorage {
     pub async fn next_transaction(
         &self,
         status: Status,
-    ) -> Result<Option<TransactionWithNonce>, Error> {
+    ) -> Result<Option<AllocatedTransaction>, Error> {
+        // Note that, instead of returning the `nonce` and `request` as
+        // separate columns, we instead return a JSON string with the nonce
+        // field already set (**without updating the `request` column**). This
+        // just makes the deserialization on the Rust side more natural (where
+        // we don't need to declare a type just for deserializing a
+        // `AllocatedTransaction` without a nonce and combine the two values
+        // afterwards). The `request` value is not affected by this query,
+        // `json_set(TEXT, ...) -> TEXT` is just a pure transformation on its
+        // inputs to an output JSON string value.
         let Some(request) = sqlx::query_scalar::<_, String>(
             "UPDATE transactions
              SET nonce = MAX(?, COALESCE(
@@ -132,7 +153,7 @@ impl TransactionStorage {
             return Ok(None);
         };
 
-        let transaction = serde_json::from_str::<TransactionWithNonce>(&request)?;
+        let transaction = serde_json::from_str::<AllocatedTransaction>(&request)?;
         Ok(Some(transaction))
     }
 
@@ -152,6 +173,11 @@ impl TransactionStorage {
              WHERE nonce = ?",
         )
         .bind(block.map(i64::try_from).transpose()?)
+        // Note that we encode the fee arguments in hexadecimal notation. This
+        // is because we use Ethereum-style QUANTITY encoding which expects
+        // big integers to be encoded as hex with no leading 0's; something
+        // which is also expected from the `AllocatedTransaction` serialization
+        // implementation.
         .bind(format!("0x{:x}", fees.max_fee_per_gas))
         .bind(format!("0x{:x}", fees.max_priority_fee_per_gas))
         .bind(i64::try_from(nonce)?)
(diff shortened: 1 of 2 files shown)


crates/core/src/tx/storage.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// account's current onchain transaction count, passed in so nonces
+    /// consumed by transactions submitted outside the queue are respected),
+    /// accounting for the nonces of other in-flight transactions. Selecting the
+    /// nonce and reserving it for the transaction happen atomically.
+    pub async fn next_transaction(
+        &self,
+        status: Status,
+    ) -> Result<Option<TransactionWithNonce>, Error> {
+        let Some(request) = sqlx::query_scalar::<_, String>(
+            "UPDATE transactions
```
rmeissner, line 135: nit: this query is quite complex and I am not sure how well versed we all are in SQL, so maybe a comment what it does would be nice. I.e. I had to google the `RETURNING json_set(request, '$.nonce', nonce)` part

crates/core/src/tx/storage.rs
```diff
@@ -0,0 +1,433 @@
...
+impl TransactionStorage {
+    /// Creates a store backed by `pool`, creating the transactions table if it
+    /// does not already exist.
+    pub async fn new(pool: SqlitePool) -> Result<Self, Error> {
+        sqlx::query(
+            "CREATE TABLE IF NOT EXISTS transactions (
+                 id           INTEGER PRIMARY KEY,
+                 request      TEXT    NOT NULL,
+                 expires_at   INTEGER NOT NULL,
+                 nonce        INTEGER DEFAULT NULL,
```
rmeissner, line 73: question: we also have the nonce in the `request` do we need it twice?

crates/core/src/tx/storage.rs
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    /// Marks the transaction with `submission.nonce` as submitted at
+    /// `submission.block`, recording the `submission.fees` it was submitted with
+    /// in the stored request. Errors if no transaction has that nonce.
+    pub async fn record_submission(&self, submission: Submission) -> Result<(), Error> {
+        let Submission { block, nonce, fees } = submission;
+        let updated = sqlx::query(
+            "UPDATE transactions
+             SET submitted_at = ?,
```
rmeissner, line 167: nit: can we use named parameters, personally this would make it easier to read the query

crates/core/src/tx/storage.rs
```diff
@@ -0,0 +1,433 @@
...
+                 request = json_set(
+                     request,
+                     '$.maxFeePerGas', ?,
+                     '$.maxPriorityFeePerGas', ?
+                 )
+             WHERE nonce = ?",
+        )
+        .bind(block.map(i64::try_from).transpose()?)
+        .bind(format!("0x{:x}", fees.max_fee_per_gas))
+        .bind(format!("0x{:x}", fees.max_priority_fee_per_gas))
```
rmeissner, line 182: obersvation: just realized here that the fees are stored as hex (even so it was defined in the type)

crates/core/src/tx/storage.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Returns the in-flight transactions due for (re)submission, ordered by
+    /// nonce: those last submitted at or before `submitted_before`, as well as
+    /// any that were assigned a nonce but never recorded as submitted (so they
+    /// are not stranded holding a reserved nonce).
+    pub async fn stale_submissions(
+        &self,
+        submitted_before: Option<u64>,
+    ) -> Result<Vec<TransactionWithNonce>, Error> {
+        sqlx::query_scalar::<_, String>(
+            "SELECT json_set(request, '$.nonce', nonce)
```
rmeissner, line 277: observation: this json_set makes sense but is surprising to me. I somehow was under the impression that we use `json_set` to update the stored data, but we only use it for in-flight adjustments on the data returned.

Can you outline why you chose not to update the stored data and only adjust in-flight responses? (related to https://github.com/safe-research/safenet/pull/482/changes#r3479874790)

crates/core/src/tx/storage.rs
```diff
@@ -0,0 +1,433 @@
...
+            .next_transaction(Status { nonce: 5, block: 0 })
+            .await
+            .unwrap()
+            .unwrap();
+
+        // The returned transaction carries the assigned nonce.
+        assert_eq!(submitted.nonce, 5);
+
+        // It is now in flight and no longer queued.
+        assert_eq!(storage.count_in_flight().await.unwrap(), 1);
```
rmeissner, line 366: nit: might make sense to check that before assigning the nonce the count in flight was 0.

crates/core/src/tx/storage.rs
```diff
@@ -0,0 +1,433 @@
...
+        let storage = storage().await;
+        storage.enqueue(request("0x5afe01"), 100).await.unwrap();
+        storage.enqueue(request("0x5afe02"), 100).await.unwrap();
+        storage.enqueue(request("0x5afe03"), 100).await.unwrap();
+
+        // Each submission picks the next free nonce above the in-flight ones, in
+        // FIFO order.
+        for (data, nonce) in [("0x5afe01", 5), ("0x5afe02", 6), ("0x5afe03", 7)] {
+            let submitted = storage
+                .next_transaction(Status { nonce: 5, block: 0 })
```
rmeissner, line 387: observation: it is not immediately clear what the `block` is used for. Not sure how to make this more clear, but you kind of have to look into the SQL query to of the `next_transaction` function to see this.

crates/core/src/tx/storage.rs
```diff
@@ -0,0 +1,433 @@
...
+        storage
+            .next_transaction(Status { nonce: 5, block: 0 })
+            .await
+            .unwrap()
+            .unwrap();
+
+        assert!(matches!(
+            storage
+                .record_submission(Submission {
+                    block: Some(42),
```
rmeissner, line 440: nit: there is no test where `block` is None, but not sure if this provides value

rmeissner (approved):

crates/core/src/tx/storage.rs
nlordell: Nice catch, and it is intentional here. We store the transaction as a JSON blob, and the transaction type expects hex-formatted numbers. This means that when we update JSON data in SQLite directly (without roundtripping it from the DB to the service for editing and then back to the DB), we need to store the fields how the type expects to decode them.

I will clarify this as a comment as it is definitely not obvious.

crates/core/src/tx/storage.rs
nlordell: `json_set(TEXT, ...) -> TEXT` takes some string input, decodes it, applies the modifications, encodes the modified blob and returns it. It does not modify JSON in place (see https://sqlite.org/json1.html#jset).

If you want to modify the JSON in a column, you do:

```sql
UPDATE my_json_table
SET my_json_column = json_set(my_json_column, '$.field', "value");
```

`json_set` just applies JSON transformations to a JSON string.

crates/core/src/tx/storage.rs
nlordell: We do not - we store the nonce in its own column and never in the `request`. `json_set` does not update the column unless you do so explicitly.

Added an explainer to the table creation.

crates/core/src/tx/storage.rs
nlordell: Would love to - but not supported by `sqlx`: https://github.com/transact-rs/sqlx/issues/199

crates/core/src/tx/storage.rs
nlordell: Added an explainer.

crates/core/src/tx/storage.rs
nlordell: A test included in #485 will cover this flow, I didn't want to over-test (because reviewing tests is also kind of annoying).

crates/core/src/tx/storage.rs
nlordell: Added.

crates/core/src/tx/storage.rs
nlordell: I will expand the doc comment to explain what block is used for.

PR #484, opened 2026-06-25
Fix Bad Update Error

nlordell:
I noticed the error description did not match the variant name. This PR fixes that.

---

## CLA signature

### DIFF
--- a/crates/core/src/state/mod.rs
+++ b/crates/core/src/state/mod.rs
@@ -22,7 +22,7 @@ pub enum Error {
     /// updates.
     #[error("end of chain")]
     EndOfChain,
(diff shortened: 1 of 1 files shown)


rmeissner (approved):

crates/core/src/state/mod.rs
```diff
@@ -23,7 +23,7 @@ pub enum Error {
     #[error("end of chain")]
     EndOfChain,
     /// Received an update that is either out-of-order or has unexpected data.
```
remedcu: nit:


```suggestion
    /// Received a bad update i.e. either out-of-order or has unexpected data.
```

remedcu (approved):

Issue #465, opened 2026-06-20
Extend Alloy Asserter to Check Inputs


nlordell:
The [`alloy::providers::transport::mock::Asserter`](https://docs.rs/alloy/latest/alloy/providers/transport/mock/struct.Asserter.html) only supports mocking responses from the client, without any facilities for asserting on the request that was used.

This issue captures the work to create a custom `Asserter` that _also_ allows adding asserts on the request data for the input bytes. This will allow our tests to have slightly strong guarantees.

We should propose the improvements to the asserter upstream to the `alloy` crate.

PR #485, opened 2026-06-25
[E 6/6]: Transaction Queue Heartbeat

nlordell:

### Context and Motivation

This PR implements the heartbeat of the transaction queue: block updates. It processes block updates and triggers (re-)submissions of non-executed transactions as well as does some housekeeping in the DB.

### Decisions and Tradeoffs

- The transaction queue was made to be reorg aware
- We do some basic defensive block update input checks like for the state machine. In practice, this should never happen

### Testing

Some high level tests were added to check basic flows. Again, there is the issue that we cannot assert on RPC input (see #465), but the tests do still cover a lot of the behaviour of the component.

---

## CLA signature

### DIFF
--- a/crates/core/src/tx/mod.rs
+++ b/crates/core/src/tx/mod.rs
@@ -6,8 +6,6 @@
 //! local [`signer`], and resubmitting with bumped fees when a transaction is
 //! stuck.
 
-#![allow(dead_code)]
-
 mod fees;
 pub mod signer;
 mod storage;
@@ -20,6 +18,7 @@ use self::{
     types::AllocatedTransaction,
 };
 pub use self::{signer::Signer, types::Transaction};
+use crate::index::BlockUpdate;
 use alloy::{
     eips::{BlockId, eip1559::Eip1559Estimation},
     primitives::U256,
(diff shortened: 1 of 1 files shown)


crates/core/src/tx/mod.rs
```diff
@@ -132,6 +131,72 @@ where
...
+    /// housekeeping: invalidating cached chain state, marking executed
+    /// transactions, pruning finalized and expired ones, resubmitting stale
+    /// transactions and submitting newly queued ones.
+    pub async fn handle_block_update(&mut self, update: BlockUpdate) -> Result<(), Error> {
+        self.nonce_cache = None;
+        self.fee_cache = None;
+        self.balance_cache = None;
+        match update {
+            BlockUpdate::New { number, safe, .. } => {
+                if self.next_block()?.is_some_and(|next| next != number) || safe > number {
```
rmeissner, line 144: question: why check for `safe > number `?

crates/core/src/tx/mod.rs
```diff
@@ -289,3 +366,241 @@ where
         }
     }
 }
+
+#[cfg(test)]
```
rmeissner, line 370: nit: there is not explicit test on the cache invalidation, right? If I see it correctly this is implicitly tested in the tests below, but might make sense to have an explicit test.

rmeissner (approved):

crates/core/src/tx/mod.rs
```diff
@@ -132,6 +131,72 @@ where
...
+                // too far in the past to execute anyway. Put us into the state
+                // where we are waiting for the next new block update, so queued
+                // transactions do not execute right away. This ensures that
+                // transactions that are queued already expired while warping do
+                // execute.
+                self.block = Block::Warping { to };
+            }
+        }
+        Ok(())
+    }
```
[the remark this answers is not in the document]
nlordell: This is intended. If submission fails after the block update, we want to retry it next block. We will not get the same update twice; the stream of updates will go on!

crates/core/src/tx/mod.rs
nlordell: I didn't want to reason about "what happens to our DB when we prune past the latest block?". This is part of the defensive block update validation:

> We do some basic defensive block update input checks like for the state machine. In practice, this should never happen.

crates/core/src/tx/mod.rs
nlordell: It is explicitly checked via assertions of specific RPC requests (which would not be there in case cache was not invalidated).

I am wary about checking internals of how we implement our cache too much...

PR #486, opened 2026-06-25
[G 1/2]: Driver Component

nlordell:
This PR introduces a driver component to facilitate the plumbing between the shared utilities.

This phase (G) was not part of the original plan, but was introduced in order to facilitate the use of the components introduced by the shared crate.

### Context and Motivation

This is the "top level" component that is expected to be initialized by each of the services and used to drive the protocols forward.

### Decisions and Tradeoffs

For now, initialization is manual (in other words, it is expected that the caller initializes all driver components and passes them in). I plan on simplifying this (although without trying on both the sentinel and validator, it is hard to know what parts of the initialization should be automatic, and which parts should be manual.

Transaction queuing was changed to be bundle based, this ensures that all actions after a state transition are processed together and added to the transaction queue storage (in case of an untimely restart).

---

## CLA signature

### DIFF
--- a/crates/core/src/driver.rs
+++ b/crates/core/src/driver.rs
@@ -1,0 +1,156 @@
+//! Service driver.
+//!
+//! Ties the core building blocks together into a running service: it follows
+//! the chain with an [`indexer`](crate::index), feeds block updates to the
+//! [`transaction queue`](crate::tx) for its per-block housekeeping, and feeds
+//! every update to the [`state machine`](crate::state). The actions produced by
+//! the state machine are encoded into transactions by the [`Service`] and queued
+//! for submission.
+
+use crate::{
+    index::{self, Update, Watcher, events::Events},
+    state::{self, StateMachine, StateTransition},
+    tx::{self, Transaction, TransactionQueue},
+};
+use alloy::providers::Provider;
+use serde::{Serialize, de::DeserializeOwned};
+use std::{pin::Pin, time::Duration};
+
+/// How long to wait after a failed step before retrying, to avoid spinning on a
+/// persistent failure (such as an unreachable RPC node).
+const STEP_RETRY_DELAY: Duration = Duration::from_millis(100);
+
+/// Whether the [`Driver::run`] loop should keep processing updates or stop.
+enum Loop {
+    /// Continue with the next iteration.
+    Continue,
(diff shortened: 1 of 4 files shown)


crates/core/src/driver.rs
```diff
@@ -0,0 +1,113 @@
...
+    Transactions(#[from] tx::Error),
+}
+
+/// A Safenet service.
+///
+/// Implementors describe how the service observes the chain (its [`Events`]) and
+/// reacts to it (its [`State`](Service::State) and
+/// [`Transition`](Service::Transition)), and how the resulting
+/// [`Action`](Service::Action)s map onto onchain transactions.
+pub trait Service: StateTransition<Self::State> {
```
rmeissner, line 51: question: tbh this one is weird for me and made it harder to understand.

The `encode_actions` is surprising to me here, as I was not expecting it as part of the "Service". Also that the Service extends StateTransition is not intuitive for me.

Why do we need this trait? Is it to connect the types?

I.e. For me the Driver fits into my mental model of the Service. It would have the following fields:
- watcher
  - > Observers the onchain events
- state (which would also hold the transition function, but imo the driver doesn't need to know about it)
  - > Consumes the events to update the state and produce actions
- encoder 
  - > Convert the actions into transactions
- transactions
  - > Handle the submission of the transactions onchain

crates/core/src/driver.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+    }
+
+    /// Runs the service, processing indexer updates forever.
+    ///
+    /// A failed step is retried on the next iteration rather than stopping the
+    /// service.
+    pub async fn run(mut self) {
+        loop {
+            let _ = self.step().await;
```
rmeissner: question: how is this driver stopped?

rmeissner (approved):

crates/core/src/driver.rs
nlordell: It was introduced in #487, and merged into here - it stops on Ctrl-C.

crates/core/src/driver.rs
nlordell: As discussed in the office hours, the `Service` represents the "service-specific-functions" more than an actual service component.

We decided we would rename this, I will do it in a follow up to not break my current PR stack.

PR #502, opened 2026-06-30
Refactor Coverage Report Submission

nlordell:
### Context and Motivation

We introduced Rust coverage reports in #501, and noticed some issues in the process:

- In an ideal world, we would not depend on 3rd-party GitHub actions for generating coverage reports (because of the supply chain risk)
- The required GitHub actions was taking `ci / check` or `rust / check` and didn't actually require both to pass
- Generally speaking, generating coverage reports should not be required as a mandatory CI checks required for merging

In order to address the issues above, we moved coverage generation to its own workflow, and merged the Rust checks into the CI workflow (now that we've fully embraced the Rust).

### Decisions and Tradeoffs

- Instead of keeping a `report-coverage` action, we instead moved the logic to a `generate-coverage-report.sh` script that can more easily be run locally in order to test the coverage report generation (and generally check the stats locally).
- We now have actions for installing the required tools (`lcov` and `cargo-llvm-cov`) in a way that does not depend on 3rd party actions
- The Rust checks are now part of the CI checks (the distinction is not useful); this also means that if CI fails, then the merge will be blocked.

---

## CLA signature


nlordell:
It looks like the `post_coverage_comment` is running with the workflow file from `main` branch (https://github.com/safe-research/safenet/actions/runs/28458631494). This doesn't work because the report is generated separately in a new `coverage` workflow (while the old workflow expects the `ci` workflow). The `coverage-report.md` looks good though:

https://github.com/safe-research/safenet/actions/runs/28458393635

I suspect that, once this PR is merged, this will resolve itself.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,62 @@
...
+      - uses: ./.github/actions/install-cargo-llvm-cov
+
+      - run: npm ci
+      - run: scripts/generate_coverage_report.sh
+
+      # Post the report as a step summary (safe for fork PRs) and upload it as
+      # an artifact so the privileged post_coverage_comment workflow can post it
+      # as a PR comment.
+      - name: Post Coverage Step Summary
+        if: github.event_name == 'pull_request'
```
rmeissner: Nit: this should always be possible

If we don't post any result then coverage is only visible in the tun logs, not sure if we then need to run this workflow at all

rmeissner (approved):

.github/workflows/coverage.yml
nlordell: Ah nice catch - this is leftover from splitting off the coverage bits from ci.yml into coverage.yml.

Will fix.

The nice thing about this part of the job is that we see coverage information in the run summary in case posting the GH comment fails.

PR #503, opened 2026-06-30
[B 1]: Add detector and state

rmeissner:
Add initial implementation of decetector and state. The detector is mostly a placeholder that checks against a block list.

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/detector.rs
+++ b/crates/sentinel/src/detector.rs
@@ -1,0 +1,62 @@
+use std::collections::HashSet;
+
+use alloy::primitives::Address;
+
+use crate::bindings::consensus::SafeTransaction;
+
+/// Decides whether a proposed oracle transaction should be approved.
+///
+/// Approves every transaction whose destination is not in the blocklist;
+/// mirrors `createDetector` from `sentinel/detector.ts`.
+pub struct Detector {
+    blocklist: HashSet<Address>,
+}
+
+impl Detector {
+    #[cfg_attr(not(test), expect(dead_code))]
+    pub fn new(blocklist: impl IntoIterator<Item = Address>) -> Self {
+        Self {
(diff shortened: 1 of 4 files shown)


crates/sentinel/src/detector.rs
```diff
@@ -0,0 +1,76 @@
...
+///
+/// Approves every transaction whose destination is not in the blocklist;
+/// mirrors `createDetector` from `sentinel/detector.ts`.
+pub struct Detector {
+    blocklist: HashSet<Address>,
+}
+
+impl Detector {
+    #[cfg_attr(not(test), expect(dead_code))]
+    pub fn new(blocklist: Vec<Address>) -> Self {
```
nlordell: Nit: 
```suggestion
    pub fn new(blocklist: impl IntoIterator<Item = Address>) -> Self {
```

This should let you create with both `Vec` like you do now, and any other thing that is an iterator (for example an array like: `Detector::new([...])`)

crates/sentinel/src/detector.rs
```diff
@@ -0,0 +1,76 @@
...
+            value: U256::ZERO,
+            data: Bytes::new(),
+            operation: Operation::CALL,
+            safeTxGas: U256::ZERO,
+            baseGas: U256::ZERO,
+            gasPrice: U256::ZERO,
+            gasToken: Address::ZERO,
+            refundReceiver: Address::ZERO,
+            nonce: U256::ZERO,
+        }
```
nlordell, line 41: Nit: you might be able to derive `Default` in the `sol` macro, then you can write this much more succinctly:

```suggestion
        SafeTransaction {
            to,
            ..Default::default()
        }
```

nlordell (approved):

nlordell:
LGTM, just some minor style suggestions

PR #504, opened 2026-06-30
[D 1]: Add (partial) Sentinel Service

rmeissner:
Partial implementation of the Sentinel service. It only contains the encoding of actions (and the required actions enum) for now.

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/action.rs
+++ b/crates/sentinel/src/action.rs
@@ -1,0 +1,28 @@
+use alloy::primitives::{B256, U256};
+
+/// An action emitted by the sentinel during a state transition.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(not(test), expect(dead_code))]
+pub enum SentinelActionKind {
+    /// Approve the fee token to be spent by the oracle (bond pre-authorisation).
+    ApproveToken { bond: U256 },
+    /// Cast an approve vote for the request with the given id.
+    CommitApprove { id: B256 },
+    /// Cast a deny vote for the request with the given id.
+    CommitDeny { id: B256 },
+    /// Finalise the committed vote for the request with the given id.
+    Finalize { id: B256 },
+    /// Claim the bond and reward for the request with the given id.
+    Claim { id: B256 },
+}
+
+/// A sentinel action tagged with the voting-deadline block by which it is
+/// no longer useful.
+///
+/// The deadline is forwarded to the `TransactionQueue` as the per-tx expiry
+/// so the queue can drop it if it goes unsubmitted past that block.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct SentinelAction {
+    pub kind: SentinelActionKind,
+    pub expires_at: u64,
+}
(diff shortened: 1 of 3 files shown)


crates/sentinel/src/state.rs
```diff
@@ -0,0 +1,72 @@
+use alloy::primitives::{B256, U256};
+use serde::{Deserialize, Serialize};
+
+/// Actions emitted by the sentinel during state transitions.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum SentinelAction {
```
nlordell: I was completely confused for a second, because this file looked completely different to the `state` module from #503, but this is only the actions.

question: Why not use an `action` module for this?

crates/sentinel/src/state.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub expires_at: u64,
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn action_serde_roundtrip() {
+        let cases: &[SentinelAction] = &[
```
nlordell: Type annotation should not be necessary here:
```suggestion
        let cases = &[
```

crates/sentinel/src/state.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// Cast a deny vote for the request with the given id.
+    CommitDeny { id: B256 },
+    /// Finalise the committed vote for the request with the given id.
+    Finalize { id: B256 },
+    /// Claim the bond and reward for the request with the given id.
+    Claim { id: B256 },
+}
+
+/// A [`SentinelAction`] tagged with the voting-deadline block by which it is
+/// no longer useful, matching the TS `SentinelActionWithTimeout` pattern.
```
nlordell: nit: since we will deprecate the TS code, references like this are (IMO) going to be confusing in the future.


```suggestion
/// no longer useful.
```

crates/sentinel/src/state.rs
nlordell: uber naming nit: Typically, when we have the pattern "(sum type) product type" you typically call the sum type `${NAME}Kind` and the product type `${NAME}` (so here, `SentinelActionKind` and `SentinelAction` (for example, from the Rust standard library: https://doc.rust-lang.org/stable/std/mem/type_info/struct.Type.html#structfield.kind).

crates/sentinel/src/state.rs
```diff
@@ -0,0 +1,72 @@
+use alloy::primitives::{B256, U256};
+use serde::{Deserialize, Serialize};
+
+/// Actions emitted by the sentinel during state transitions.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
nlordell, line 6: question: why is serialization necessary on this type?

rmeissner: Good, question. I thin that was assumed because in the TS version these were serialized into the db, but here we only serialize the transactions resulting from an action, so it should not be necessary.

crates/sentinel/src/state.rs
rmeissner: will change. Iirc this was inspired by the TS version.

nlordell (approved):

PR #505, opened 2026-07-01
[B 2i]: Add State transition event handling

rmeissner:
Added the event handling for the Sentinel state transitions. The state transition function implementations are empty and will be implemented in follow up PRs.

We currently also handle event that do not trigger any state changes (e.g. Claimed and DisputeResolved) to have an exhaustive list.

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/bindings.rs
+++ b/crates/sentinel/src/bindings.rs
@@ -5,15 +5,6 @@ pub mod oracle {
 
     sol! {
         #[derive(Debug)]
-        enum SentinelOracleRequestState {
-            PENDING,
-            FROZEN,
-            RESOLVED_APPROVED,
-            RESOLVED_DENIED,
-            TIMED_OUT
-        }
-
-        #[derive(Debug)]
         contract SentinelOracle {
             event NewRequest(
                 bytes32 indexed requestId,
@@ -34,17 +25,6 @@ pub mod oracle {
                 address indexed proposer,
                 bytes result,
                 bool approved
-            );
-            event Claimed(
-                bytes32 indexed requestId,
-                address indexed sentinel,
-                uint256 bondReturn,
-                uint256 feeReward
-            );
-            event DisputeResolved(
-                bytes32 indexed requestId,
-                SentinelOracleRequestState outcome,
-                uint256 slashed
             );
 
             function commitApprove(bytes32 requestId) external;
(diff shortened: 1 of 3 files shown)


crates/sentinel/src/service.rs
```diff
@@ -86,6 +132,52 @@ impl SentinelService {
...
+            SentinelEvents::Oracle(SentinelOracle::SentinelOracleEvents::Committed(event)) => {
+                self.handle_committed(state, event)
+            }
+            SentinelEvents::Oracle(SentinelOracle::SentinelOracleEvents::OracleResult(event)) => {
+                self.handle_resolved(state, event)
+            }
+            SentinelEvents::Oracle(
+                SentinelOracle::SentinelOracleEvents::Claimed(_)
+                | SentinelOracle::SentinelOracleEvents::DisputeResolved(_),
+            ) => (state, Vec::new()),
```
nlordell: ~~nit~~ question: Why handle these events at all? Can't you just remove them from the bindings? It just adds additional node/RPC pressure to additional query these events if there is no plan whatsoever to include them.

rmeissner: I mentioned it in the PR description, but yeah can remove them

Edit: is this a change request, since you did not approve?

nlordell: Ooops, I meant it as a `question` I wanted answer before approving. Edited my initial comment.

nlordell: > I mentioned it in the PR description, but yeah can remove them

I saw, I just didn't understand the rationale behind it and wanted to ask for more clarification.

I would prefer removing it (less pressure on the RPC endpoint/node), but not a blocker. I was just trying to understand the reasoning behind it.

nlordell (approved):

crates/sentinel/src/bindings.rs
```diff
@@ -35,17 +26,6 @@ pub mod oracle {
                 bytes result,
                 bool approved
             );
-            event Claimed(
-                bytes32 indexed requestId,
-                address indexed sentinel,
-                uint256 bondReturn,
-                uint256 feeReward
-            );
-            event DisputeResolved(
```
rmeissner, line 44: Currently this is not needed, we will potentially have to add it again to handle the arbitration flow

PR #509, opened 2026-07-01
[B 2ii]: Handle propose and new request transitions

rmeissner:
Implement the first part of the sentinel oracle flow: voting. For this new proposal are tracked to extract the transaction data and then once a request is recieved the vote is issued.

### Design Decisions
- The voting window is a config param instead of being read from chain, as it is immutable in the contract

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/service.rs
+++ b/crates/sentinel/src/service.rs
@@ -5,7 +5,9 @@ use crate::{
         consensus::Consensus,
         oracle::{ERC20, SentinelOracle},
     },
-    state::State,
+    detector::Detector,
+    hashing::oracle_tx_proposal_hash,
+    state::{RequestStatus, SentinelRequestState, State},
 };
 use alloy::{
     primitives::{Address, U256},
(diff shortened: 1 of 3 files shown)


crates/sentinel/src/service.rs
```diff
@@ -5,49 +5,128 @@ use crate::{
...
+        );
+        let approve = self.detector.approve(&event.transaction);
+        state.0.insert(
+            request_id,
+            SentinelRequestState {
+                deadline: block.saturating_add(self.voting_window),
+                approve,
+                status: RequestStatus::Preparing,
+            },
+        );
```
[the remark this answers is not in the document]
rmeissner: > Missing Chain ID Validation:

This is not correct, as these are not the same chain ids (self.chain_id is the id of the chain where the oracle/ consensus lives, the event.chainId is the chain id where the Safe transaction should be executed).

crates/sentinel/src/service.rs
```diff
@@ -19,35 +21,112 @@ use safenet_core::{Service, index::EventLog, state::StateTransition, tx::Transac
...
-    pub fn new(oracle: Address, fee_token: Address) -> Self {
-        Self { oracle, fee_token }
+    pub fn new(
+        oracle: Address,
+        fee_token: Address,
+        consensus: Address,
+        chain_id: U256,
+        voting_window: u64,
+        detector: Detector,
+    ) -> Self {
```
nlordell, line 44: nit: Might make sense to initialize from a configuration object.

nlordell (approved):

crates/sentinel/src/service.rs
rmeissner: to avoid too many parameters?

nlordell: Exactly. Doesn't need to be done here, but I imagine that most of these come from the configuration file. You can create a `sentinel: SentinelConfig` as a section on the `Config` object so you can specify all sentinel-specific configurations under a `[sentinel]` section in the TOML.

rmeissner: will keep in mind when we wire this up.

PR #510, opened 2026-07-01
[B 2iii]: Handle claiming flow

rmeissner:
Implement the required state transitions to claim tokens from the sentinel oracle. For this it is required to track the committed state and react to the oracle result event.

Implementation is fully based on the typescript implementation.

### Design Decisions
- Use max uint256 as the expire date, as claiming should always be tried, to avoid manual intervention to recover funds that are locked up.

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/service.rs
+++ b/crates/sentinel/src/service.rs
@@ -3,7 +3,7 @@ use crate::{
     bindings::{
         SentinelEvents,
         consensus::Consensus,
-        oracle::{ERC20, SentinelOracle},
+        oracle::{ERC20, ResolveReason, SentinelOracle},
     },
     detector::Detector,
     hashing::oracle_tx_proposal_hash,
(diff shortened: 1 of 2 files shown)


crates/sentinel/src/service.rs
```diff
@@ -132,25 +136,62 @@ impl SentinelService {
...
+        };
+        let vote_won = reason == ResolveReason::TIMEOUT || event.approved == existing.approve;
+        let actions = if vote_won {
+            vec![SentinelAction {
+                kind: SentinelActionKind::Claim {
+                    id: event.requestId,
+                },
+                // Claiming has no onchain deadline, so the action must never
+                // expire in the `TransactionQueue`.
+                expires_at: u64::MAX,
```
nlordell, line 189: uber-nano-nit: this makes me think that `SentinelAction` should not have an `expires_at` (since its value has no meaning for certain kinds). If _most_ sentinel actions need this, then I think it is OK as-is; otherwise I would consider lifting the `expires_at` to where it is relevant, or further breaking down the type ("sentinel game actions" that expire, "accounting actions" that don't, etc.).

nlordell (approved):

crates/sentinel/src/service.rs
rmeissner: that is true, for now I would leave it as is, but definetly worth observing.

PR #512, opened 2026-07-01
Pure State Transition Functions

nlordell:
This PR is an attempt at implementing pure state transition functions with an effect handling system. This allows the state transition function to no longer take a `&mut self` and no longer be async. Processes that require side-effects must be done through the effect system.

### Context and Motivation

When we introduced the state machine component, we originally made the state transition function be an `async fn(&mut self)`. While state transition functions are _generally_ side-effect free, this is not generally true for all cases in the validator:

1. Nonce generation and use is impure, if we use a nonce once, and then we reorg and require the same nonce again, **we cannot reuse it**
2. Random coefficients during DKG is impure, if we reorg we cannot generate new ones because the action with their commitment may already have been submitted making them non-adjustable for a given DKG ceremony

### Design and Tradeoffs

The implementation changes the state transition function to be pure (`fn(&self)`) and adds an effect system where the impure operations can be modeled.

- Nomenclature for the inputs and output were inspired by [Elm](https://guide.elm-lang.org/) which was a pioneer in the FRP model that we apply here:
  - `Message`s are inputs to the state transition (or "update") function; previously there were `new_block` and `event` and they are now `NewBlock`, `Event` and `Resume`. `Resume` is a message that provides a result from performing an effect
  - `Command`s are the outputs to the state transition beside the new state; previously they were `Action`s, and now they are `Action`s and `Effect`s. `Effect`s are the new system which allow for some internal impure computation to happen which must produce some result (error results must be provided back to the state machine), while actions are dedicated for onchain transacions.
- As previously discussed, the `Service` is now broken up into different components that implement the different parts required by the `Driver`. This allows clearer isolation between the pure and impure parts (not to mention things like contract addresses are not generally needed for state transitions and only used for transaction encoding).

#### Porting Sentinel Service

To avoid making this PR too big, I just ported the Sentinel service 1:1. Now that we split the service into its components, it would make sense for the service to split up into parts, but as a separate PR. The Sentinel specific changes were kept in a separate commit to make reviewing easier: https://github.com/safe-research/safenet/pull/512/commits/f2f4a94170318a5a722764a8d6710f54b52741c6.

---

## CLA signature

### DIFF
--- a/crates/core/src/state/mod.rs
+++ b/crates/core/src/state/mod.rs
@@ -10,7 +10,7 @@ use self::storage::SnapshotStore;
 use crate::index::{BlockUpdate, EventLog, EventUpdate, Update};
 use serde::{Serialize, de::DeserializeOwned};
 use sqlx::SqlitePool;
-use std::{mem, range::RangeInclusive};
+use std::{collections::VecDeque, convert::Infallible, mem, range::RangeInclusive};
 use tokio::sync::Mutex;
 
 /// Error produced by the [`StateMachine`].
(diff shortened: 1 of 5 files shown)


crates/core/src/state/mod.rs
```diff
@@ -206,11 +272,62 @@ where
...
+            effects,
+            state,
+            actions: vec![],
+        }
+    }
+
+    async fn apply(mut self, message: Message<T::Event, T::Resume>) -> Self {
+        let (state, commands) = self.transition.apply_transition(self.state, message);
+        self.state = state;
+        self.actions.reserve(commands.len());
```
rmeissner, line 313: question: why is this a good choice?

crates/core/src/state/mod.rs
```diff
@@ -227,11 +344,30 @@ fn is_next_in_range(range: impl Into<RangeInclusive<u64>>, sub: RangeInclusive<u
...
+        match effect {}
+    }
+}
+
+impl<E, R, H> EffectHandler<E, R> for &mut H
+where
+    H: EffectHandler<E, R>,
+{
+    fn perform_effect(&mut self, effect: E) -> impl Future<Output = R> {
+        (*self).perform_effect(effect)
```
rmeissner: question: I don't understand this, looks circular 😅 

Edit: cleared up in the office, this is to dereference arbitrary levels.

@nlordell wants to check if still necessary.

rmeissner (approved):

crates/core/src/state/mod.rs
nlordell: Added a justifying comment.

crates/core/src/state/mod.rs
nlordell: This is no longer needed, so I removed it.

PR #513, opened 2026-07-02
[E 1]: Setup Sentinel Driver

rmeissner:
The required config parameters are defined and loaded, the Sentinel Service instantiated and then used to wire up the Sentinel Driver in main.rs.

### Implementation Decision

- There are currently not defaults for the confi that can be refined in later phases.

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/config.rs
+++ b/crates/sentinel/src/config.rs
@@ -1,7 +1,10 @@
-use safenet_core::observability;
+use alloy::primitives::Address;
+use safenet_core::{driver, observability, tx::Signer};
 use serde::Deserialize;
+use sqlx::sqlite::SqliteConnectOptions;
 use std::path::Path;
 use tokio::{fs, io};
+use url::Url;
 
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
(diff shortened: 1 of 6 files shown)


crates/sentinel/src/detector.rs
```diff
@@ -8,12 +8,12 @@ use crate::bindings::consensus::SafeTransaction;
 ///
 /// Approves every transaction whose destination is not in the blocklist;
 /// mirrors `createDetector` from `sentinel/detector.ts`.
+#[derive(Clone)]
 pub struct Detector {
     blocklist: HashSet<Address>,
```
nlordell: nit: AFAIU, the detector is created once and then can be shared in multiple places, but the set of addresses never changes while running. I would change this to:


```suggestion
#[derive(Clone)]
pub struct Detector {
    blocklist: Arc<HashSet<Address>>,
```

So the actual hash set is shared instead of duplicated.

crates/sentinel/src/config.rs
```diff
@@ -24,3 +51,26 @@ impl Config {
...
+    fn default() -> Self {
+        Self {
+            rpc: "http://localhost:8545".parse().unwrap(),
+            // The default Anvil signer.
+            signer: SigningKey::from_slice(
+                b"\xac\x09\x74\xbe\xc3\x9a\x17\xe3\x6b\xa4\xa6\xb4\xd2\x38\xff\x94\
+                  \x4b\xac\xb4\x78\xcb\xed\x5e\xfc\xae\x78\x4d\x7b\xf4\xf2\xff\x80",
+            )
+            .map(Signer::new)
+            .unwrap(),
```
nlordell: nit: As discussed in the office, this should probably not have defaults and be a required configuration.

Same thing applies to the `validator::Config`, but I can settle that in a separate PR.

crates/sentinel/src/config.rs
```diff
@@ -11,10 +14,34 @@ pub enum Error {
...
+    pub consensus: Address,
+    /// The ERC-20 fee token approved for bonds.
+    pub fee_token: Address,
+    /// The number of blocks a `Preparing` request is kept alive for before
+    /// being cleaned up.
+    pub voting_window: u64,
+    /// Transaction destinations that are always denied.
+    // TODO(epic Phase E2): flesh this out further, e.g. loading the blocklist
+    // from a nested `[detector]` table with its own deserialization tests.
+    pub blocklist: Vec<Address>,
```
nlordell, line 53: nit: It might make sense to have sentinel-specific configurations in its own object.

This allows you to clearly separate the things that are needed by the driver and for observability, from the shared things (the RPC and DB is shared by everyone) and the configuration bits that _only_ the sentinel's internal logic cares about (like the fee token and block list).

It will also make the TOML be separated into nice sections which I find aesthetically pleasing.

NOte that you don't have to do the grouping now (it might make sense to wait until the sentinel is complete so you have a clearer picture of which grouping makes sense), but I do think it would be a welcome refactor, and allow you to pass a larger config struct around for initializing the sentinel service.

nlordell (approved):

PR #517, opened 2026-07-02
[A 1/2] Implement Contract Bindings

nlordell:
### Decisions and Tradeoffs

We chose to implement contract bindings separately from the `sentinel` crate. In the future, we can consider adding a `bindings` crate in order to generate the bindings via the proc-macro only once; we consider that a future improvement.

---

## CLA signature

### DIFF
--- a/crates/validator/src/bindings.rs
+++ b/crates/validator/src/bindings.rs
@@ -1,0 +1,273 @@
+//! `sol!`-generated bindings for the onchain contracts the validator interacts
+//! with including the [`ValidatorEvents`] set the state machine consumes.
+//!
+//! Function bindings only need their calldata (the `*Call` types), so return
+//! values are transcribed only where a return is actually decoded onchain
+//! (`getValidatorStaker`); the mutating calls omit them.
+
+use alloy::sol;
+use safenet_core::watcher_events;
+
+sol! {
+    /// A secp256k1 point in affine (uncompressed) coordinates, as encoded
+    /// onchain (`Secp256k1.Point`).
+    #[derive(Debug, Default)]
+    struct Point {
+        uint256 x;
+        uint256 y;
+    }
+
+    /// A FROST group signature (`FROST.Signature`): the group commitment point
+    /// `r` and the scalar component `z`. Onchain attestations and completed
+    /// signatures share this shape.
+    #[derive(Debug, Default)]
+    struct Attestation {
+        Point r;
+        uint256 z;
+    }
+
+    /// Safe transaction operation type; mirrors `Enum.Operation` onchain.
+    #[derive(Debug, Default)]
+    enum Operation {
+        #[default]
+        CALL,
+        DELEGATECALL,
+    }
+
+    /// A full Safe transaction as carried by the `(Oracle)TransactionProposed`
+    /// events (the 12-field `SafeTransaction.T` tuple).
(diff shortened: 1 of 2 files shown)


remedcu:
Is this something which can be generated with a script based on the contracts provided?

nlordell:
> Is this something which can be generated with a script based on the contracts provided?

In theory yes, and worth investigating.

crates/validator/src/bindings.rs
```diff
@@ -0,0 +1,273 @@
...
+            address safe,
+            bytes32 safeTxStructHash,
+            bytes32 signatureId
+        ) external;
+        function setValidatorStaker(address staker) external;
+        function getValidatorStaker(address validator) external view returns (address staker);
+    }
+
+    #[derive(Debug)]
+    contract Coordinator {
```
rmeissner, line 185: nit/ question: did you consider splitting the coordinator and consensus (and oracle) into separate mods?

crates/validator/src/bindings.rs
```diff
@@ -0,0 +1,273 @@
...
+    enum Operation {
+        #[default]
+        CALL,
+        DELEGATECALL,
+    }
+
+    /// A full Safe transaction as carried by the `(Oracle)TransactionProposed`
+    /// events (the 12-field `SafeTransaction.T` tuple).
+    #[derive(Debug, Default)]
+    struct SafeTransaction {
```
rmeissner, line 40: question: Somehow I was expecting to also see the SafeTx struct for EIP-712 (used in the oracle proposal hash), but maybe that is not relevant yet.

rmeissner (approved):

rmeissner:
> In theory yes, and worth investigating.

I would do this once we think about moving them to a shared crate.

remedcu (approved):

crates/validator/src/bindings.rs
nlordell: `sol!` macro already creates modules for contracts:

image

This was different for the `SafeTransaction`/`SafeTx` because those were top-level definitions and not part of a contract.

crates/validator/src/bindings.rs
nlordell: I don't know whether or not it is needed - I think I plan on using the `safeTxHash` that comes with the event which means that I do not need to compute the 712 struct hash locally.

I may be wrong, in which case we may need to add an additional binging for the SafeTx type.

nlordell:
@remedcu and @rmeissner - I took a quick look at using `sol!("path/to/Solidity.sol")` and it unfortunately does not work for us

1. The macro does not support Solidity that contains identifiers with `$` (which our code does)
2. It does not follow imports (meaning it cannot resolve structs that are imported for other files)

I also tried importing ABI JSON, but that also doesn't work: we lose the `struct` compatibility across contract bindings. In particular, we would have a `sol!(Consensus, "path/to/Consensus.json")` and `sol!(Coordinator, "path/to/FROSTCoordinator.sol")`, both use structs like `Secp256k1.Point` internally, but the macro would generate _different_ types `Consensus::Point` and `FROSTCoordinator::Point` that are not equivalent (despite having the exact same shape).

While I would have loved to not do the manual translation, it looks like it gives us the most ergonomics ATM.

PR #518, opened 2026-07-02
[A 2/2] Implement Validator Configuration

nlordell:
### Context and Motivation

This PR adds the validator configuration that can be loaded from the TOML file. It is included as a separate `ValidatorConfig` section so it can more easily be passed to the validator service component when it is implemented.

### Decisions and Tradeoffs

I removed two configuration options that exist in the TypeScript counterpart:

- The coordinator address (this is intended to be read onchain from the consensus contract instead of requiring a separate configuration option).
- The `skip_genesis` configuration (just like the coordinator address, this will be read onchain based on the state of the consensus contract)

Additionally, as discussed in the offline, we decided that not all options should have a default - we distinguish between mandatory and optional configuration options now.

---

## CLA signature

### DIFF
--- a/crates/validator/src/config.rs
+++ b/crates/validator/src/config.rs
@@ -1,8 +1,10 @@
-use alloy::signers::k256::ecdsa::SigningKey;
+#![cfg_attr(not(test), expect(dead_code))]
+
+use alloy::primitives::{Address, B256};
 use safenet_core::{driver, observability, tx::Signer};
 use serde::Deserialize;
 use sqlx::sqlite::SqliteConnectOptions;
-use std::path::Path;
+use std::{num::NonZeroU64, path::Path};
 use tokio::{fs, io};
 use url::Url;
 
@@ -18,7 +20,7 @@ pub enum Error {
 }
 
 #[derive(Debug, Deserialize)]
-#[serde(default, deny_unknown_fields)]
+#[serde(deny_unknown_fields)]
 pub struct Config {
     /// The RPC endpoint used to initialize the chain provider.
     pub rpc: Url,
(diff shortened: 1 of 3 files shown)


crates/validator/src/config.rs
```diff
@@ -84,6 +151,34 @@ mod tests {
...
+            config.validator.consensus,
+            address!("0x1111111111111111111111111111111111111111")
+        );
+        assert_eq!(config.validator.staker, None);
+        assert!(config.validator.participants.is_empty());
+        assert!(config.validator.oracles.is_empty());
+        assert_eq!(config.validator.genesis_salt, None);
+        assert_eq!(
+            config.validator.blocks_per_epoch,
+            ValidatorConfig::default_blocks_per_epoch()
```
rmeissner: nit: isn't this kind of testing serde?

remedcu (approved):

crates/validator/src/config.rs
```diff
@@ -44,35 +49,97 @@ impl Config {
...
-
     use super::*;
+    use alloy::primitives::{address, b256};
 
     #[test]
-    fn deserializes_required_fields_and_flattened_driver_config() {
-        // Only the required fields are set; the driver config (flattened
-        // alongside them) and observability fall back to their defaults.
-        let config = toml::from_str::<Config>("").unwrap();
+    fn deserializes_required_fields() {
```
rmeissner, line 129: nit: Personally I would have expected that required fields are required and then that it generally works with the required fields (but without checking defaults), so the first part of this test.

But we could argue that we are mostly testing the toml and serde crate here ... so not sure how much we want to test here.

rmeissner (approved):

crates/validator/src/config.rs
nlordell: Removed some of the over testing.

PR #519, opened 2026-07-03
[F 1]: Interop tests

rmeissner:
Add a new sentinel integration test script that runs the TS and Rust version together to ensure interoperability.

The integration test workflow has been adjusted to also run this test.

- Design Decisions
  - Use a MockCoordinator to avoid FROST group initialization.

---

## CLA signature

### DIFF
--- a/scripts/run_sentinel_integration_test.sh
+++ b/scripts/run_sentinel_integration_test.sh
@@ -1,0 +1,271 @@
+#!/bin/bash
+# Interop/integration test for the sentinels (epic Phase F1): runs the TS and
+# Rust sentinel implementations side by side against the same dispute on
+# Anvil, and asserts they agree (no arbitration) and settle fees/bonds
+# correctly. Unlike scripts/run_integration_test.sh, this does not require a
+# full validator/FROST genesis: a `TestConsensus` contract stands in for
+# `Consensus`, only emitting the `OracleTransactionProposed` event the
+# sentinels need.
+set -eo pipefail
+# Job control, so each `&`-backgrounded command below gets its own process
+# group (its PID doubling as its PGID) instead of sharing this script's.
+# Cleanup then kills each job's *group* (`kill -- -$pid`), reaping the
+# node/tsx or compiled-binary grandchildren `npm run`/`cargo run` spawn,
+# without touching whatever else happens to share this script's own process
+# group (e.g. a CI runner's step wrapper) the way `kill 0` would.
+set -m
+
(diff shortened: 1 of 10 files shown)


crates/sentinel/src/action.rs
```diff
@@ -1,5 +1,13 @@
 use alloy::primitives::{B256, U256};
 
+/// The `expires_at` for an action that must never expire in the
+/// `TransactionQueue`.
+///
+/// Not `u64::MAX`: the queue's SQLite storage persists block numbers as
+/// `i64`, and `u64::MAX` overflows that conversion (`Storage(Overflow)`).
+/// `i64::MAX` blocks away is still effectively never.
+pub const NEVER_EXPIRES: u64 = i64::MAX as u64;
```
nlordell: request change: This is a pretty big code-smell to me (and I didn't think about "never expire" transactions when designing the transaction queue 😅).

Can we change the action encoder interface to be `(Transaction, Option<u64>)` where `None` (or `NULL` in SQLite) is used to indicate "never expires".

I would ask to do that as a separate PR (ideally, with just this change and then rebase this PR over top of that one); doing it as a follow up is also acceptable if that is easier for you.

scripts/run_sentinel_integration_test.sh
```diff
@@ -0,0 +1,248 @@
...
+fee_token = "$FEE_TOKEN"
+voting_window = $VOTING_WINDOW
+blocklist = []
+
+[index]
+block_time = $((BLOCK_TIME_SECONDS * 1000))
+EOF
+
+echo "Starting the Rust sentinel..."
+"$ROOT/target/debug/sentinel" --config-file "$RUST_SENTINEL_CONFIG" >"$ROOT/rust_sentinel_logs.txt" 2>&1 &
```
nlordell: nit: I would use `cargo run` directly.

```suggestion
cargo run --package sentinel -- --config-file "$RUST_SENTINEL_CONFIG" >"$ROOT/rust_sentinel_logs.txt" 2>&1 &
```

nlordell (changes requested):
LGTM, requesting one change around the expiry code smell.

crates/sentinel/src/action.rs
rmeissner: I will create a separate PR and then rebase to that

rmeissner: https://github.com/safe-research/safenet/pull/522

nlordell (approved):

PR #522, opened 2026-07-03
Make expires_at optional

rmeissner:
Some actions never expire (i.e. claiming rewards from the Sentinel oracle) and therefore should not expire. Changing epires_at to an Option avoids the workaround using i64::max.

---

## CLA signature


crates/core/src/tx/storage.rs
nlordell: nit: I would prefer if this is explicit:


```suggestion
        sqlx::query("DELETE FROM transactions WHERE nonce IS NULL AND expires_at IS NOT NULL AND expires_at <= ?")
```

nlordell (approved):

PR #523, opened 2026-07-03
[B 1]: Implement ECDH

nlordell:
This PR introduces a FROST module to the validator and implements the ECDH protocol used for secret share encryption.

> [!IMPORTANT]
> The order of the phase B sub-phases is different from what was specified in the epic; I chose a slightly different order to implement things that made more sense to me.

### Decisions and Tradeoffs

The `frost` module in the validator is designed to provide an interface that works with the Solidity interface of the FROST coordinator contracts. That is, it handles marshaling to and from Solidity ABI structs to internal `frost-secp256k1` and `k256` representations internally.

The `EncryptionKey` implements serialization traits so that it can read and write to an SQLite store.

---

## CLA signature

### DIFF
--- a/crates/validator/src/frost/ecdh.rs
+++ b/crates/validator/src/frost/ecdh.rs
@@ -1,0 +1,146 @@
+//! ECDH-XOR encryption of FROST secret shares for the onchain publishing
+//! channel.
+
+use crate::bindings;
+
+use super::{error::Error, marshal};
+use alloy::primitives::B256;
+use k256::{
+    NonZeroScalar, ProjectivePoint, Scalar,
+    elliptic_curve::{
+        hash2curve::{self, ExpandMsgXmd},
+        point::AffineCoordinates as _,
+        zeroize::{Zeroize, ZeroizeOnDrop},
+    },
+    sha2::Sha256,
+};
+use rand::{CryptoRng, RngCore};
+use serde::{Deserialize, Serialize};
+
+/// A locally-generated ECDH encryption key. The secret scalar is sampled by the
+/// effect handler and never leaves the secret store; only [`public_key`] is
+/// published onchain.
+///
+/// [`public_key`]: EncryptionKey::public_key
+#[derive(Clone, Deserialize, Serialize)]
+pub struct EncryptionKey(NonZeroScalar);
+
+impl EncryptionKey {
+    /// Samples a fresh encryption key from `rng`.
+    pub fn generate<R>(mut rng: R) -> Self
+    where
+        R: CryptoRng + RngCore,
+    {
+        let mut entropy = [0; 32];
+        rng.fill_bytes(&mut entropy);
+        Self(hash_to_scalar(b"enc", &entropy))
+    }
+
+    /// The public key `q` published onchain for peers to encrypt shares to.
+    pub fn public_key(&self) -> bindings::Point {
+        let public_key = ProjectivePoint::GENERATOR * *self.0;
+        marshal::solidity_point(&public_key)
+    }
+
+    /// Encrypts or decrypts a 32-byte secret-share `msg` against a peer's
+    /// encryption public key. See [`ecdh`].
+    pub fn ecdh(&self, public_key: &bindings::Point, msg: B256) -> Result<B256, Error> {
+        let public_key = marshal::frost_point(public_key)?;
+        let encrypted = ecdh(&self.0, &public_key, msg.0)?;
+        Ok(encrypted.into())
+    }
+}
+
+impl Drop for EncryptionKey {
+    fn drop(&mut self) {
+        self.0.zeroize();
+    }
+}
+
+impl ZeroizeOnDrop for EncryptionKey {}
+
+/// Encrypts or decrypts `msg` via ECDH: `msg XOR (receiver_pubkey * sender_privkey).x`.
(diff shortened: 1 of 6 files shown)


crates/validator/src/frost/ecdh.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+impl Drop for EncryptionKey {
+    fn drop(&mut self) {
+        self.0.zeroize();
+    }
+}
+
+impl ZeroizeOnDrop for EncryptionKey {}
```
rmeissner, line 60: question: isn't this the same as the Drop implementation above?

crates/validator/src/frost/ecdh.rs
```diff
@@ -0,0 +1,146 @@
...
+
+    let shared_secret = (*receiver_pubkey * **sender_privkey).to_affine().x();
+    let mut result = msg;
+    for (byte, secret) in result.iter_mut().zip(shared_secret) {
+        *byte ^= secret;
+    }
+    Ok(result)
+}
+
+fn hash_to_scalar(domain: &[u8], msg: &[u8]) -> NonZeroScalar {
```
rmeissner: nit: shouldn't `domain` be `discriminant`?

crates/validator/src/frost/marshal.rs
```diff
@@ -0,0 +1,74 @@
...
+pub fn solidity_point(point: &k256::ProjectivePoint) -> bindings::Point {
+    if point.is_identity().into() {
+        return bindings::Point {
+            x: U256::ZERO,
+            y: U256::ZERO,
+        };
+    }
+
+    let encoded = point.to_encoded_point(false);
+    let (chunks, _) = encoded.as_bytes()[1..].as_chunks::<32>();
```
rmeissner, line 24: wow, the size of the chunks is a generic param 🤯

crates/validator/src/frost/marshal.rs
```diff
@@ -0,0 +1,74 @@
...
+    bindings::Point {
+        x: U256::from_be_bytes(chunks[0]),
+        y: U256::from_be_bytes(chunks[1]),
+    }
+}
+
+/// Converts a secp256k1 scalar to a `U256`.
+pub fn solidity_scalar(scalar: &k256::Scalar) -> U256 {
```
rmeissner, line 32: question: could these be defined as "extension" by passing the scalar as a "self", so that you can call it as `scalar.solidity_scalar()`?

crates/validator/src/frost/marshal.rs
```diff
@@ -0,0 +1,74 @@
...
+        .ok_or_else(Error::malformed_element)
+}
+
+/// Converts an ABI Schnorr signature `(Point r, uint256 z)` to a FROST signature.
+pub fn frost_signature(r: &bindings::Point, z: &U256) -> Result<frost_secp256k1::Signature, Error> {
+    Ok(frost_secp256k1::Signature::new(
+        frost_point(r)?,
+        frost_scalar(z)?,
+    ))
+}
```
rmeissner, line 74: uber nit: No tests for these? (maybe more my old habbit 🙈 )

rmeissner (approved):

crates/validator/src/frost/marshal.rs
rmeissner: also why not use Into?

crates/validator/src/frost/ecdh.rs
nlordell: No, this is a marker trait for the `Zeroize` machinery. Not strictly necessary, but it can matter when we interact with other code that relies on this marker trait.

crates/validator/src/frost/ecdh.rs
nlordell: Ah right, technically the `domain = "FROST-secp256k1-SHA256-v1" + discriminant`. Good catch.

crates/validator/src/frost/marshal.rs
nlordell: Its so cool - it makes it so the type of `chunks` is `&[[u8; 32]]`, so a slice of 32-byte arrays. This matters for writing optimized code (the compiler can skip bounds checks when it can prove that indexing into the array is done within its size type).

Rust is cool :sunglasses:.

crates/validator/src/frost/marshal.rs
nlordell: > could these be defined as "extension"

Yes, but it would make the code much more verbose. If these were to be used outside of the `frost` module, then I see an argument for it (to make it more convenient for module users), but the module is designed so all marshaling is internal to the `frost` module, so the extra ceremony is not worth it.

> also why not use Into?

To implement `From`/`Into` (note that you typically implement `From`), either one of the types (`From`, `k256::Scalar`, or `U256`) need to be from this crate, but none of them are, so it would not work.

crates/validator/src/frost/marshal.rs
nlordell: All marshaling code should be covered by tests higher up:
image

nlordell: The code is so simple (just marshaling types), I don't think any tests here are worth it really.

PR #524, opened 2026-07-03
[B 2]: Implement Key Generation

nlordell:
This PR adds the logic require for performing DKG in a way that is compatible with the TypeScript implementation.

### Design and Tradeoffs

It wraps the `frost-secp256k1` crate in a pretty straight forward way and adds some basic marshalling methods that are required for going to and from the Solidity representation of the FROST primitives.

Note that I made some small changes to the ECDH implementation, and made it be completely internal to the `frost` module; this is because I do not want external callers to be performing low-level cryptography.

---

## CLA signature

### DIFF
--- a/crates/validator/src/frost/ecdh.rs
+++ b/crates/validator/src/frost/ecdh.rs
@@ -1,10 +1,7 @@
 //! ECDH-XOR encryption of FROST secret shares for the onchain publishing
 //! channel.
 
-use crate::bindings;
-
-use super::{error::Error, marshal};
-use alloy::primitives::B256;
+use super::error::Error;
 use k256::{
     NonZeroScalar, ProjectivePoint, Scalar,
     elliptic_curve::{
@@ -37,17 +34,19 @@ impl EncryptionKey {
     }
 
     /// The public key `q` published onchain for peers to encrypt shares to.
-    pub fn public_key(&self) -> bindings::Point {
-        let public_key = ProjectivePoint::GENERATOR * *self.0;
-        marshal::solidity_point(&public_key)
+    pub(super) fn public_key(&self) -> ProjectivePoint {
+        ProjectivePoint::GENERATOR * *self.0
     }
 
     /// Encrypts or decrypts a 32-byte secret-share `msg` against a peer's
     /// encryption public key. See [`ecdh`].
-    pub fn ecdh(&self, public_key: &bindings::Point, msg: B256) -> Result<B256, Error> {
-        let public_key = marshal::frost_point(public_key)?;
-        let encrypted = ecdh(&self.0, &public_key, msg.0)?;
-        Ok(encrypted.into())
+    pub(super) fn ecdh(
+        &self,
+        public_key: &ProjectivePoint,
+        msg: [u8; 32],
+    ) -> Result<[u8; 32], Error> {
+        let encrypted = ecdh(&self.0, public_key, msg)?;
+        Ok(encrypted)
     }
 }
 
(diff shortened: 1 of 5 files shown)


crates/validator/src/frost/ecdh.rs
```diff
@@ -141,6 +136,6 @@ mod tests {
     #[test]
     fn ecdh_rejects_degenerate_inputs() {
         let alice = key(2);
-        assert!(alice.ecdh(&bindings::Point::default(), B256::ZERO).is_err());
+        assert!(alice.ecdh(&ProjectivePoint::IDENTITY, [0xff; 32]).is_err());
```
rmeissner, line 139: nit: [0xff; 32] is not B256::ZERO, right? So technically you are changing the test :P

crates/validator/src/frost/ecdh.rs
```diff
@@ -119,7 +114,7 @@ mod tests {
     fn ecdh_is_commutative() {
         let alice = key(2);
         let bob = key(3);
-        let msg = m([0x42; 32]);
+        let msg = [0x00; 32];
```
rmeissner, line 117: question: why did you adjust the message to `[0x00; 32]` from `[0x42; 32]`

crates/validator/src/frost/keygen.rs
```diff
@@ -0,0 +1,175 @@
...
+    pub commitment: bindings::KeyGenCommitment,
+}
+
+/// Runs DKG round 1 for `me`, producing the round 1 secret package and the
+/// onchain [`KeyGenCommitment`](bindings::KeyGenCommitment). The `encryption_key`
+/// and `rng` are supplied (and the secrets persisted) by the caller.
+pub fn generate_round1<R>(
+    rng: &mut R,
+    me: Address,
+    max_signers: u16,
```
rmeissner, line 31: nit: personally I would prefer if we stick to our naming for these counts. 

max_signers -> count (or signer_count)
min_signers -> threshold

I assume you switched to a different wording as this is the one from the frost crate?

crates/validator/src/frost/mod.rs
```diff
@@ -10,4 +10,79 @@
...
+mod participants;
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloy::primitives::address;
+    use std::collections::BTreeMap;
+
+    #[test]
+    fn ceremony() {
```
rmeissner, line 24: ❤️

crates/validator/src/frost/keygen.rs
```diff
@@ -0,0 +1,175 @@
...
+/// finalizing the group key material.
+pub fn generate_round3(
+    me: Address,
+    encryption_key: &EncryptionKey,
+    secret_package: &round2::SecretPackage,
+    commitments: &BTreeMap<Address, bindings::KeyGenCommitment>,
+    shares: &BTreeMap<Address, bindings::KeyGenSecretShare>,
+) -> Result<Round3, Error> {
+    let (encryption_keys, round1_packages) = decode_peer_commitments(me, commitments)?;
+    let round2_packages = shares
```
rmeissner, line 117: nit: for this nested loop it would be nice to have some code comments imo

rmeissner (approved):
took some more time to review as the logic is slightly different how we currently do it in the TS version. From the Rust code it looks like that first all peerShares from every participant are collected and only then they are handled. In the TS version the peerShares from each participant are immediately handled/ verified (in `handleKeygenSecrets`).

crates/validator/src/frost/ecdh.rs
nlordell: Yes, when I read the test I thought it wasn't clear whether the `IDENTITY` or the 0-hash was the degenerate input.

crates/validator/src/frost/ecdh.rs
nlordell: To assert that the 0-hash is an allowed ECDH input. It wasn't clear before from the tests.

crates/validator/src/frost/keygen.rs
nlordell: Correct, I can change.

nlordell:
> From the Rust code it looks like that first all peerShares from every participant are collected and only then they are handled

I noticed this too when I was thinking "how do you assign culpability". I addressed this in the `B3` phase stack.

crates/validator/src/frost/keygen.rs
nlordell: Fixed

crates/validator/src/frost/keygen.rs
nlordell: This will change significantly in the later commits, I will add comments there.

PR #527, opened 2026-07-04
[B 3ii]: Verified Commitments and Shares

nlordell:
... using the power of types!

### Context and Motivation

As a follow up to #526, where we want to start assigning attribution to cryptographic errors (in order to react to malicious validators and kick them out of consensus). As a next step, we define new `Round1Commitment` and `Round2Share` types which represent _validated_ commitments and shares from other participants.

### Decisions and Tradeoffs

When a Solidity `KeyGenCommitment` or `KeyGenShare` is received from onchain, we do additional validation in order to construct verified types (in the spirit of [_Parse, don't Validate_](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)). This allows us to, with more certainty, know that the DKG rounds will not fail when we do them (unless the caller did not correctly use the inputs).

This does have the downside of doing additional work (as the `frost_secp256k1::keys::dkg` will **also** do this work internally), but it is important in order be able to reliably determine culpability in face of DKG protocol failures.

---

## CLA signature

### DIFF
--- a/crates/validator/src/frost/keygen.rs
+++ b/crates/validator/src/frost/keygen.rs
@@ -3,15 +3,15 @@
 
 use super::{ecdh::EncryptionKey, error::Error, marshal, participants};
 use crate::{bindings, frost::ecdh::EncryptionPublicKey};
-use alloy::primitives::Address;
+use alloy::primitives::{Address, U256};
 use frost_secp256k1::{
     Identifier,
     keys::{
-        self, PublicKeyPackage,
+        self,
         dkg::{self, round1, round2},
     },
 };
-use k256::elliptic_curve::PrimeField as _;
+use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 
 /// The output of DKG round 1: the secret polynomial (persisted to the secret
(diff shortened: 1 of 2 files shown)


crates/validator/src/frost/keygen.rs
```diff
@@ -45,131 +45,207 @@ where
...
+        .filter(|address| **address != peer)
+        .position(|address| participants::identifier(*address) == *secret_package.identifier())
+        .ok_or(frost_secp256k1::Error::UnknownIdentifier)?;
+    let encrypted_share = share
+        .f
+        .get(index)
+        .ok_or(frost_core::Error::IncorrectNumberOfShares)?;
+    let secret_share = encryption_key.ecdh(encryption_public_key, encrypted_share.to_be_bytes());
+
+    let y = marshal::frost_point(&share.y)?;
```
rmeissner, line 190: nit: why only verify this here and not directly after getting the verifying share?

crates/validator/src/frost/keygen.rs
```diff
@@ -45,131 +45,207 @@ where
...
-            )))
-            .collect(),
-    )?;
+    let round1_peer_packages =
+        peer_packages(secret_package.identifier(), commitments, |commitment| {
+            commitment.package.clone()
+        });
+
+    let (secret_package, round2_packages) =
+        dkg::part2(secret_package.clone(), &round1_peer_packages)?;
```
rmeissner, line 102: nit to the frost library (so out of scope for this): The naming of part2 is super hard to understand if you are not familiar with the frost related dkg ... I now have to look up what part 2 was ... super hard to review.

So if I looked this up correctly it evaluated the polynomial for each of the participants. The (round 1) secret package contains the [coefficients](https://docs.rs/frost-secp256k1/latest/frost_secp256k1/keys/dkg/round1/type.SecretPackage.html) for this.

Learning: before reviewing any of this we should have done a short recap on how the frost rust library works and what the type are. https://docs.rs/frost-secp256k1/ is gold (and required for review)

crates/validator/src/frost/keygen.rs
```diff
@@ -45,131 +45,207 @@ where
...
+        })
+        .filter(|(participant, _)| participant != secret_package.identifier())
+        .map(|(peer, encryption_public_key)| {
             let package = round2_packages
-                .get(&participants::identifier(*peer))
+                .get(&peer)
                 .ok_or(frost_secp256k1::Error::UnknownIdentifier)?;
             let signing_share = package.signing_share().to_scalar().to_bytes().into();
-            Ok(encryption_key.ecdh(encryption_key_peer, signing_share))
+            Ok(encryption_key.ecdh(encryption_public_key, signing_share))
```
rmeissner, line 132: uber nit: preferred the previous name

crates/validator/src/frost/keygen.rs
```diff
@@ -45,131 +45,207 @@ where
...
     Ok(Round2 {
         secret_package,
+        public_key_package,
         share,
     })
 }
 
+/// A validated round 2 signing share from a peer.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct Round2Share {
```
rmeissner, line 146: uber nit: this naming kills me. The comment on the struct is super valuable. I was writing a longer text, but I think a PR review is not the right place. It is not blocking so lets talk about it in person.

rmeissner (approved):

crates/validator/src/frost/keygen.rs
nlordell: I will address this naming in a follow up.

crates/validator/src/frost/keygen.rs
nlordell: Will address this naming disparity in a follow up.

crates/validator/src/frost/keygen.rs
nlordell: This will move in the next PR and make more sense why it is at the end (it is grouped with the things where there is a culprit).

crates/validator/src/frost/keygen.rs
nlordell: I am  not a fan of the previous name - it was confusing with "is the peers encryption key, or public key?" Adding an additional `_peer_` to the name made it way too verbose IMO.

PR #528, opened 2026-07-04
[B 3iii]: Error With Culpability

nlordell:
### Context and Motivation

The third and final step for assigning culpability to FROST operations: we change the error type to clearly differentiate between an unexpected cryptographic FROST error, and one that was caused by a misbehaving participant.

This is important for consensus to function correctly and remove misbehaving peers from the set of participants in order to continue with consensus.

### Decisions and Tradeoffs

We changed the internals to all produce `frost_secp256k1::Error`s, with **no manual conversion** to the `Error` type. This forces code to explicitly consider what type of error it is and assign culpability where appropriate.

We will be able to reuse this error type for the signing ceremony as well.

---

## CLA signature

### DIFF
--- a/crates/validator/src/frost/error.rs
+++ b/crates/validator/src/frost/error.rs
@@ -1,23 +1,46 @@
+use alloy::primitives::Address;
+
 /// A Safenet FROST error.
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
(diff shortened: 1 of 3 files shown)


rmeissner (approved):
Probably worth looking at the Gemini comments.

PR #529, opened 2026-07-06
[B 0]: Implement Hmac on Signer

rmeissner:
Allow to generate deterministic hashes that are derived from the signer private key. This uses hmac and sha2 to follow a well defined standard.

## CLA signature

### DIFF
--- a/crates/core/src/tx/signer.rs
+++ b/crates/core/src/tx/signer.rs
@@ -1,5 +1,6 @@
 //! The local account used to sign transactions for submitting onchain.
 
+use crate::kdf;
 use alloy::{
     consensus::{SignableTransaction as _, TxEip1559},
     eips::Encodable2718 as _,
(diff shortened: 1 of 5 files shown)


crates/core/src/tx/signer.rs
```diff
@@ -46,6 +52,19 @@ impl Signer {
...
         Ok(SignedTransaction(raw_tx))
     }
+
+    /// Deterministically derives a 32-byte value from this account's private key and an
+    /// arbitrary `message`, using HMAC-SHA256. [`HMAC_DOMAIN_PREFIX`] is always folded in ahead
+    /// of `message`, so callers don't need to (and can't forget to) domain-separate themselves.
+    ///
+    /// Since the output is keyed by the account's own private key, it is
+    /// reproducible without persisting anything beyond `message` itself.
+    pub fn hmac(&self, message: &[u8]) -> B256 {
```
nlordell: request change: This is not an HMAC, but an HKDF. We should use the https://docs.rs/hkdf/latest/hkdf/.

crates/core/src/tx/signer.rs
```diff
@@ -25,6 +26,11 @@ pub struct Signer(PrivateKeySigner);
 /// A raw signed transaction.
 pub struct SignedTransaction(Vec<u8>);
 
+/// Domain-separation prefix folded into every [`Signer::hmac`] derivation, keeping this key's
+/// HMAC derivations distinct from any other use of the same private key. Enforced here rather
+/// than left to call sites, so it can't be accidentally omitted by a future caller.
+const HMAC_DOMAIN_PREFIX: &[u8] = b"safenet-sentinel-reveal-salt";
```
nlordell: request change: This is specific to the sentinel, and should live in the sentinel crate (other wise, we may run into domain collisions across crates) and should be passed into the HKDF. We discussed we can assert the domain is not empty.

nlordell (approved):
LGTM

crates/core/src/tx/signer.rs
```diff
@@ -46,6 +47,22 @@ impl Signer {
...
+    /// `domain` must be a non-empty, caller-chosen constant identifying the specific use case
+    /// (e.g. `"safenet-sentinel-reveal-salt"`), so that derivations for unrelated purposes over
+    /// the same private key can never collide. See [`kdf::derive_key`] for details.
+    ///
+    /// Since the output is keyed by the account's own private key, it is
+    /// reproducible without persisting anything beyond `domain` and `message`.
+    pub fn derive_key(&self, domain: &[u8], message: &[u8]) -> B256 {
+        let mut key = self.0.to_bytes();
+        let derived = kdf::derive_key(key.as_slice(), domain, &[message]);
+        key.0.zeroize();
```
nlordell, line 63: Nice! I would have forgotten about this.

PR #530, opened 2026-07-06
[A 1]: Implement Commit Reveal Lib

rmeissner:
Adjust the commitments Solidity library to handle the commit/ reveal flow.

- Design Decisions:
  - Verify the hash in the reveal method. This requires to provide the salt to the lib.
  - The code does not include tests, these will be added once the implementation is stable
  - To avoid breaking tests the changes are implemented in a copied library. Once the implementation is completed the original will be removed.

## CLA signature

### DIFF
--- a/contracts/src/libraries/SentinelOracleCommitmentsV2.sol
+++ b/contracts/src/libraries/SentinelOracleCommitmentsV2.sol
@@ -1,0 +1,116 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+library SentinelOracleCommitment {
+    // ============================================================
+    // ENUMS
+    // ============================================================
+
+    enum Vote {
+        NONE, // never committed
+        PENDING, // committed, not yet revealed
+        APPROVED,
+        DENIED
+    }
+
+    // ============================================================
+    // STRUCTS
+    // ============================================================
+
+    struct Commitment {
+        bytes32 commitHash;
+        uint256 bondAmount;
+        Vote vote;
+        bool claimed;
+    }
+
+    // ============================================================
+    // ERRORS
+    // ============================================================
+
+    error NothingToClaim();
+    error AlreadyClaimed();
+    error NotRevealed();
+
+    // ============================================================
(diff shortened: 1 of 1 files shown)


contracts/src/libraries/SentinelOracleCommitments.sol
```diff
@@ -19,18 +29,28 @@ library SentinelOracleCommitment {
...
         return snapshot;
     }
+
+    function computeHash(address sentinel, bytes32 requestId, bool approve, bytes32 salt)
+        internal
+        pure
+        returns (bytes32)
+    {
+        return keccak256(abi.encodePacked(approve, salt, sentinel, requestId));
+    }
```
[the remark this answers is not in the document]
rmeissner: Specified differently in the epic. Can be optimized later once the contracts are finalized (as part of gas optimization)

contracts/src/libraries/SentinelOracleCommitmentsV2.sol
```diff
@@ -0,0 +1,116 @@
...
+
+    // ============================================================
+    // INTERNAL FUNCTIONS
+    // ============================================================
+
+    function checkNotCommitted(T storage self, bytes32 requestId, address sentinel) internal view {
+        require(self.commitments[requestId][sentinel].commitHash == 0, AlreadyCommitted());
+    }
+
+    function add(T storage self, bytes32 requestId, address sentinel, bytes32 commitHash, uint256 bondAmount) internal {
```
remedcu, line 91: nit: why `add` and not `commit`?

```suggestion
    function commit(T storage self, bytes32 requestId, address sentinel, bytes32 commitHash, uint256 bondAmount) internal {
```

remedcu (approved):
Small nit, rest LGTM 👍🏾

contracts/src/libraries/SentinelOracleCommitmentsV2.sol
rmeissner: I would say because it is a commitment, so committing a commitment feels duplicated. But no strong opinion.

PR #531, opened 2026-07-07
[A 2]: Adjust Request Lib

rmeissner:
Implement commit reveal logic in the request lib.

### Design Decisions
- Reward calculation has been simplified. Every winning sentinel gets an equal cut.
- Early finalization is possible if everyone revealed their vote.
- Adjustments have been made in a separate file to avoid breaking current tests. Old versions will be removed on completion.

### Review
- Compare SentinelOracleRequests libs: `diff -u contracts/src/libraries/SentinelOracleRequests.sol contracts/src/libraries/SentinelOracleRequestsV2.sol`

---

## CLA signature

### DIFF
--- a/contracts/src/libraries/SentinelOracleRequestsV2.sol
+++ b/contracts/src/libraries/SentinelOracleRequestsV2.sol
@@ -1,0 +1,211 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+library SentinelOracleRequest {
+    // ============================================================
+    // ENUMS
+    // ============================================================
+
+    enum State {
+        PENDING,
+        FROZEN,
+        RESOLVED_APPROVED,
+        RESOLVED_DENIED,
+        TIMED_OUT
+    }
+
+    enum ResolveReason {
+        UNANIMOUS_APPROVE,
+        UNANIMOUS_DENY,
+        TIMEOUT,
+        ARBITRATION
+    }
+
+    // ============================================================
+    // STRUCTS
+    // ============================================================
+
+    struct Request {
+        address proposer;
+        uint256 fee;
+        uint256 bondTarget;
+        uint256 commitDeadline;
+        uint256 revealDeadline;
+        State state;
+        uint256 committedCount;
+        uint256 revealedCount;
+        uint256 approveSentinelCount;
+        uint256 denySentinelCount;
+    }
+
+    // ============================================================
+    // ERRORS
+    // ============================================================
+
+    error RequestNotPending();
+    error RequestNotFrozen();
+    error RequestNotResolved();
+    error CommitWindowClosed();
+    error RevealWindowNotOpen();
+    error RevealWindowClosed();
+    error FinalizeTooEarly();
+
+    // ============================================================
+    // INTERNAL FUNCTIONS
+    // ============================================================
+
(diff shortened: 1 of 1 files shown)


contracts/src/libraries/SentinelOracleRequestsV2.sol
```diff
@@ -0,0 +1,211 @@
...
+        UNANIMOUS_DENY,
+        TIMEOUT,
+        ARBITRATION
+    }
+
+    // ============================================================
+    // STRUCTS
+    // ============================================================
+
+    struct Request {
```
remedcu, line 28: Could this struct be further optimized at this stage? Or is it planned for later? Example having `state` and `proposer` back to back for 1 slot. Similarly, do we need `uint256` for each of the deadlines? Do we expect more than 2**64 sentinels, if no, then we can have a single struct with counts.

contracts/src/libraries/SentinelOracleRequestsV2.sol
```diff
@@ -0,0 +1,211 @@
...
+            return (State.RESOLVED_APPROVED, 0, unrevealedBond);
+        }
+        if (denyMet) {
+            self.state = State.RESOLVED_DENIED;
+            return (State.RESOLVED_DENIED, 0, unrevealedBond);
+        }
+        self.state = State.TIMED_OUT;
+        refundFee = self.fee;
+        self.fee = 0;
+        return (State.TIMED_OUT, refundFee, unrevealedBond);
```
remedcu, line 107: nit: explicit return not required.

```suggestion
        self.fee = 0;
        self.state = State.TIMED_OUT;
        newState = State.TIMED_OUT;
        refundFee = self.fee;
```

contracts/src/libraries/SentinelOracleRequestsV2.sol
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    function requireResolved(Request storage self) internal view returns (State) {
+        State state = self.state;
+        require(
+            state == State.RESOLVED_APPROVED || state == State.RESOLVED_DENIED || state == State.TIMED_OUT,
+            RequestNotResolved()
+        );
```
remedcu, line 115: nit: One less check in `TIMED_OUT` case


```suggestion
        require(state != State.PENDING && state != State.FROZEN, RequestNotResolved());
```

remedcu (approved):
Approving to avoid back and forth. Only few comments, rest LGTM 👍🏾

contracts/src/libraries/SentinelOracleRequestsV2.sol
rmeissner: Gas optimizations happen once we aim for a public release, so not as part of this epic.

contracts/src/libraries/SentinelOracleRequestsV2.sol
rmeissner: I would leave this out for now, as the original contract worked this way and I want to keep the diff small. I do think we need to review and refactor the whole contract before release in any case

contracts/src/libraries/SentinelOracleRequestsV2.sol
rmeissner: Also here this is the original contract code, would move optimizations to a later stage

PR #534, opened 2026-07-08
[B 4]: Implement Signing Nonces Chunking

nlordell:
### Context and Motivation

This PR implements the 1024 nonces chunks which need to be committed to (in the `FROSTCoordinator.preprocess` phase).

### Decisions and Tradeoffs

We use `rayon` in order to compute the chunk in parallel on all cores. This, along side the nonces pair specific seeded RNG, gives us performance gains linear to the number of cores on the system (on my system with 8 cores, it takes ~290ms instead of 2.4s; in `release` builds, nonce generation takes less than 20ms).

Note that we also chose to store proofs with each of the nonces - this allows the `Nonces` to be written to an SQLite database individually, and have the proof be read per nonce instead of requiring the whole tree to be loaded into memory.

### Testing

For now, we test on some internal details of the `Nonces` type (in particular, we need to compute the leaf hash for the nonce pair). I decided to keep it this way, as the validator never needs to verify nonce inclusion proofs (so there is no public API to use for testing).

---

## CLA signature

### DIFF
--- a/crates/validator/src/frost/nonces.rs
+++ b/crates/validator/src/frost/nonces.rs
@@ -1,0 +1,132 @@
+//! The 1024-nonce preprocessing scheme layered over standard FROST nonces.
+//!
+//! Each nonces' merkle proof is **precomputed at generation time** (when the
+//! tree is built once anyway) and stored alongside it in a [`Nonces`], so
+//! revealing a single nonce needs only that one entry rather than the whole
+//! chunk resident in memory.
+
+use super::marshal;
+use crate::{
+    bindings,
+    merkle::{MerkleRoot, MerkleTree},
+};
+use alloy::primitives::{B256, keccak256};
+use frost_secp256k1::{keys::SigningShare, round1};
+use rand::{CryptoRng, RngCore, SeedableRng as _};
+use rand_chacha::ChaCha12Rng;
+use rayon::iter::{IntoParallelIterator as _, ParallelIterator as _};
+
+/// The number of nonces committed to per `preprocess`, matching the onchain
+/// `FROSTNonceCommitmentSet` sequence chunk size.
+pub const SEQUENCE_CHUNK_SIZE: u64 = 1024;
+
+/// A single preprocessing nonce pair and its precomputed inclusion proof in the
+/// chunk's merkle tree. Because the proof is stored with the nonce, a nonce can
+/// be revealed in isolation without the rest of the chunk.
+pub struct Nonces {
+    signing_nonces: round1::SigningNonces,
+    proof: Vec<B256>,
+}
+
+impl Nonces {
+    /// Compute the signature nonces reveal parameters the FROST sining nonce
+    /// pair.
+    pub fn reveal(&self) -> (bindings::SignNonces, &[B256]) {
+        (
+            marshal::solidity_sign_nonces(self.signing_nonces.commitments()),
+            &self.proof,
+        )
+    }
+}
+
+/// A freshly generated chunk of preprocessing nonces: the per-offset secrets
+/// (each carrying its own proof) to persist and the merkle `commitment` to
+/// publish onchain.
+pub struct NonceChunk {
+    /// The `SEQUENCE_CHUNK_SIZE` nonces, indexed by offset. Secret; persisted to
+    /// the secret store and never rolled back.
+    pub nonces: Vec<Nonces>,
+    /// The merkle root over the nonce commitments, published by `preprocess`.
+    pub commitment: MerkleRoot,
+}
+
+impl NonceChunk {
+    /// Generates a fresh chunk of `SEQUENCE_CHUNK_SIZE` nonces for
+    /// `signing_share`, the merkle root committing to them, and each nonce's
+    /// inclusion proof. The RNG is supplied by the caller.
+    pub fn generate<R>(signing_share: &SigningShare, rng: &mut R) -> Result<NonceChunk, rand::Error>
+    where
+        R: RngCore + CryptoRng,
+    {
+        // Parallelize nonce generation to speed up the process. Note that this
+        // requires us to seed one RNG per nonce pair, as the `R` passed in
+        // cannot be shared across threads. The choice of [`ChaCha12Rng`] is
+        // based on the fact that it is the standard RNG used by [`rand`] (both
+        // the `ThreadRng` and `StdRng`), it is a cryptographically secure RNG,
+        // and it allows unique random streams per nonce generation.
+        let rngs = (0..SEQUENCE_CHUNK_SIZE)
+            .map({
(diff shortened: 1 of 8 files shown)


crates/validator/src/frost/nonces.rs
```diff
@@ -0,0 +1,131 @@
...
+                let mut rng = ChaCha12Rng::from_rng(&mut *rng)?;
+                rng.set_stream(offset.wrapping_add(1));
+                Ok((offset, rng))
+            })
+            .collect::<Result<Vec<_>, rand::Error>>()?;
+        let (signing_nonces, leaves) = rngs
+            .into_par_iter()
+            .map(|(offset, mut rng)| {
+                let signing_nonces = round1::SigningNonces::new(signing_share, &mut rng);
+                let leaf = nonces_leaf(offset, signing_nonces.commitments());
```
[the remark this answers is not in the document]
nlordell: Nice suggestion!

crates/validator/src/frost/nonces.rs
```diff
@@ -0,0 +1,117 @@
...
+        // cannot be shared across threads. The choice of [`ChaCha12Rng`] is
+        // based on the fact that it is the standard RNG used by [`rand`] (both
+        // the `ThreadRng` and `StdRng`), it is a cryptographically secure RNG,
+        // and it allows unique random streams per nonce generation.
+        let rngs = (0..SEQUENCE_CHUNK_SIZE)
+            .map({
+                let seed = ChaCha12Rng::from_rng(&mut *rng)?;
+                move |offset| {
+                    let mut rng = seed.clone();
+                    rng.set_stream(offset.wrapping_add(1));
```
rmeissner: question: does the wrapping_add mean that theoretically at some point we get the original value again?

crates/validator/src/frost/nonces.rs
```diff
@@ -0,0 +1,117 @@
...
+        // Parallelize nonce generation to speed up the process. Note that this
+        // requires us to seed one RNG per nonce pair, as the `R` passed in
+        // cannot be shared across threads. The choice of [`ChaCha12Rng`] is
+        // based on the fact that it is the standard RNG used by [`rand`] (both
+        // the `ThreadRng` and `StdRng`), it is a cryptographically secure RNG,
+        // and it allows unique random streams per nonce generation.
+        let rngs = (0..SEQUENCE_CHUNK_SIZE)
+            .map({
+                let seed = ChaCha12Rng::from_rng(&mut *rng)?;
+                move |offset| {
```
rmeissner, line 70: question: what does move do? 😅

crates/validator/src/frost/nonces.rs
```diff
@@ -0,0 +1,117 @@
...
+            commitment: tree.root(),
+        })
+    }
+}
+
+/// The leaf hash for a nonce commitment at `offset` in the nonce tree. Defined
+/// as `keccak256(abi.encode(offset, d.x, d.y, e.x, e.y))`.
+pub fn nonces_leaf(offset: u64, commitments: &round1::SigningCommitments) -> B256 {
+    let mut buf = [0u8; 160];
+    buf[24..32].copy_from_slice(&offset.to_be_bytes());
```
rmeissner, line 108: question: the first 24 bytes are empty as we are encoding a u64 into a u256, right?

crates/validator/src/frost/nonces.rs
```diff
@@ -0,0 +1,132 @@
...
+        Ok(NonceChunk {
+            nonces,
+            commitment: tree.root(),
+        })
+    }
+}
+
+/// The leaf hash for a nonce commitment at `offset` in the nonce tree. Defined
+/// as `keccak256(abi.encode(offset, d.x, d.y, e.x, e.y))`.
+fn nonces_leaf(offset: u64, nonces: &bindings::SignNonces) -> B256 {
```
rmeissner, line 106: nit: maybe a test for this with data that is mirrored from the contracts part?

crates/validator/src/merkle.rs
```diff
@@ -0,0 +1,122 @@
...
+fn hash_pair(a: B256, b: B256) -> B256 {
+    let (left, right) = if a <= b { (a, b) } else { (b, a) };
+    let mut data = [0u8; 64];
+    data[..32].copy_from_slice(left.as_slice());
+    data[32..].copy_from_slice(right.as_slice());
+    keccak256(data)
+}
+
+#[cfg(test)]
+mod tests {
```
rmeissner, line 97: nit: could we mirror the tests from the TS validator here? https://github.com/safe-research/safenet/blob/main/validator/src/consensus/merkle.test.ts

rmeissner (approved):
uber nit: theoretically the Merkle Tree logic could have been a separate PR 😛

crates/validator/src/frost/nonces.rs
nlordell: That is a good point - this should be `checked_add(1).expect(...)`!

crates/validator/src/frost/nonces.rs
nlordell: This makes `seed` owned by the closure. This is only needed because I create a `seed` only in the block starting at line 54 (with the `{`) and without the move, seed would not live long enough to be referenced by the closure. I do this to make explicit that `seed` is only needed for the closure.

crates/validator/src/frost/nonces.rs
```diff
@@ -0,0 +1,132 @@
...
+use super::marshal;
+use crate::{
+    bindings,
+    merkle::{MerkleRoot, MerkleTree},
+};
+use alloy::primitives::{B256, keccak256};
+use frost_secp256k1::{keys::SigningShare, round1};
+use rand::{CryptoRng, RngCore, SeedableRng as _};
+use rand_chacha::ChaCha12Rng;
+use rayon::iter::{IntoParallelIterator as _, ParallelIterator as _};
```
rmeissner, line 17: question: what does this do?

nlordell: It pulls in the traits into the current scope.

In particular, imagine:

```rust
mod mod1 {
    pub trait Foo {
        fn foo(&self);
    }
  
    impl Foo for u32 {
        fn foo(&self) {
            println!("FOO {self}");
        }
    }
}

mod mod2 {
    pub fn do_some_bad_foo() {
        1337_u32.foo(); // error
    }
}

mod mod3 {
    use super::mod1::Foo;
    pub fn do_some_good_foo() {
        42_u32.foo();
    }
}
```

The `mod2` would fail to compile, but but 3 would not, as the trait `Foo` is in scope. Additionally, you can import a trait as `use ... as _` to say "add this trait to the current scope, but not its name" that means, I cannot reference `Foo`, but I can use its trait implementations.

---

In particular, these traits are used for parallel iteration for nonce generation (to take advantage of all cores to generate the nonces).

PR #535, opened 2026-07-08
[B 5]: Implement Signing Ceremony

nlordell:
### Context and Motivation

This PR adds nonce reveal verification and signature share computation for signing ceremonies. 

### Decisions and Tradeoffs

As with the `keygen` types, we progressively validate nonce reveals as they come in - this allows us to exclude invalid signers from the selection right away.

### Testing

There is notably no testing done on the signature generation logic in this PR. This will be handled in a follow up where we extend the `ceremony` test in the `frost` module to also sign a message.

---

## CLA signature

### DIFF
--- a/crates/validator/src/frost/signing.rs
+++ b/crates/validator/src/frost/signing.rs
@@ -1,0 +1,203 @@
+//! Signature-share creation for a signing ceremony.
+//!
+//! This module wraps the [`frost_secp256k1`] crate with the Safenet-specific
+//! signer-set selection.
+
+use super::{
+    error::{Culprit as _, Error},
+    marshal,
+    nonces::Nonces,
+    participants,
+};
+use crate::{bindings, frost::marshal::frost_signing_commitments, merkle::MerkleTree};
+use alloy::primitives::{Address, B256, U256, keccak256};
+use frost_secp256k1::{Secp256K1Sha256, SigningPackage, keys, round1, round2};
+use k256::elliptic_curve::PrimeField as _;
+use serde::{Deserialize, Serialize};
+use std::collections::BTreeMap;
+
+/// A validated revealed nonce commitment from a signer.
+///
+/// Revealing happens onchain (`signRevealNonces`), so a peer's `d`/`e` points
+/// have already been merkle-checked against their preprocessing commitment;
+/// this additionally attributes malformed or identity point to a misbehaving
+/// participant.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct RevealedNonces {
+    commitments: round1::SigningCommitments,
+}
+
+/// Verifies a signer's revealed nonce commitments.
+///
+/// This just verifies that the nonce commitments are valid values (i.e. they
+/// are non-identity points on the secp256k1 curve). We do not verify the
+/// inclusion proof that the nonces were part of a nonces chunk commitment
+/// submitted during the preprocessing phase (this is verified by the FROST
+/// coordinator contract onchain).
+pub fn verify_revealed_nonces(
+    participant: Address,
+    nonces: &bindings::SignNonces,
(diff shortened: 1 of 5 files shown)


crates/validator/src/frost/signing.rs
```diff
@@ -0,0 +1,170 @@
...
+            .expect("binding factor always serializes the correct number of bytes")
+            .into(),
+    )
+    .expect("binding factor is always a valid scalar")
+}
+
+/// The leaf hash for a participant in the signer-set selection tree. Defined as
+/// `keccak256(participant, r.x, r.y, l, group_r.x, group_r.y)`, matching
+/// `FROSTSignatureShares._hash` (the participant occupies a full 32-byte slot).
+fn signer_leaf(
```
nlordell, line 164: nit: Similar to https://github.com/safe-research/safenet/pull/534#discussion_r3542959859 - I should add a test from the contract parts.

crates/validator/src/frost/signing.rs
```diff
@@ -0,0 +1,170 @@
...
+/// Revealing happens onchain (`signRevealNonces`), so a peer's `d`/`e` points
+/// have already been merkle-checked against their preprocessing commitment;
+/// this additionally attributes malformed or identity point to a misbehaving
+/// participant.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct RevealedNonces {
+    commitments: round1::SigningCommitments,
+}
+
+/// Verifies a signer's revealed nonce commitment.
```
rmeissner: question: this is only question for correct unmarshaling of the nonce, right? is this something that still has to be implemented in this PR or will this happen later?

crates/validator/src/frost/signing.rs
```diff
@@ -0,0 +1,170 @@
...
+    // mixing revealed nonces, signing nonces or key packages from different
+    // signing ceremonies). There is, therefore, no culprit.
+    round2::sign(&signing_package, nonces.signing_nonces(), key_package)
+        .and_then(|signature_share| {
+            // Each signer's commitment share and Lagrange coefficient are
+            // public, so every participant reconstructs the identical signer
+            // selection, ordered by FROST identifier.
+            let binding_factors =
+                frost_core::compute_binding_factor_list(&signing_package, group_public_key, &[])?;
+            let signers = revealed
```
rmeissner, line 90: uber nit: I don't find the name `signers` represents what this is, but I have no better name in mind, therefore no action required.

crates/validator/src/frost/signing.rs
```diff
@@ -0,0 +1,170 @@
...
+
+            // Extract the signature share values that are needed onchain along
+            // with the selection inclusion Merkle proof.
+            let (_, r, l) = signers
+                .get(key_package.identifier())
+                .ok_or(frost_secp256k1::Error::UnknownIdentifier)?
+                .clone();
+            let z = marshal::solidity_signing_share(&signature_share);
+            let share = bindings::SignatureShare { r, z, l };
+            let index = signers.range(..key_package.identifier()).count();
```
rmeissner, line 134: nit/ question: is this really good code in Rust? We create a mapping (ordered) which we most of the time use a array, only case where it is used as a mapping is when we look up the "me" info. And then we do a range to get the index, which is super unintuitive for me

rmeissner: resolved in meeting: BTrees are sorted by identifier/ key. @nlordell add more context in code

crates/validator/src/frost/signing.rs
rmeissner: @nlordell add a comment that we trust the contract (as it verifies the merkle proof)

rmeissner (approved):

PR #538, opened 2026-07-09
[B 7/7]: Refactor Public FROST API

nlordell:
This PR refactors the public FROST API to make it harder to use incorrectly, make errors more consistent across different validators, and not leak any of the `frost_secp256k1` types in the public API.

This should address the naming complaints from the `keygen` module from previous PRs, and be the final piece in the validator's FROST library.

### Context and Motivation

There were a few rough edges to using the FROST API:

- The naming (round N is not very descriptive)
- Error proneness of use
- Inconsistent error states across different validators

While the nonces and signing API, this was already the case, this did cause some larger changes in the key generation API:

| Before | After |
| --- | --- |
| `generate_round1 -> Round1` | `setup -> Setup` |
| `verify_round1_commitment -> Round1Commitment` | `verify_commitment -> VerifiedCommitment` |
| `generate_round2 -> Round2` | `generate_secret_shares` |
| `verify_round2_share -> Round2Share` | `verify_secret_share -> VerifiedSecretShare` |
| `generate_round3 -> Round3` | `finalize -> KeyShare` |

### Decisions and Tradeoffs

We encapsulate all of the required forward-needed values into structs returned by each of the steps. This no only decreases the arity of the function, it also makes it harder to mix values from different DKG ceremonies to catch bugs easier.

Furthermore, we now check the `me`'s commitments and shares for correctness in the different DKG rounds (something that was not done before); this is to keep consistent error states across different validators (otherwise a `me` with a bad commitment would not be caught by `me` but tagged as misbehaving by all other validators). This was done to ensure consensus remains consistent under reorgs (although is more of a theoretical risk than a real one).

For more consistent naming with the contracts, we renamed `nonces` to `preprocess` and `signing` to `sign`.

Additional comments were added to the DKG process as requested in https://github.com/safe-research/safenet/pull/524#discussion_r3520326496.

---

## CLA signature

### DIFF
--- a/crates/validator/src/frost/keygen.rs
+++ b/crates/validator/src/frost/keygen.rs
@@ -18,23 +18,30 @@ use frost_secp256k1::{
 use serde::{Deserialize, Serialize};
 use std::collections::BTreeMap;
 
-/// The output of DKG round 1: the secret polynomial (persisted to the secret
-/// store) and the onchain commitment to publish.
-pub struct Round1 {
-    pub encryption_key: EncryptionKey,
-    pub secret_package: round1::SecretPackage,
+/// A participant's generated secrets created at [`setup`] and required by
+/// [`verify_commitment`] and [`generate_secret_shares`]. Persisted to the
+/// secret store between rounds.
+#[derive(Clone, Deserialize, Serialize)]
+pub struct Secrets {
+    encryption_key: EncryptionKey,
+    secret_package: round1::SecretPackage,
+}
+
+/// The result of [`setup`]: the secrets to persist across rounds and the
+/// onchain commitment to publish.
+///
+/// Note that the secrets are generated with randomness, meaning that they must
+/// be persisted in a reorg-resistant way.
+pub struct Setup {
+    pub secrets: Secrets,
     pub commitment: bindings::KeyGenCommitment,
 }
 
-/// Runs DKG round 1 for `me`, producing the round 1 secret package and the
-/// onchain [`KeyGenCommitment`](bindings::KeyGenCommitment). The `encryption_key`
-/// and `rng` are supplied (and the secrets persisted) by the caller.
-pub fn generate_round1<R>(
-    rng: &mut R,
-    me: Address,
-    count: u16,
-    threshold: u16,
-) -> Result<Round1, Error>
+/// Sets up key generation for `me`: samples the ECDH encryption key and the
+/// secret polynomial (to persist to the secret store) alongside the onchain
+/// [`KeyGenCommitment`](bindings::KeyGenCommitment) to publish. The `rng` is
+/// supplied by the caller.
+pub fn setup<R>(rng: &mut R, me: Address, count: u16, threshold: u16) -> Result<Setup, Error>
 where
     R: rand::RngCore + rand::CryptoRng,
 {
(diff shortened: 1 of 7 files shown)


nlordell:
Note to reviewers: this is a slightly larger PR (the thread I pulled ended up being a little longer than anticipated). If this is too hard to review - let me know and I can try to break it up into more manageable chunks.

crates/validator/src/frost/keygen.rs
```diff
@@ -43,76 +55,102 @@ where
...
+/// Unlike [`Secrets`], this can be reconstructed using the generated secrets
+/// from [`setup`] and onchain state, and therefore can be stored in the state
+/// machine (and does not require a reorg-resistant storage).
+#[derive(Clone, Deserialize, Serialize)]
+pub struct SharingState {
+    encryption_key: EncryptionKey,
+    secret_package: round2::SecretPackage,
+    verifying_share: keys::VerifyingShare,
+    commitments: BTreeMap<Address, VerifiedCommitment>,
+}
```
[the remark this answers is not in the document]
nlordell: This is a good suggestion, will add it back in.

I had originally removed the `PublicKeyPackage` from the parameters as a simplification, but since we now carry the state around in a single struct, it doesn't make sense not to keep it around.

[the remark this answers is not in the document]
rmeissner: This change was not implemented, right? Asking because the thread was resolved ;)

crates/validator/src/frost/keygen.rs
```diff
@@ -142,142 +183,194 @@ pub fn generate_round2(
...
+        |commitment| commitment.package.clone(),
+    )?;
+    let (round2_peer_packages, round2_me_package) =
+        peer_packages(secret_state.secret_package.identifier(), &shares, |share| {
+            share.package.clone()
+        })?;
+
+    // Verify that the me package matches the secret state. This is otherwise
+    // not checked by the FROST library.
+    if round2_me_package.signing_share().to_scalar() != secret_state.secret_package.secret_share() {
```
rmeissner: question: why is this important? before this was not checked, right?

crates/validator/src/frost/keygen.rs
```diff
@@ -43,76 +55,102 @@ where
...
     // culprits (as the `commitments` have already verified the length of the
     // commitments and the proof of knowledge). The only way we encounter an
     // error here is if the caller were to mix secrets and commitments from
     // different DKG ceremonies.
-    dkg::part2(secret_package.clone(), &round1_peer_packages)
+    dkg::part2(secrets.secret_package.clone(), &round1_peer_packages)
         .and_then(|(secret_package, round2_packages)| {
+            // Ensure that the `me` commitment is also valid, the FROST library
+            // doesn't check this by default.
+            if round1_me_package.commitment() != secret_package.commitment() {
```
rmeissner, line 143: question: Same here as below

crates/validator/src/frost/keygen.rs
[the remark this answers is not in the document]
nlordell: Fixed, just not pushed. Should be there now.

rmeissner (approved):

crates/validator/src/frost/keygen.rs
nlordell: From the PR description:

> Furthermore, we now check the me's commitments and shares for correctness in the different DKG rounds (something that was not done before); this is to keep consistent error states across different validators (otherwise a me with a bad commitment would not be caught by me but tagged as misbehaving by all other validators). This was done to ensure consensus remains consistent under reorgs (although is more of a theoretical risk than a real one).

So, not important, just defensive: helps catch bugs and ensures more consistent state across validators (especially WRT reorgs).

crates/validator/src/frost/keygen.rs
nlordell: See above.

PR #540, opened 2026-07-09
[B3a]: Implement action encoding and transition matcher

rmeissner:
The implementation follows the service.rs implementation for action encoding and state transition handling.

### Design Decisions
  - Compelte side by side with the original implementation to allow cross cheking
  - Split into multiple smaller tasks for reviewability (epic has been updated)

### Review:
- Check against https://github.com/safe-research/safenet/blob/c800c0c06416cda312c3401861bbc35f8692c26f/crates/sentinel/src/service.rs#L257-L353

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/servicev2.rs
+++ b/crates/sentinel/src/servicev2.rs
@@ -1,0 +1,294 @@
+use crate::{
+    action::{SentinelActionKindV2 as SentinelActionKind, SentinelActionV2 as SentinelAction},
+    bindings::{
+        SentinelEventsV2 as SentinelEvents,
+        consensus::Consensus,
+        oracle::{ERC20, SentinelOracleV2 as SentinelOracle},
+    },
+    detector::Detector,
+    state::StateV2 as State,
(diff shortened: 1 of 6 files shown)


crates/sentinel/src/servicev2.rs
```diff
@@ -0,0 +1,294 @@
...
+/// events.
+#[expect(dead_code)]
+pub struct SentinelTransition {
+    oracle: Address,
+    /// The `Consensus` contract whose `OracleTransactionProposed` events are
+    /// hashed into request ids.
+    consensus: Address,
+    /// Our own account, used to compute commitment hashes and identify votes
+    /// we committed onchain.
+    signer: Signer,
```
nlordell, line 46: At first, I though "why do you need a signer? you should only need the signer address" - then I remembered "BUT THE HMAC!?!?!?".

And you mention it in the comments, so its my fault for not reading.

nlordell (approved):

PR #541, opened 2026-07-09
[B3b]: Handle propose and new request

rmeissner:
Implement function to handle proposal of a new transaction and handling of a new request related to that proposal.

There are no tests added as these will be handled in a later phase.

### Review:
- Check against https://github.com/safe-research/safenet/blob/c800c0c06416cda312c3401861bbc35f8692c26f/crates/sentinel/src/service.rs#L89-L162

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/servicev2.rs
+++ b/crates/sentinel/src/servicev2.rs
@@ -6,7 +6,8 @@ use crate::{
         oracle::{ERC20, SentinelOracleV2 as SentinelOracle},
     },
     detector::Detector,
-    state::StateV2 as State,
+    hashing::{RevealSalt as _, commit_hash, oracle_tx_proposal_hash},
+    state::{SentinelRequestStateV2 as RequestState, StateV2 as State},
 };
 use alloy::{
     primitives::{Address, U256},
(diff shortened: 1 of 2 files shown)


crates/sentinel/src/servicev2.rs
```diff
@@ -87,27 +88,83 @@ impl SentinelService {
...
+                approve,
+                commit_deadline,
+                reveal_deadline,
+                committed_count: 0,
+                self_committed: false,
+            },
+        );
+        let actions = vec![
+            SentinelAction {
+                kind: SentinelActionKind::ApproveToken {
```
nlordell, line 153: I guess what is nice about our strict transaction ordering is that you _know_ you will only ever have `event.bondTarget` or `0` approval to the contract, so damage in case of a bad contract is fairly minimal.

crates/sentinel/src/servicev2.rs
```diff
@@ -87,27 +88,83 @@ impl SentinelService {
...
+        // after it has advanced further).
+        if state.0.contains_key(&request_id) {
+            return (state, Vec::new());
+        }
+        let approve = self.detector.approve(&event.transaction);
+        state.0.insert(
+            request_id,
+            RequestState::WaitingForRequest {
+                approve,
+                deadline: block.saturating_add(self.voting_window),
```
nlordell, line 119: question: What is the point of the deadline here? It isn't used anywhere as far as I can tell (in fact, this field is ignore in the `handle_new_request`) - will this be used for some timeout transition later on?

nlordell (approved):

crates/sentinel/src/servicev2.rs
rmeissner: Block advancements (and therefore deadlines) are handled in #545

PR #542, opened 2026-07-09
[B3c]: Handle Commitments and Votes

rmeissner:
Implements the collection of commitments and revealed votes. The sentinel keeps track of the counts of the revealed votes to determine if all votes have been revealed and if a dispute is happening. The commitment phase is only completed when the commitment window closes. This transition will be handled in a follow up PR when implementing block advancements.

No tests have been added as these will be handled in a later phase.

### Review:
- Previous implementation for reference (logic changed, so cannot be compared 1:1) https://github.com/safe-research/safenet/blob/c800c0c06416cda312c3401861bbc35f8692c26f/crates/sentinel/src/service.rs#L165-L181

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/servicev2.rs
+++ b/crates/sentinel/src/servicev2.rs
@@ -10,7 +10,7 @@ use crate::{
     state::{SentinelRequestStateV2 as RequestState, StateV2 as State},
 };
 use alloy::{
-    primitives::{Address, U256},
+    primitives::{Address, B256, U256},
     sol_types::SolCall,
 };
 use safenet_core::{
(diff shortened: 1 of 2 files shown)


crates/sentinel/src/servicev2.rs
```diff
@@ -110,26 +110,137 @@ impl SentinelTransition {
...
+            return (state, Vec::new());
+        };
+        let SentinelRequestStateV2::CollectingCommitments {
+            committed_count,
+            self_committed,
+            ..
+        } = existing
+        else {
+            return (state, Vec::new());
+        };
```
nlordell: comment: I don't know if you are aware, but you can combine these two:


```suggestion
        let Some(SentinelRequestStateV2::CollectingCommitments {
            committed_count,
            self_committed,
            ..
        }) = state.0.get_mut(&event.requestId) else {
            return (state, Vec::new());
        };
```

IMO - it does not sacrifice legibility.

crates/sentinel/src/servicev2.rs
```diff
@@ -110,26 +110,137 @@ impl SentinelTransition {
...
+        *committed_count += 1;
+        if event.sentinel == self.signer.address() {
+            *self_committed = true;
+        }
+        (state, Vec::new())
```
nlordell, line 188: Is this a terminal state? As in - it doesn't progress to the next state when the `committed_count` reaches some threshold? (or is that handled by a separate event)?

crates/sentinel/src/servicev2.rs
```diff
@@ -108,26 +108,123 @@ impl SentinelTransition {
...
+            *approve_count += 1;
+        } else {
+            *deny_count += 1;
+        }
+        if event.sentinel == self.signer.address() {
+            *self_revealed = true;
+        }
+        if *revealed_count < *committed_count {
+            return (state, Vec::new());
+        }
```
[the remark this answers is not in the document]
nlordell: This is a false positive right? You should only enter `CollectingCommitments` state if you are there at the beginning right?

crates/sentinel/src/servicev2.rs
```diff
@@ -108,26 +108,123 @@ impl SentinelTransition {
...
+        self.finalize(state, event.requestId)
+    }
+
+    /// Shared finalize step, reached from either the early-finalize check
+    /// in [`Self::handle_committed`]/[`Self::handle_revealed`] or the
```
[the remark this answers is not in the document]
nlordell: Left over from V1 right?

crates/sentinel/src/servicev2.rs
```diff
@@ -110,26 +110,137 @@ impl SentinelTransition {
...
+    /// `Finalize` is emitted even when we never revealed our own vote, but
+    /// only when nobody else revealed either (`revealed_count == 0`, a
+    /// genuine timeout) -- that's the one case where no other sentinel's
+    /// FSM has a revealed vote to finalize from either, so nobody else in
+    /// the fleet has a reason to call it. If anyone else did reveal, their
+    /// own FSM reaches this same step and finalizes; a non-revealer with
+    /// nothing to claim doesn't need to redundantly do so as well. `Claim`
+    /// stays conditional on our own reveal having landed, since a
+    /// `claim()` without a matching revealed commitment reverts with
+    /// `NotRevealed`.
```
nlordell: That sentence is hard to follow 😅

nlordell:
One thing that isn't clear to me is how it goes from the `committing` -> `votes` state. I would have (naively) imagined it happens once a certain `committed_count` is reached (or a deadline), but I am not sure (didn't see anything in the comments or in the PR description).

crates/sentinel/src/servicev2.rs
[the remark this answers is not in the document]
rmeissner: yes, this is a false possitive

crates/sentinel/src/servicev2.rs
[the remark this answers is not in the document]
rmeissner: yes, I will have to do a pass on all the comments ... Claude likes to add a lot.

crates/sentinel/src/servicev2.rs
rmeissner: there is no predefined committed count (at least in the current state). It will go into the next state once the deadline is reached (implemented in #545)

That being said, it is probably worth to evaluate (as a iteration)  if we can early finish the commitment phase if all sentinels committed. Will track this here: https://github.com/safe-research/safenet/issues/546

nlordell: This wasn't immediately clear from the PR - might be worth clarifying somewhere (I haven't thought enough about the sentinel game to have the state machine clearly mapped out in my head).

nlordell (approved):

PR #543, opened 2026-07-09
[C 1i]: Implement KeyGen Secret Store

nlordell:
### Context and Motivation

Certain randomized secrets need to persist across reorgs. In particular the DKG secrets are one of them, as the validator may submit a commitment onchain, see a reorg, and roll back to before the secrets were generated, but have their commitment transaction still end up in the new canonical chain.

This PR introduces a secret store (backed by SQLite - using the same connection pool as the rest of the service), which can register and read KeyGen secrets which persist in a reorg-resistant way.

### Decisions and Tradeoffs

Right now the store is just a "dumb" store, you can register, read and prune. Managing stored secrets will be handled by the effects handler. Note that registering secrets does **not** overwrite the store. This simplifies the implementation for the effects handler where they do not require atomicity to perform an "upsert" (it can insert, and then read back from the DB).

This store will also handle nonces, but this will be added in a follow up PR to not make it too big to review.

### Testing

I added a `#[cfg(test)]` only `dummy()` constructor for `KeyShare` (which allowed us to remove an old constructor that we used before).

---

## CLA signature

### DIFF
--- a/crates/validator/src/secrets.rs
+++ b/crates/validator/src/secrets.rs
@@ -1,0 +1,210 @@
+//! The separate, reorg-resistant store for locally-generated random secrets.
+//!
+//! Two kinds of secret are sampled locally and then committed to onchain, and
+//! neither may live in the reorg-aware snapshot state on its own: the **DKG
+//! polynomial secrets** (a participant's random coefficients and ECDH
+//! encryption key) and the **FROST signing nonces**. A reorg that rolled either
+//! back while the transaction committing to it is re-included on the reorged
+//! chain would strand a keygen (the validator could no longer produce the
+//! matching shares) or risk reusing a nonce (which leaks the signing share).
+//!
+//! This store therefore lives in the shared [`SqlitePool`] but is deliberately
+//! **not** rolled back on reorg. It is reached only through the validator's
+//! effect handler, and its two kinds of secret are handled differently:
+//!
+//! - **DKG secrets** are reused (not resampled) when already present, so a
+//!   reorged-and-re-included commitment stays consistent with the shares the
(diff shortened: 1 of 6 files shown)


crates/validator/src/secrets.rs
```diff
@@ -0,0 +1,211 @@
...
+/// SQLite-backed store for locally-generated random secrets, over the shared
+/// pool. Unlike the snapshot store, it is never rolled back on reorg.
+pub struct SecretStore {
+    pool: SqlitePool,
+}
+
+impl SecretStore {
+    /// Creates the store backed by `pool`, creating its tables if absent.
+    pub async fn new(pool: SqlitePool) -> Result<Self, Error> {
+        // The `chunk` column of `nonces_chunks` is NULL until an onchain
```
rmeissner: nit: this comment doesn't seem to describe the dable created below (as there is no `nonces_chunks` table)

rmeissner (approved):
Probably worth fixing the Gemini comments

crates/validator/src/secrets.rs
nlordell: I accidentally left the comment behind when splitting the PR (the next one adds the nonces chunk table). Removing for now.

crates/validator/src/secrets.rs
```diff
@@ -0,0 +1,211 @@
...
+            .await?;
+        Ok(())
+    }
+}
+
+/// Encodes a fixed-byte value (group id, nonce root or address) as its
+/// lowercase hex text key, deterministic across calls.
+fn key(value: impl LowerHex) -> String {
+    format!("{value:x}")
+}
```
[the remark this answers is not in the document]
nlordell: TIL!

PR #544, opened 2026-07-09
[C 1ii]: Implement Nonces Storage

nlordell:
### Context and Motivation

This builds on #543 and adds nonce storage to the secret store as well.

### Decisions and Tradeoffs

The nonce store _removes_ the nonce on use (with `take_nonce`). This is done to prevent nonce reuse.

Note that we have two methods for reading nonces reveal info (one to read the reveal info, but not return a `Nonce` instance; since the reveal info is made public and therefore has no reuse restrictions) and another to take the nonce (for use when signing). The advantage of this over the simpler "just take nonce" is that it makes signing slightly more resilient to reorgs (as the reorg would only delete the nonce once the secret share action was already submitted, meaning the transaction may end up on chain after all, and produce an attestation even if the nonce is lost).

As a minor comment - we call the `offset` column `offs`, a common abbreviation for the term since `OFFSET` is a reserved SQL keyword.

---

## CLA signature

### DIFF
--- a/crates/validator/src/secrets.rs
+++ b/crates/validator/src/secrets.rs
@@ -22,7 +22,13 @@
 
 #![cfg_attr(not(test), expect(dead_code))]
 
-use crate::frost::keygen::Secrets;
+use crate::{
+    bindings,
+    frost::{
+        keygen::Secrets,
+        preprocess::{NonceChunk, Nonces},
+    },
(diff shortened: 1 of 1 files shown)


rmeissner (approved):
Gemini comment looks valid

PR #545, opened 2026-07-09
[B3d]: Handle new block

rmeissner:
Implements block advancement for the Sentinel state machine. Will handle transition to reveal phase and potential finalization. Otherwise drop stale request.

### Review

- Reference: https://github.com/safe-research/safenet/blob/c800c0c06416cda312c3401861bbc35f8692c26f/crates/sentinel/src/service.rs#L225-L255

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/servicev2.rs
+++ b/crates/sentinel/src/servicev2.rs
@@ -195,14 +195,17 @@ impl SentinelTransition {
         mut state: State,
         event: SentinelOracle::Revealed,
     ) -> (State, Vec<SentinelAction>) {
-        let Some(RequestState::CollectingVotes {
+        let Some(entry) = state.0.get_mut(&event.requestId) else {
+            return (state, Vec::new());
+        };
+        let RequestState::CollectingVotes {
             committed_count,
             revealed_count,
             approve_count,
             deny_count,
             self_revealed,
             ..
-        }) = state.0.get_mut(&event.requestId)
+        } = entry
         else {
             return (state, Vec::new());
         };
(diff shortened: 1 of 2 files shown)


crates/sentinel/src/servicev2.rs
```diff
@@ -168,7 +169,80 @@ impl SentinelTransition {
...
+    }
+
+    /// Drops requests we never got to commit on in time, reveals (or drops)
+    /// requests past their commit deadline, and finalizes requests past
+    /// their reveal deadline.
+    fn handle_block_advance(&self, mut state: State, block: u64) -> (State, Vec<SentinelAction>) {
+        let mut actions = Vec::new();
+        let ids: Vec<B256> = state.0.keys().copied().collect();
+        for id in ids {
+            let Some(existing) = state.0.get(&id).cloned() else {
```
nlordell: nit: cloning here feels like a code smell, note that there is an `Entry` API that lets you modify/remove/update, etc.

https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.entry

nlordell (approved):
LGTM,

The loop looks a little clunky - but not worth blocking the PR.

rmeissner:
@nlordell I changed the loop to use retain, can you check if you prefer this

crates/sentinel/src/servicev2.rs
```diff
@@ -218,36 +221,131 @@ impl SentinelTransition {
...
     ///
-    /// There are the following cases when the `Finalize` action is emitted:
-    /// - no one voted: a genuine timeout, where the bonds can be re-claimed
-    /// - unanimous vote: it is possible to claim the bond and reward
-    ///
-    /// In other cases it doesn't make sense to trigger the finalization for
-    /// this sentinel.
-    fn finalize(&self, mut state: State, request_id: B256) -> (State, Vec<SentinelAction>) {
-        let Some(RequestState::CollectingVotes {
+    /// `Finalize` is emitted whenever we stand to gain from it -- a dispute
```
rmeissner, line 317: note: there is some back and forth with this method, which is because of a misunderstanding in spec with Claude. This will be resolved in #548

nlordell (approved):

PR #547, opened 2026-07-09
[C 2]: Service Skeleton

nlordell:
### Context and Motivation

This PR adds skeleton types and trait implementations for the validator service, these will be filled progressively over phase D.

### Decisions and Tradeoffs

Originally C2 established actions, state with full types, but I thought that it was not a good approach (as it meant solidifying the FSM upfront). While this may seem reasonable (given the TS as a base), this also means that we cannot migrate away from TS-isms to Rust-isms throughout the phase D (state machine porting).

This PR also pre-maturely adds the requires `serde` traits to the bindings types (so they can exist in the state).

The epic was adjusted accordingly.

---

## CLA signature


crates/validator/src/service/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    bindings::{Consensus, Coordinator, Oracle},
+    state::{self, State},
+};
+use safenet_core::{driver::Service, watcher_events};
+
+/// The validator service bundle: the state transition, effect handler and
+/// action encoder that the driver runs.
+pub struct ValidatorService;
+
+watcher_events! {
```
rmeissner, line 20: uber nit: was expecting this to be in the binding mod

Edit: I see that you removed it from the binding mod, why? I don't really care where it is, but I think we should be consistent between validator and sentinel, so if we change it here I would also adjust it in the sentinel 😉

rmeissner (approved):

crates/validator/src/service/mod.rs
nlordell: This definition is specific to the service and unrelated to the contract bindings (where the actual events are defined).

In particular this type defines "which events the service watches" and not "which events exist on the contract".

PR #551, opened 2026-07-10
[D1i]: Group and Participant Set Derivation

nlordell:
### Context and Motivation

The first part of the validator state machine that we want to handle is to start key generation. In order to do this, we need some base libraries to compute group parameters that will be used for publishing the key generation commitment onchain.

### Decisions and Tradeoffs

We structure the API differently than in the TypeScript counterpart in that we produce an `Option<ParticipantSet>` given a validator set and an epoch. It needs to be an `Option` to return `None` in case the participant set is not valid (too few participants).

The `Config::active_before` became a `NonZeroU64`: a 0 value does not make sense as it would add a participant to the list which is impossible to include.

### Testing

I took data from an actual `keyGenAndCommit` transaction from our very own validator to make sure the computations are compatible with what is currently landing onchain.

---

## CLA signature

### DIFF
--- a/crates/validator/src/consensus/group.rs
+++ b/crates/validator/src/consensus/group.rs
@@ -1,0 +1,404 @@
+//! Key-generation group derivation.
+//!
+//! # Threshold and Minimum Participant Count
+//!
+//! Let:
+//!
+//! - `N` be the default number of participants;
+//! - `D` be the number of participants that drop out;
+//! - `M` be the number of malicious participants;
+//! - `T` be the minimum number of signers required; and
+//! - `t` be the fractional signing threshold.
+//!
+//! For every subgroup, the malicious participants must remain below the signing
+//! threshold:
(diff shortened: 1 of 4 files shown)


crates/validator/src/consensus/group.rs
```diff
@@ -0,0 +1,398 @@
...
+//!
+//! `D < N - N / (3 * t)`
+//!
+//! With `t = 1 / 2`, this becomes:
+//!
+//! `D < N / 3`
+//!
+//! Therefore, fewer than one third of the participants may drop out. Equivalently,
+//! the subgroup must contain strictly more than two thirds of the default
+//! participant set.
```
nlordell, line 38: I moved over this documentation section to preserve it once the TS validator gets deprecated.

rmeissner (approved):

PR #552, opened 2026-07-10
[B3g]: Flow test

rmeissner:
Implement flow tests for the transitions of the sentinel service to check that the triggered actions are as expected. The flows tested are:
- Commit and Reveal without dispute
- Dispute win and loss
- Timeout (non reveal of all parties)

### Design Decision
- The implemented tests are limited to the most common flows to avoid overtesting in an early state, but still provide confidence in the implementation

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/servicev2.rs
+++ b/crates/sentinel/src/servicev2.rs
@@ -63,7 +63,7 @@ pub struct SentinelEncoder {
     fee_token: Address,
 }
 
-#[expect(dead_code)]
+#[cfg_attr(not(test), expect(dead_code))]
 impl SentinelService {
     pub fn new(
         oracle: Address,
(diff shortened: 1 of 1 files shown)


crates/sentinel/src/servicev2.rs
```diff
@@ -530,3 +530,467 @@ impl Service for SentinelService {
...
+        );
+        assert!(commands.is_empty());
+        assert_eq!(
+            state.0[&id],
+            RequestState::WaitingForRequest {
+                approve: true,
+                deadline: 1 + VOTING_WINDOW,
+            },
+        );
+
```
nlordell, line 713: nit: The comments above each of the messages was useful for review.

nlordell (approved):
LGTM, tokio::test comments seem valid (but not important, it just means you setup async tests when you don't need them).

PR #553, opened 2026-07-10
[D 1ii]: Handle Genesis Key Generation

nlordell:
### Context and Motivation

This PR adds handling for the genesis `KeyGen` event that is manually emitted and used to signal "go" to the validators.

### Decisions and Tradeoffs

We use the effect system to build the commitments to ensure that the DKG secrets are stable across reorgs. Note that we chose to _not_ resubmit commitments in case secrets have already been generated, which assumes that the actions will always end up on chain.

Additionally, instead of making the group triggering fallible (as it is in the TS version), we prepare the genesis participant set upfront at startup (erroring out in case of an invalid configuration) and make the transition be infallible (the genesis group/participant set is guaranteed to be valid).

### Testing

For now, we do not include state transition tests. As discussed we want to check state transitions at a higher level (using mocked chain to test how the system as a whole behaves) instead of adding equality assertions on the state values themselves (which will change _dramatically_ over the next few PRs).

### Port

This is port of:

https://github.com/safe-research/safenet/blob/8ab73cfbf36ad2f6d7132ef9da6dff07b7c1055d/validator/src/machine/keygen/genesis.ts#L18-L49

https://github.com/safe-research/safenet/blob/8ab73cfbf36ad2f6d7132ef9da6dff07b7c1055d/validator/src/machine/keygen/trigger.ts#L17-L60

---

## CLA signature

### DIFF
--- a/crates/validator/src/service/action.rs
+++ b/crates/validator/src/service/action.rs
@@ -1,16 +1,64 @@
 //! Validator actions and their encoding into transactions.
 
+use crate::{
+    bindings::{self, Coordinator},
+    merkle::MerkleRoot,
+};
+use alloy::{
+    primitives::{Address, B256, U256},
+    sol_types::SolCall,
+};
 use safenet_core::{driver::ActionEncoder, tx::Transaction};
 
 /// An onchain action the validator emits during a state transition.
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub enum Action {}
+pub enum Action {
+    /// An action to publish key generation commitments onchain.
+    KeyGenAndCommit {
+        participants: MerkleRoot,
+        count: u16,
+        threshold: u16,
+        context: B256,
+        poap: Vec<B256>,
+        commitment: bindings::KeyGenCommitment,
+        expires_at: Option<u64>,
+    },
+}
 
-/// Encodes [`ValidatorAction`]s into the transactions the queue submits.
-pub struct Encoder;
+/// Encodes [`Action`]s into the transactions the queue submits.
+pub struct Encoder {
+    pub coordinator: Address,
+}
 
 impl ActionEncoder<Action> for Encoder {
     fn encode_action(&self, action: Action) -> (Transaction, Option<u64>) {
-        match action {}
+        match action {
+            Action::KeyGenAndCommit {
+                participants,
+                count,
+                threshold,
+                context,
+                poap,
+                commitment,
+                expires_at,
+            } => (
+                Transaction {
+                    to: self.coordinator,
+                    value: U256::ZERO,
+                    data: Coordinator::keyGenAndCommitCall {
+                        participants: participants.0,
+                        count,
+                        threshold,
+                        context,
+                        poap,
+                        commitment,
+                    }
+                    .abi_encode()
+                    .into(),
+                    gas: 250_000,
+                },
+                expires_at,
+            ),
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


crates/validator/src/service/effect.rs
```diff
@@ -1,20 +1,110 @@
...
+                let mut rng = rand::thread_rng();
+                let setup = frost::keygen::setup(&mut rng, self.account, count, threshold)?;
+                let stored = self
+                    .secrets
+                    .store_keygen_secrets(id, self.account, &setup.secrets)
+                    .await?;
+                if !stored {
+                    tracing::info!(group_id = %id, "not resubmitting already created keygen commitments");
+                    return Ok(Resume::Noop);
+                }
```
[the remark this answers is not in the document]
nlordell: I do not do this, as it will be true for 99.999% of the time, and therefore the additional DB round trip is not worth it.

crates/validator/src/service/action.rs
```diff
@@ -1,16 +1,64 @@
...
+                        threshold,
+                        context,
+                        poap,
+                        commitment,
+                    }
+                    .abi_encode()
+                    .into(),
+                    gas: 250_000,
+                },
+                expires_at.unwrap_or(u64::MAX),
```
nlordell: This is like this as I still need to rebase @rmeissner's changes (at which point it will be `None` for not expire).

crates/validator/src/service/effect.rs
```diff
@@ -1,20 +1,110 @@
...
+}
 
 impl EffectHandler<Effect, Resume> for Handler {
     async fn perform_effect(&mut self, effect: Effect) -> Resume {
-        match effect {}
+        match self.try_perform_effect(effect.clone()).await {
+            Ok(resume) => resume,
+            Err(err) => {
+                tracing::warn!(?effect, %err, "failed to perform effect");
+                Resume::Noop
```
rmeissner, line 87: question: so if an effect fails we continue operation? Can this be a problem? Somehow I would have expected that this never happens and therefore we should fail loudly.

crates/validator/src/state/keygen.rs
```diff
@@ -0,0 +1,50 @@
...
+        let genesis = self.genesis.group();
+        if state.rollover != RolloverState::WaitingForGenesis || event.gid != genesis.id() {
+            return (state, Vec::new());
+        }
+
+        // The genesis group generation is not subject to a rollover deadline.
+        state.rollover = RolloverState::CollectingCommitments {
+            group_id: genesis.id(),
+            next_epoch: 0,
+            deadline: u64::MAX,
```
rmeissner: nit: I was told that that this is a code smell and we should rather have an option 😛

rmeissner: nit: I was told that that this is a code smell and we should rather have an option 😛

rmeissner (approved):

crates/validator/src/state/keygen.rs
nlordell: :100: (this was verbatim from the TS state).

crates/validator/src/service/effect.rs
nlordell: For me, `warn!` is a loud failure (`error!` is reserved for me for "we are stopping"). We should not get any warnings.

In theory, _in the general case_, we should continue with the state machine even if the effect fails (for a regular epoch: I cannot write to my DB, but I can still observe the other KeyGen events and then decide that I misbehaved and am not included in the next attempt at keygen).

Granted, for genesis this is a little different (as we cannot proceed). I will have to think about it a little more on how to codify the genesis exception.

rmeissner: yeah, not a blocker, was just trying to align my mental model for the effects. I am also totally fine to say, lets keep it as is and adjust if necessary later.

PR #557, opened 2026-07-10
[D 1ii/Refactor] Refactor Secrets Sampling

nlordell:
### Context and Motivation

I noticed in further changes to the state machine that we _need_ the secrets value right away for verifying commitments. This change modifies the effect to just sample the random secrets, store them, and then resume the state machine with them; the state machine is responsible to push the commit action once it sees the result of the effect.

This allows the state machine to hold a copy of the `Secrets` value so that it can use in its state transitions.

As before, secrets are just sampled once and reused per group ID.

### Decisions and Tradeoffs

A new `WiatingForSetup` state was introduced. This gives us a nice bonus as failure to write to the storage or generate the secrets gives us a chance to resample. This makes the state machine more resilient (resampling will be added once we start handling block updates - if we see a new block update and are still in the `WaitingForSetup` phase, then we know that we should try sampling again).

---

## CLA signature

### DIFF
--- a/crates/validator/src/service/effect.rs
+++ b/crates/validator/src/service/effect.rs
@@ -1,6 +1,9 @@
 //! The validator effect system and its handler.
 
-use crate::{frost, merkle::MerkleRoot, secrets::SecretStore, service::Action};
+use crate::{
+    frost::{self, keygen::Secrets},
+    secrets::SecretStore,
+};
 use alloy::primitives::{Address, B256};
 use safenet_core::state::EffectHandler;
 use std::{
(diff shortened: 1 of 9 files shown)


crates/validator/src/service/effect.rs
```diff
@@ -44,35 +48,21 @@ pub struct Handler {
...
-                poap,
-                expires_at,
             } => {
                 let mut rng = rand::thread_rng();
-                let setup = frost::keygen::setup(&mut rng, self.account, count, threshold)?;
+                let secrets = frost::keygen::setup(&mut rng, self.account, count, threshold)?;
                 let stored = self
                     .secrets
-                    .store_keygen_secrets(id, self.account, &setup.secrets)
+                    .store_keygen_secrets(group_id, self.account, secrets)
```
rmeissner, line 60: question: Can this not fail anymore?

crates/validator/src/state/keygen.rs
```diff
@@ -13,38 +19,93 @@ impl Transition {
     /// rotation is triggered on new blocks.
     pub(super) fn handle_genesis_key_gen(
         &self,
-        mut state: State,
+        state: State,
         event: &Coordinator::KeyGen,
     ) -> (State, Commands<State, Self>) {
         let genesis = self.genesis.group();
-        if state.rollover != RolloverState::WaitingForGenesis || event.gid != genesis.id() {
+        if !matches!(state.rollover, RolloverState::WaitingForGenesis) || event.gid != genesis.id()
```
rmeissner, line 26: question: why switch to the matches! macro?

rmeissner (approved):

crates/validator/src/state/keygen.rs
nlordell: Because I needed to remove `PartialEq` and `Eq` from the type (generally, FROST types are opaque).

crates/validator/src/service/effect.rs
nlordell: It can still fail, there is an `.await?`.

rmeissner: I was asking before we had the explicit check

PR #565, opened 2026-07-11
[D 1viii]: Schedule KeyGen Secrets Pruning

nlordell:
### Context and Motivation

KeyGen DKG secrets need to be pruned from the secret store. This PR does the first step in that direction by registering group IDs for DKG secret pruning.

### Decisions and Tradeoffs

KeyGen secrets need to be pruned, but in a reorg-aware way. The simplest solution to this is to queue up things to prune with the block on which they should be pruned, and the emit the correct effects when we observe `safe` blocks past the pruning point. The actual pruning will happen when we introduce block handling in D4.

I also simplified some of the failure paths with a macro (the reason for using a macro is explained in its doc comment). The `use ${MACRO_NAME}` pattern just allows the macro to appear anywhere in the file (instead of before first use, macros have a different and weird scoping rules).

---

## CLA signature

### DIFF
--- a/crates/validator/src/state/keygen.rs
+++ b/crates/validator/src/state/keygen.rs
@@ -1,4 +1,4 @@
-use super::{RolloverState, State, Transition};
+use super::{Prune, RolloverState, State, Transition};
 use crate::{
     bindings::Coordinator,
     consensus::{
@@ -177,13 +177,7 @@ impl Transition {
                     Err(err) => {
                         // There was an issue with the verified commitments,
                         // which is an unexpected an unrecoverable error.
-                        return (
-                            State {
-                                rollover: rollover_failure(next_epoch, err),
-                                ..state
-                            },
-                            Vec::new(),
-                        );
+                        return fail_rollover!(state, block, next_epoch, group.id(), err);
                     }
                 };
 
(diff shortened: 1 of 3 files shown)


crates/validator/src/state/keygen.rs
```diff
@@ -340,13 +334,7 @@ impl Transition {
                             Err(err) => {
                                 // Finalization failures are unexpected, since
                                 // all secret shares were already verified.
-                                return (
-                                    State {
-                                        rollover: rollover_failure(next_epoch, err),
-                                        ..state
-                                    },
-                                    commands,
```
nlordell, line 348: `commands` is always empty here - so while it _may_ look like a breaking change, it isn't.

crates/validator/src/state/keygen.rs
```diff
@@ -684,13 +683,7 @@ impl Transition {
                     Err(err) => {
                         // Finalization failures are unexpected, as all secret
                         // shares were already verified.
-                        return (
-                            State {
-                                rollover: rollover_failure(next_epoch, err),
-                                ..state
-                            },
-                            commands,
```
nlordell, line 692: idem.

crates/validator/src/state/keygen.rs
```diff
@@ -667,3 +815,22 @@ fn rollover_failure(next_epoch: EpochId, err: impl Display) -> RolloverState {
...
+            State {
+                rollover: rollover_failure(next_epoch, err),
+                ..($state)
+            }
+            .and_prune(block, Prune::KeyGenSecrets { group_id }),
+            Vec::new(),
+        )
+    }};
+}
+use fail_rollover;
```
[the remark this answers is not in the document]
nlordell: computer says no.

(this suggestion breaks compilation.).

remedcu (approved):

PR #574, opened 2026-07-12
[D 3iii]: Oracle Result Handling

nlordell:
### Context and Motivation

Handle oracle results and proceed with signing.

### Port

https://github.com/safe-research/safenet/blob/4c3188fd65d977c17c5e26c4785ec7eb8193f2a5/validator/src/machine/consensus/oracleResult.ts#L14-L36

---

## CLA signature

### DIFF
--- a/crates/validator/src/state/sign.rs
+++ b/crates/validator/src/state/sign.rs
@@ -1,9 +1,9 @@
 use super::{Packet, SigningState, State, Transition};
 use crate::{
-    bindings::{Coordinator, SignNonces},
+    bindings::{Coordinator, Oracle, SignNonces},
     service::{Action, Effect},
 };
-use alloy::primitives::B256;
+use alloy::primitives::{Address, B256};
 use safenet_core::state::{Command, Commands};
 use std::collections::BTreeMap;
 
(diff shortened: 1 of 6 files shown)


remedcu (approved):

crates/validator/src/state/sign.rs
```diff
@@ -122,4 +122,73 @@ impl Transition {
...
+                )
+            }
+            Some(SigningState::WaitingForOracle {
+                signature_id,
+                oracle: expected,
+                ..
+            }) if expected == oracle && !event.approved => {
+                // Rejected: drop the session, along with the signature id
+                // index entry eagerly set when the round was opened.
+                state.signature_id_to_message.remove(&signature_id);
```
rmeissner, line 180: Why do we not trigger an onchain decline for this?

nlordell: The answer was simply "cuz TS did it": https://github.com/safe-research/safenet/blob/e74bf0c181fdd81e1036335c0e5aaa66ce5dcc0a/validator/src/machine/consensus/oracleResult.ts#L23-L26

We talked synchronously about whether or not this is a good idea, or if we should explicitly decline signing:

- declining signing was (at least AFAIU) a stop-gap to allow for better UX where we could _show_ to the user that validators were choosing to not attest something (because the transaction was invalid). For oracle transactions, this reason is no longer valid (because the oracle result already provides enough information for interfaces to display an "invalid transaction" in the interface).
- on the other hand, it does mean that we can't really tell which validators are participating or not for a given oracle transaction. That being said, declining to sign is already indicative only, and I don't think we should consider it for participation (WRT rewards payouts specifically for example, where we already don't consider events that don't produce attestations for participation).

For that reason, I would actually leave it as is.

PR #580, opened 2026-07-13
[D 4i]: Handle Epoch Rollovers

nlordell:
### Context and Motivation

Start hooking up the the block handler in order to schedule rollovers for staged blocks.

### Decisions and Tradeoffs

> [!NOTE]
> Unlike the TypeScript version, we are more exact with the epochs that we reap and prune: we remove the previously active epoch _as long as there are no active signing ceremonies_. This allows us to be more precise with our epoch pruning compared to the TypeScript implementation. It does not cause consensus differences. TypeScript defensively keeps the previous epoch around.

This does not handle _all_ block updates for the keygen logic. In particular timeouts will be handled separately in a follow-up PR.

### Port

https://github.com/safe-research/safenet/blob/4c3188fd65d977c17c5e26c4785ec7eb8193f2a5/validator/src/machine/consensus/rollover.ts#L18-L78

https://github.com/safe-research/safenet/blob/4c3188fd65d977c17c5e26c4785ec7eb8193f2a5/validator/src/machine/consensus/rollover.ts#L101-L120

---

## CLA signature

### DIFF
--- a/crates/validator/src/consensus/epoch.rs
+++ b/crates/validator/src/consensus/epoch.rs
@@ -18,24 +18,6 @@ pub enum EpochId {
     Number { number: NonZeroU64 },
 }
 
-impl Serialize for EpochId {
-    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        serializer.serialize_u64(self.raw_value())
-    }
-}
-
-impl<'de> Deserialize<'de> for EpochId {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        Ok(Self::from_raw(u64::deserialize(deserializer)?))
-    }
-}
-
 impl EpochId {
     /// Returns the epoch ID for a raw value.
     pub const fn from_raw(value: u64) -> Self {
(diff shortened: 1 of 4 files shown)


crates/validator/src/consensus/epoch.rs
```diff
@@ -18,24 +18,6 @@ pub enum EpochId {
...
-}
-
-impl<'de> Deserialize<'de> for EpochId {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        Ok(Self::from_raw(u64::deserialize(deserializer)?))
-    }
-}
```
nlordell, line 37: Moved down for consistence (so that all trait implementations are grouped together after the main `impl` block).

crates/validator/src/state/keygen.rs
```diff
@@ -841,30 +934,6 @@ impl Transition {
...
-                number.get(),
-                number
-                    .get()
-                    .saturating_mul(self.config.blocks_per_epoch.get()),
-            )
-                .abi_encode()
-                .into(),
-        })
-    }
-
```
nlordell, line 867: This just moved down verbatim for consistency (so that `start` and `restart` keygen are beside eachother - this was positioned weirdly in between the two before).

remedcu (approved):

crates/validator/src/state/keygen.rs
```diff
@@ -909,6 +978,66 @@ impl Transition {
...
+
+impl RolloverState {
+    /// The epoch this rollover state is working toward becoming active, or
+    /// `None` for the terminal [`Self::Halted`] state.
+    fn next_epoch(&self) -> Option<EpochId> {
+        match self {
+            RolloverState::WaitingForGenesis => Some(EpochId::Genesis),
+            RolloverState::EpochSkipped { next_epoch }
+            | RolloverState::SigningRollover { next_epoch, .. }
+            | RolloverState::EpochStaged { next_epoch, .. } => Some(EpochId::Number {
```
rmeissner, line 1015: Interesting that the different states have different types for the epoch number.

PR #582, opened 2026-07-13
[D 4iii]: Handle Epoch Staged

nlordell:
### Context and Motivation

Handle the epoch staged event, registering the epoch if participating (so we can use it for signing).

### Decisions and Tradeoffs

> [!NOTE]
> Unlike TS, if we see a staged epoch while we are waiting for genesis, we assume that we joined partway through consensus and skip to that epoch. This has the same effect as `skip_genesis` but in a less hacky way (does not require additional configuration, and is automatically deduced just by observing events).

Previously, in #578, we had temporarily hooked up rollover signing to be able to test the genesis rollover change in #579; this was slightly incorrect (the epoch should only get registered once there is an attestation for it - otherwise it might stay dangling since the concensus contract will never roll over to it). This is fixed in this PR.

### Port

https://github.com/safe-research/safenet/blob/4c3188fd65d977c17c5e26c4785ec7eb8193f2a5/validator/src/machine/consensus/epochStaged.ts#L14-L64

---

## CLA signature

### DIFF
--- a/crates/validator/src/state/keygen.rs
+++ b/crates/validator/src/state/keygen.rs
@@ -3,14 +3,14 @@ use super::{
     RolloverState, SigningState, State, Transition,
 };
 use crate::{
-    bindings::Coordinator,
+    bindings::{Consensus, Coordinator},
     consensus::{
         epoch::{self, EpochId},
-        group::{self, ParticipantSet},
+        group::{self, Group, ParticipantSet},
     },
     frost::{
         self,
-        keygen::{GroupCommitments, Secrets},
+        keygen::{GroupCommitments, KeyShare, Secrets},
     },
     service::{Action, Effect},
 };
@@ -23,6 +23,7 @@ use std::{
     collections::{BTreeMap, BTreeSet},
     fmt::Display,
     iter, mem,
+    num::NonZeroU64,
     sync::Arc,
 };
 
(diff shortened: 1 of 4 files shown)


crates/validator/src/state/keygen.rs
```diff
@@ -515,12 +496,12 @@ impl Transition {
...
                                 rollover: RolloverState::SigningRollover {
                                     next_epoch: proposed_epoch,
-                                    group_id,
+                                    group,
+                                    key_share,
                                     message,
                                 },
                                 ..state
-                            }
-                            .and_prune(block, Prune::KeyGenSecrets { group_id }),
```
nlordell, line 523: pruning is now handled in the same place: during `finalize_key_gen`.

crates/validator/src/state/transactions.rs
```diff
@@ -71,36 +70,14 @@ impl Transition {
...
-        match &signing {
-            Some(SigningState::WaitingForAttestation { .. }) | None => {}
-            Some(_) => tracing::warn!(
-                %message,
-                signature_id = %event.signatureId,
-                "received attestation on unexpected signing state"
-            ),
-        }
-
-        (state, Vec::new())
```
nlordell, line 103: Moved verbatim to `handle_sign_attested`.

crates/validator/src/state/transactions.rs
```diff
@@ -160,49 +137,13 @@ impl Transition {
...
-    }
```
nlordell, line 193: idem

crates/validator/src/state/transactions.rs
```diff
@@ -160,49 +137,13 @@ impl Transition {
...
-    /// Returns the known signature ID for a signing state, or `None` if none
-    /// have been assigned yet.
-    fn signature_id(&self) -> Option<B256> {
-        match self {
-            SigningState::WaitingForAttestation { signature_id, .. }
-            | SigningState::WaitingForOracle { signature_id, .. }
-            | SigningState::CollectNonceCommitments { signature_id, .. }
-            | SigningState::CollectSigningShares { signature_id, .. } => Some(*signature_id),
-            SigningState::WaitingForRequest { .. } | SigningState::WaitingToDecline { .. } => None,
-        }
```
nlordell, line 206: Moved verbatim to the `sign` module, where it made more sense anyway.

remedcu (approved):

PR #583, opened 2026-07-13
[D 4iv A]: Attestation Actions

nlordell:
### Context and Motivation

This PR introduces attestation actions and their encoding implementation that are needed for handling signing timeouts.

It is submitted as a separate PR in effort to facilitate review.

### Decisions and Tradeoffs

I noticed that some expiries were `Option<u64>` when they should have been `u64` (all signing states and transactions have an expiry, unlike keygen - because of genesis); so I corrected the types.

There are a couple `#[expect(dead_code)]`, as the added variants and methods here do not get used until the next PR where they are removed.

---

## CLA signature

### DIFF
--- a/crates/validator/src/service/action.rs
+++ b/crates/validator/src/service/action.rs
@@ -1,7 +1,8 @@
 //! Validator actions and their encoding into transactions.
 
 use crate::{
-    bindings::{self, Coordinator},
+    bindings::{self, Consensus, Coordinator},
+    consensus::epoch::EpochId,
     merkle::MerkleRoot,
 };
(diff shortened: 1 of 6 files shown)


remedcu (approved):

nlordell:
~~@rmeissner - I created a [`rebase/oxval-Div-a`](https://github.com/safe-research/safenet/tree/rebase/oxval-D4iv-a) branch that is rebased on `main` (they are identical). I left this branch as-is in order to keep #584 reviewable (without needing to rebase the entire stack).~~

PR #593, opened 2026-07-20
[Epic] Add reason to vote

rmeissner:
To provide insight into why a specific vote was given a reason string is added to a vote. This allows better observability and UX.

---

## CLA signature

### DIFF
--- a/epics/2026_07_20_sentinel_vote_reason.md
+++ b/epics/2026_07_20_sentinel_vote_reason.md
@@ -1,0 +1,102 @@
+# Plan: Sentinel Vote Reason
+
+Component: `contracts` (Solidity) and `crates/sentinel` (Rust).
+
+---
+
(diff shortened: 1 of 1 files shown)


remedcu (commented):
Why a `string` within the event and not a specific number? Similar to error codes for Safe, so, for section one, everything starts with 1XXXXXX, for section 1.1.2, we can have 1001002

We could even do with enums, if the charter is somewhat fixed or limited in range? Or if extended later can append to the enum?

rmeissner:
I don't think we can do enums, as it is an ever evolving set (as the charter might change and there is "common law"). Also string is the most flexible, we can encode anything. Number would most likely do it too, but we explicitly don't do gas optimizations yet, therefore I opted to the most flexible for the start.

remedcu (approved):

PR #604, opened 2026-07-20
[Guard - Step 5] Add SafenetGuard contract

remedcu:
Adds the `SafenetGuard` contract itself: a standalone Safe transaction guard that gates every owner-signed `execTransaction` behind a FROST threshold-signature attestation, with the nonce-free escape hatch for liveness. Includes the `via_ir` compile restriction it requires and the guard README.

### Context and Motivation

This is the core of the stack, composed from the libraries and interface below it: epoch state (`EpochRollover`), announcement state and hashing (`TransactionAnnouncement`), trailer decode (`AttestationTrailer`), and the self-call gate (`GuardAutoAllow`). `checkTransaction` authorises in order — auto-allowed self-call, then a valid attestation trailer, then a matured announcement — and reverts `AttestationNotFound` otherwise.

### Decisions and Tradeoffs

- `foundry.toml` adds `src/guard/SafenetGuard.sol` to `compilation_restrictions` with `via_ir = true`; the contract otherwise fails to compile with *Stack too deep*.
- A recognised trailer commits to the attestation path (an untrusted key, bad signature, or malformed trailer reverts rather than falling through), so an attested transaction never consumes a matching announcement.
- The trusted `(groupKey, epoch)` forest is kept forever: shares are destroyed after rollover, so retaining historical keys as valid is an accepted trust assumption (documented in the README). A compromised historical key cannot replay past transactions because the Safe nonce binds them.
- Scope is transaction-guard only; module executions bypass it. Documented in the README as an explicit deployment constraint.

### Testing

- `cd contracts && forge build` (compiles with the added `via_ir` restriction), `forge fmt --check`, and `forge lint --deny notes` pass.
- Full behavioural coverage — real Safe + real FROST signing — lands in the next PR to keep this one focused on the implementation.

### Stack

- Base: `guard/4-guard-auto-allow` (this PR is stacked on it, transitively on all libraries and the interface).
- Previous: the GuardAutoAllow library PR.
- Next: the SafenetGuard test-suite PR, then the deployment-script PR.


contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+        Enum.Operation operation,
+        uint256 safeTxGas,
+        uint256 baseGas,
+        uint256 gasPrice,
+        address gasToken,
+        address payable refundReceiver,
+        bytes calldata signatures,
+        address /* msgSender */
+    ) external override {
+        if (_isAutoAllowed(to, value, data, operation)) return;
```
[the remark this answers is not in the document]
remedcu: All tests are in #605

contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * @inheritdoc ISafenetGuard
+     */
+    function announceTransaction(TransactionAnnouncement.AnnouncedTransaction calldata announcement) external {
+        bytes32 announcementHash = TransactionAnnouncement.hash(announcement);
+        (uint256 activeFrom, uint256 activeUntil) =
+            $announcements.announce(msg.sender, announcementHash, _ALLOW_TX_DELAY, _ALLOW_TX_WINDOW);
+
+        // Hash-only event on Ethereum mainnet (dear log data), full parameters elsewhere.
```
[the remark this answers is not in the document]
remedcu: Updated 👍🏾

contracts/src/guard/README.md
```diff
@@ -4,6 +4,115 @@ This directory contains a collection of Safe Guard implementations for Safenet.
 
 ---
 
+## SafenetGuard
+
+**File:** `SafenetGuard.sol`
+
+SafenetGuard is a refinement of the SafenetGuardA design (inline attestation, transaction guard only), re-assembled from focused, independently-auditable libraries so that future changes to any one concern touch a single isolated audit surface. It implements Safe's `BaseTransactionGuard` and gates every owner-signed transaction behind a FROST threshold-signature attestation.
+
+**Scope.** This is a *transaction* guard: it gates only owner-signed `execTransaction` calls. Safe module executions do not invoke transaction-guard hooks, so an enabled module can move assets without an attestation or announcement. Deployments must prohibit modules or treat each enabled module as an explicit bypass of this guard's policy.
```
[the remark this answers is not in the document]
remedcu: Updated 👍🏾

contracts/src/guard/README.md
```diff
@@ -4,6 +4,115 @@ This directory contains a collection of Safe Guard implementations for Safenet.
 
 ---
 
+## SafenetGuard
```
rmeissner: nit/ non blocking request: would have been nice to split the readme into a separate PR to keep the PR size small.

contracts/src/guard/README.md
```diff
@@ -4,6 +4,115 @@ This directory contains a collection of Safe Guard implementations for Safenet.
...
+
+**File:** `SafenetGuard.sol`
+
+SafenetGuard is a refinement of the SafenetGuardA design (inline attestation, transaction guard only), re-assembled from focused, independently-auditable libraries so that future changes to any one concern touch a single isolated audit surface. It implements Safe's `BaseTransactionGuard` and gates every owner-signed transaction behind a FROST threshold-signature attestation, except when a matured announcement authorises execution via the nonce-free escape hatch.
```
rmeissner: request for change: I would not reference SafenetGuardA here, as I would expect that we remove the legacy guards once this canonical SafenetGuard is merged.

contracts/src/guard/README.md
```diff
@@ -4,6 +4,115 @@ This directory contains a collection of Safe Guard implementations for Safenet.
...
+**Scope.** This is a *transaction* guard: it gates only owner-signed `execTransaction` calls. Safe module executions do not invoke transaction-guard hooks, so an enabled module can move assets without an attestation or announcement. Deployments must prohibit modules or treat each enabled module as an explicit bypass of this guard's policy.
+
+### What it reuses
+
+- **`EpochRollover`** (`../libraries/EpochRollover.sol`) — the trusted epoch state and FROST-verified rollover, shared with the rest of Safenet rather than reimplemented in the guard.
+- **`TransactionAnnouncement`** (`../libraries/TransactionAnnouncement.sol`) — the escape-hatch announcement type and nonce-free hashing (`AnnouncedTransaction`, `hash`) plus the time-windowed state (`announce` / `cancel` / `consume`).
+- **`AttestationTrailer`** (`../libraries/AttestationTrailer.sol`) — recognising and decoding the inline attestation trailer from Safe's `signatures` bytes.
+- **`GuardAutoAllow`** (`../libraries/GuardAutoAllow.sol`) — the structural gate for guard self-calls (target, zero value, `CALL`-not-`DELEGATECALL`, selector extraction).
+
+### How it differs from SafenetGuardA
```
rmeissner: request for change: same as above I would remove all references to previous guards and rather rephrase this as properties of this implementations

contracts/src/guard/README.md
```diff
@@ -4,6 +4,115 @@ This directory contains a collection of Safe Guard implementations for Safenet.
...
+
+**Epoch forest instead of a linear chain.** SafenetGuardA keeps a single `$activeEpoch` and a one-key-per-epoch mapping. SafenetGuard delegates epoch state to `EpochRollover`, which tracks a *forest* of trusted `(group key, epoch)` pairs: any trusted pair may sign a rollover to any strictly-greater epoch, an epoch may hold more than one key (reorg branches), and every pair is kept forever. There is no single "active" epoch — `updateEpoch` names the exact `(parentKey, parentEpoch)` to roll over from, and the view function is `isKnownEpoch(groupKey, epoch)` rather than `activeEpoch()`.
+
+**Attestation trailer carries the group key, framed by a magic word.** Because the forest is keyed by key coordinates (no reverse "key for epoch N" lookup) and an epoch may hold several keys, the inline attestation carries the key explicitly. The trailer is a fixed 192-byte payload followed by a 32-byte magic word:
+
+```
+[safe owner signatures] [192-byte abi.encode(epoch, Secp256k1.Point groupKey, FROST.Signature)] [32-byte MAGIC]
```
rmeissner: nit/ potential for clarification: This will be the first canonical guard that the Safe project publishes that uses signature extension, right? I think it is worth mentioning this and also documenting this as a future standard. The type hash (here called MAGIC) is a good separator as it can be easily extended. Would really love if we can "standardize" this (at least for ourselves).

contracts/src/guard/README.md
```diff
@@ -4,6 +4,115 @@ This directory contains a collection of Safe Guard implementations for Safenet.
...
+### Pros
+
+- Thin contract body: each concern (epoch state, announcements, trailer parsing, auto-allow) lives in a small, separately-auditable library. Future changes are minimal to re-audit.
+- Shares the same `EpochRollover` library as the rest of Safenet, so epoch semantics cannot drift between components.
+- Nonce-free escape hatch: a queued announcement survives unrelated nonce advances, so owners can keep using the Safe via attestation while an announcement matures — the hatch stays usable even under concurrent activity.
+- Announce-by-parameters: signers see the full transaction they authorise (not an opaque hash), and the guard derives the hash on-chain so it cannot silently diverge from what `checkTransaction` recomputes.
+- Bounded validity window: announcements expire after `activeUntil`, so a stale critical transaction cannot be executed indefinitely; both window bounds share one storage slot.
+- Inherits SafenetGuardA's advantages: inline happy path with no separate registration, nonce-based replay protection for attested execution, permissionless rollover, and a liveness escape hatch.
+
+### Cons
```
rmeissner: uber nit: I am not sure where to best document this. For me it feels misplaced in the "root" readme, considering that this is the final version. Maybe having a dedicated "design decisions" md file would make sense. Also there I would then not phrase it as "pro and cons" rather as decisions taken. 

For this PR it can stay here, this is more a future clean up (together with removal of the other guards).

contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+
+    // ============================================================
+    // CONSTANTS & IMMUTABLES
+    // ============================================================
+
+    /**
+     * @dev Ethereum mainnet; announcements there emit the hash-only event (log data is costly, and the
+     *      parameters are recoverable from the announcement calldata).
+     */
+    uint256 private constant _ETHEREUM_CHAIN_ID = 1;
```
rmeissner: nit: as mentioned before I would remove this, as I feel this adds unnecessary complexity.

contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+     * @param initialGroupKey Genesis FROST group key; must be a non-zero secp256k1 point.
+     * @param allowTransactionDelay Escape-hatch embargo (seconds); non-zero, at most `type(uint64).max`.
+     * @param allowTransactionWindow Escape-hatch window (seconds); non-zero, at most `type(uint64).max`.
+     */
+    constructor(
+        uint256 consensusChainId,
+        address consensusAddress,
+        uint64 initialEpoch,
+        Secp256k1.Point memory initialGroupKey,
+        uint256 allowTransactionDelay,
```
rmeissner, line 97: question: I assume not passing the delay and window as a uint64 is a gas optimization?

contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+    // ============================================================
+
+    /**
+     * @notice Deploys SafenetGuard seeded with the Safenet genesis epoch.
+     * @param consensusChainId Chain ID hosting the Consensus contract (100 for Gnosis Chain).
+     * @param consensusAddress Consensus contract address; must be non-zero. With `consensusChainId` it
+     *                         derives the immutable domain separator (uncorrectable post-deploy).
+     * @param initialEpoch Genesis epoch number.
+     * @param initialGroupKey Genesis FROST group key; must be a non-zero secp256k1 point.
+     * @param allowTransactionDelay Escape-hatch embargo (seconds); non-zero, at most `type(uint64).max`.
```
rmeissner, line 89: question: why was uint64 chosen? I understand that the sum of the two and the current block timestamp always has to be smaller uint128 max. Is this the primary reason?

contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+                    safeTxGas: safeTxGas,
+                    baseGas: baseGas,
+                    gasPrice: gasPrice,
+                    gasToken: gasToken,
+                    refundReceiver: address(refundReceiver),
+                    nonce: nonce
+                })
+            );
+            // Forest membership implies a non-zero key (enforced on record), so no extra check is needed.
+            require($epochs.isKnown(groupKey, epoch), UntrustedAttestationKey());
```
rmeissner: nit: wouldn't it make sense to check this before we calculate the hash?

contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+        }
+
+        revert AttestationNotFound();
+    }
+
+    /**
+     * @notice Post-execution hook. Intentionally empty — all authorisation happens in `checkTransaction`.
+     * @dev Parameters are unnamed as they are unused; Safe passes the transaction hash and success flag.
+     */
+    function checkAfterExecution(bytes32, bool) external pure override {}
```
rmeissner: question (maybe worth mentioning): We don't need any check if the guard is still present as we assume that this is handled via Safenet, right?

contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * @inheritdoc IERC165
+     * @dev Advertises `ISafenetGuard` alongside `ITransactionGuard` and `IERC165`.
+     */
+    function supportsInterface(bytes4 interfaceId) external pure override returns (bool) {
+        return interfaceId == type(ISafenetGuard).interfaceId || interfaceId == type(ITransactionGuard).interfaceId
```
rmeissner, line 266: nit: just realized why does the SafenetGuard interface not extend the TransactionGuard interface?

contracts/src/guard/SafenetGuard.sol
```diff
@@ -0,0 +1,351 @@
...
+     * @param data The call data.
+     * @param operation The Safe operation type.
+     * @return allowed True if the call is an auto-allowed escape-hatch self-call.
+     */
+    function _isAutoAllowed(address to, uint256 value, bytes memory data, Enum.Operation operation)
+        private
+        view
+        returns (bool)
+    {
+        bytes4 selector = GuardAutoAllow.selfCallSelector(to, value, data, operation, address(this));
```
rmeissner: nit: now that I see how the GuardAutoAllow lib is used I am not sure if it was worth extracting that into a separate contract. Personally I would opt for explicity here.

rmeissner (changes requested):
All my request changes are related to the README, maybe worth splitting into a separate PR to avoid this blocking the contract.

contracts/src/guard/README.md
remedcu: Moved to #611

contracts/src/guard/README.md
remedcu: Removed 👍🏾 in #611

contracts/src/guard/README.md
remedcu: Done

contracts/src/guard/README.md
remedcu: Noted 👍🏾

contracts/src/guard/SafenetGuard.sol
remedcu: Removed 👍🏾

contracts/src/guard/SafenetGuard.sol
remedcu: Moved 👍🏾

contracts/src/guard/SafenetGuard.sol
remedcu: Yes

contracts/src/guard/SafenetGuard.sol
remedcu: Updated 👍🏾

contracts/src/guard/SafenetGuard.sol
remedcu: Updated 👍🏾

rmeissner (approved):

PR #605, opened 2026-07-20
[Guard - Step 6] Add SafenetGuard test suite

remedcu:
Adds the behavioural test suite for `SafenetGuard`, using a real Safe with the guard actually set and real FROST signing rather than a mock.

### Context and Motivation

Split from the implementation PR because the suite is ~1,400 lines; reviewing it separately from the ~460-line contract keeps both diffs digestible. The tests exercise every authorisation path end-to-end: auto-allowed self-calls, attestation-trailer acceptance/rejection (untrusted key, bad signature, malformed/absent trailer), the announce → delay → execute → single-use escape hatch, cancellation, epoch rollover, and the mainnet vs off-mainnet event variants.

### Decisions and Tradeoffs

- Uses an actual Safe and sets the guard via `setGuard`, then drives real `execTransaction` calls, so the trailer parsing and nonce binding are validated against Safe's true behaviour.
- FROST signatures are produced with the repo's `ForgeSecp256k1` test utility against the same message construction the guard verifies.

### Testing

- `cd contracts && forge test --match-path 'test/SafenetGuard.t.sol'` — 67 tests, all passing. `forge fmt --check` and `forge lint --deny notes` pass.

### Stack

- Base: `guard/5-safenet-guard` (this PR is stacked on it).
- Previous: the SafenetGuard contract PR (the code under test).
- Next: the deployment-script PR.

### DIFF
--- a/contracts/test/SafenetGuard.t.sol
+++ b/contracts/test/SafenetGuard.t.sol
@@ -1,0 +1,1333 @@
+// SPDX-License-Identifier: GPL-3.0-only
+pragma solidity ^0.8.30;
+
+import {Test} from "@forge-std/Test.sol";
+import {ConsensusMessages} from "@/libraries/ConsensusMessages.sol";
+import {EpochRollover} from "@/libraries/EpochRollover.sol";
+import {FROST} from "@/libraries/FROST.sol";
+import {SafeTransaction} from "@/libraries/SafeTransaction.sol";
+import {Secp256k1} from "@/libraries/Secp256k1.sol";
+import {TransactionAnnouncement} from "@/libraries/TransactionAnnouncement.sol";
+import {AttestationTrailer} from "@/libraries/AttestationTrailer.sol";
+import {ISafenetGuard} from "@/interfaces/ISafenetGuard.sol";
+import {Enum} from "@safe/interfaces/Enum.sol";
+import {ISafe} from "@safe/interfaces/ISafe.sol";
+import {IGuardManager} from "@safe/interfaces/IGuardManager.sol";
+import {ITransactionGuard} from "@safe/base/GuardManager.sol";
+import {Safe} from "@safe/Safe.sol";
+import {SafeProxyFactory} from "@safe/proxies/SafeProxyFactory.sol";
+import {SafenetGuard} from "@/guard/SafenetGuard.sol";
+import {ForgeSecp256k1} from "@test/util/ForgeSecp256k1.sol";
+import {MockERC1271} from "@test/util/MockERC1271.sol";
+
+/**
+ * @title SafenetGuardTest
+ * @notice Behavioural tests for `SafenetGuard`. Every happy path drives a real Safe (singleton +
+ *         proxy factory) with the guard actually installed via `setGuard`, and exercises real
+ *         `execTransaction` calls carrying real ECDSA owner signatures. FROST attestations are real at
+ *         the on-chain verifier level — signatures are synthesized from a known aggregate secret via the
+ *         Schnorr equation (`ForgeSecp256k1`), not produced by a multi-party DKG/signing ceremony — so
+ *         the on-chain FROST/EC verification runs end-to-end, but validator interoperability is out of
+ *         scope here.
+ */
+contract SafenetGuardTest is Test {
+    using ForgeSecp256k1 for ForgeSecp256k1.P;
+
+    SafenetGuard public guard;
+
+    // ============================================================
+    // CONSTANTS — DEPLOYMENT
+    // ============================================================
+
+    uint256 public constant CONSENSUS_CHAIN_ID = 100; // Gnosis Chain
+    address public constant CONSENSUS_ADDR = address(0xC01115E1115115);
+    uint256 public constant ALLOW_TX_DELAY_SECONDS = 1 days;
+    uint256 public constant ALLOW_TX_WINDOW_SECONDS = 3 days;
+    uint64 public constant GENESIS_EPOCH = 1;
+    uint64 public constant ROLLOVER_BLOCK = 100;
+
+    address public other = address(0xABCDE);
+
+    // ============================================================
+    // CONSTANTS — KEY PAIRS
+    // Secret keys for ForgeSecp256k1.g(k): public key = k*G
+    // ============================================================
+
+    uint256 public constant GENESIS_SK = 1; // genesis group secret key; public key = G
+    uint256 public constant GENESIS_NK = 2; // genesis signing nonce; R = 2*G
+
+    uint256 public constant EPOCH2_SK = 5; // epoch 2 group secret key
+    uint256 public constant EPOCH2_NK = 6; // epoch 2 signing nonce
+
+    uint256 public constant FORK_SK = 7; // an alternative epoch-2 branch key
+    uint256 public constant FORK_NK = 8;
+
+    uint256 public constant UNKNOWN_SK = 9; // a key never recorded in the forest
+    uint256 public constant UNKNOWN_NK = 10;
+
+    // ============================================================
+    // CONSTANTS — DEFAULT TRANSACTION PARAMETERS
+    // ============================================================
+
+    address public constant TX_TO = address(0xBEEF);
+    uint256 public constant TX_VALUE = 0;
+    bytes public constant TX_DATA = hex"deadbeef";
+    Enum.Operation public constant TX_OP = Enum.Operation.Call;
+
+    // ============================================================
+    // SAFE DEPLOYMENT
+    // ============================================================
+
+    Safe public singleton;
+    SafeProxyFactory public factory;
+    ISafe public safe;
+    uint256 public ownerKey;
+
+    function setUp() public {
+        ownerKey = 0xA11CE;
+
+        singleton = new Safe();
+        factory = new SafeProxyFactory();
+
+        address[] memory owners = new address[](1);
+        owners[0] = vm.addr(ownerKey);
+        bytes memory initializer = abi.encodeCall(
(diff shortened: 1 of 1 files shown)


contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1385 @@
...
+
+    /// @dev Produces a packed ECDSA signature from the owner for a Safe transaction hash.
+    function _signSafeTx(bytes32 txHash) internal view returns (bytes memory) {
+        (uint8 v, bytes32 r, bytes32 s) = vm.sign(ownerKey, txHash);
+        return abi.encodePacked(r, s, v);
+    }
+
+    /// @dev Builds an inline FROST attestation trailer in the version-1 framing:
+    ///      `[192-byte abi.encode(epoch, groupKey, signature)][MAGIC | VERSION: 32 bytes]`.
+    ///      The group key is derived from `sk` so it matches the signing key.
```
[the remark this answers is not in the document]
remedcu: Updated 👍🏾

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1385 @@
...
+        // Keys ordered by derived address ascending (Safe requires this for packed multi-sigs).
+        uint256 keyA = 0xB2;
+        uint256 keyB = 0xC3;
+        uint256 keyC = 0xA1;
+
+        address[] memory owners = new address[](3);
+        owners[0] = vm.addr(keyA);
+        owners[1] = vm.addr(keyB);
+        owners[2] = vm.addr(keyC);
+
```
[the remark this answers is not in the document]
remedcu: Updated 👍🏾

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        if (mode == ExecMode.Attested) {
+            safeSig = bytes.concat(safeSig, _buildInlineAttestation(txHash, GENESIS_EPOCH, GENESIS_SK, GENESIS_NK));
+        }
+        safe.execTransaction(to, value, data, op, 0, 0, 0, address(0), payable(address(0)), safeSig);
+    }
+
+    /// @dev Like _execSafeTx(Direct) but takes a pre-computed nonce, making safe.execTransaction the
+    ///      only external call. Use when vm.expectRevert is set immediately before, so the nonce read
+    ///      does not consume the cheat code.
+    function _execSafeTxWithNonce(address to, uint256 value, bytes memory data, Enum.Operation op, uint256 nonce)
```
rmeissner, line 197: nit: feels weird to have 3 different exec method, which could probably all be covered by the first one.

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        vm.expectRevert(ISafenetGuard.AttestationNotFound.selector);
+        _execSafeTxWithNonce(TX_TO, TX_VALUE, TX_DATA, TX_OP, nonce);
+    }
+
+    function test_announcement_consumedOnUse() public {
+        bytes32 h = _defaultAnnouncementHash();
+        _announce(_defaultAnnouncement());
+        vm.warp(block.timestamp + ALLOW_TX_DELAY_SECONDS);
+        _execSafeTx(TX_TO, TX_VALUE, TX_DATA, TX_OP, ExecMode.Direct);
+        assertEq(_announcedActiveFrom(h), 0);
```
rmeissner, line 503: nit: maybe worth showing that a second exec safe tx call reverts.

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        );
+        _announce(_defaultAnnouncement());
+    }
+
+    function test_announceTransaction_revertsOnDuplicate() public {
+        _announce(_defaultAnnouncement());
+        uint256 nonce = safe.nonce();
+        bytes memory data = abi.encodeCall(SafenetGuard.announceTransaction, (_defaultAnnouncement()));
+        vm.expectRevert(TransactionAnnouncement.AnnouncementAlreadyExists.selector);
+        _execSafeTxWithNonce(address(guard), 0, data, Enum.Operation.Call, nonce);
```
rmeissner, line 529: nit: we should be able to just call `_announce(_defaultAnnouncement());` again, right?

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        _cancelAnnouncement(h); // no warp — cancellation is not delayed
+        assertEq(_announcedActiveFrom(h), 0);
+
+        vm.warp(block.timestamp + ALLOW_TX_DELAY_SECONDS);
+        uint256 nonce = safe.nonce();
+        vm.expectRevert(ISafenetGuard.AttestationNotFound.selector);
+        _execSafeTxWithNonce(TX_TO, TX_VALUE, TX_DATA, TX_OP, nonce);
+    }
+
+    function test_integration_announcementFullFlow() public {
```
rmeissner, line 611: nit: this feels like a duplicated test, as in this form more or less it has been tested before

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        uint256 keyB = 0xC3;
+        uint256 keyC = 0xA1;
+
+        // Sort owners ascending by address so the owner set is well-formed regardless of which
+        // addresses the hard-coded keys derive to; `_sign2of3` independently orders the signatures.
+        address[] memory owners = new address[](3);
+        owners[0] = vm.addr(keyA);
+        owners[1] = vm.addr(keyB);
+        owners[2] = vm.addr(keyC);
+        for (uint256 i = 0; i < owners.length; i++) {
```
rmeissner: nit: we theoretically know the address to each key (this is deterministic), so no need to sort each time ... precalculate this once.

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        assertFalse(guard.supportsInterface(0xdeadbeef));
+    }
+
+    // ============================================================
+    // SCOPE PREMISE — NO MODULES
+    // ============================================================
+
+    /// @notice This guard is transaction-guard only; the test Safe must have no modules enabled, so the
+    ///         suite genuinely reflects the "modules out of scope" security premise (see DD / F-05).
+    function test_setup_noModulesEnabled() public view {
```
rmeissner, line 1013: nit: not sure if this test is meaningful

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        (address[] memory modules,) = Safe(payable(address(safe))).getModulesPaginated(address(0x1), 10);
+        assertEq(modules.length, 0);
+    }
+
+    // ============================================================
```
rmeissner, line 1018: nit: we have constructor related tests at the beginning (https://github.com/safe-research/safenet/pull/605/changes#diff-c0435e5b4a3a95d6c98a5b2e9f8378c2aa9a57f311b9629e3d592862da536562R292-R294)  and now here again

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        _rollover(EPOCH2_SK, EPOCH2_NK, GENESIS_EPOCH + 1, GENESIS_EPOCH + 2, FORK_SK);
+        // The genesis key still attests a brand-new transaction at the current nonce.
+        _execAttestedWith(GENESIS_EPOCH, GENESIS_SK, GENESIS_NK); // must not revert
+    }
+
+    /// @notice Documents the accepted policy: a historical parent can sign a brand-new far-future
+    ///         rollover branch, which is then usable for attestation.
+    function test_forest_historicalParentCreatesFutureBranch() public {
+        _rollover(GENESIS_SK, GENESIS_NK, GENESIS_EPOCH, GENESIS_EPOCH + 1, EPOCH2_SK);
+        Secp256k1.Point memory branchKey = _rollover(GENESIS_SK, GENESIS_NK, GENESIS_EPOCH, 1000, UNKNOWN_SK);
```
rmeissner, line 1073: nit: feels weird to call the new key "unknown" here, as it is not unknown ... we did a proper rollover to it. Also it is not in a "far future" rollover branch, as it is the direct child. Both are only wording things, so no need to adjust the logic of the test.

contracts/test/SafenetGuard.t.sol
```diff
@@ -0,0 +1,1391 @@
...
+        _execAttestedWith(1000, UNKNOWN_SK, UNKNOWN_NK); // must not revert
+    }
+
+    // ============================================================
+    // SAFE-HASH PARITY (adapter correctness)
+    // ============================================================
+
+    /// @notice The guard-side attestation hash must equal Safe's own `getTransactionHash` across all
+    ///         parameters, including the gas/refund fields our other tests leave zero.
+    function testFuzz_safeHashParity(
```
rmeissner: request for change (will still approve, but imo that should be changed): this does not belong here. It doe not test anything related to the guard (i.e. it does not call the guard).

rmeissner (approved):
This PR is hard to review and would love that we improve this in the future:
- A big PR with many tests that imo can easily be split up
- Ordering of tests and methods is weird (there some grouping but it is not consistent)
- Tests that are not relevant to the code

Anyways, generally there is nothing blocking this PR.

contracts/test/SafenetGuard.t.sol
remedcu: Removed 👍🏾

PR #608, opened 2026-07-21
[Phase 3] Calldata decoding utilities

rmeissner:
Add utilities to decode token and multi send transactions. This allows to detect allowances and transfers (including batched ones).

The list of known multi send deployments has been moved to the multisend mod for a centralized maintainability.

The current logic is very simple and tries to decode using different ABIs without regarding known token addresses. This should be improved in the future.

---

## CLA signature

### DIFF
--- a/crates/safe-tx/src/target_effects.rs
+++ b/crates/safe-tx/src/target_effects.rs
@@ -1,0 +1,459 @@
+//! Pure calldata decoding: extracts the address a Safe transaction (or, when
+//! it's a MultiSend batch, each of its sub-calls) sends value to, or grants
+//! transfer/spending authority over tokens to. No policy decisions are made
+//! here — see `crate::checks` for the existing base-guarantee policy, and
+//! future `Check`s (elsewhere) that will consume this output.
+
+use crate::multi_send::{decode_multi_send, known_deployment};
+use crate::types::{Operation, SafeTransaction, erc20, erc721, erc1155, multi_send_bindings};
+use alloy::{
+    primitives::{Address, U256},
+    sol_types::SolCall as _,
+};
+
+/// A decoded value- or authority-granting effect that a Safe transaction (or
+/// one of its MultiSend sub-calls) has on a target address.
+#[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


crates/safe-tx/src/target_effects.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// The address receiving value, tokens, or approved spending/transfer
+    /// authority.
+    pub recipient: Address,
+    pub kind: EffectKind,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum EffectKind {
+    /// A native value transfer, or an ERC-20 `transfer`/`transferFrom`.
+    ValueTransfer { amount: U256 },
```
remedcu, line 27: Question: Why are we conflating these two?

crates/safe-tx/src/target_effects.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// An ERC-20 `approve`. Note: ERC-721's single-token
+    /// `approve(address,uint256)` has the exact same selector and is
+    /// indistinguishable from calldata alone — it decodes the same way.
+    Erc20Approval { amount: U256 },
```
remedcu, line 31: nit: As it is technically both Token approval, should this be:

```suggestion
    TokenApproval { amount: U256 },
```

remedcu (approved):
A question and a nit, rest LGTM 👍🏾

crates/safe-tx/src/target_effects.rs
rmeissner: To be honest I have not looked at this in detail, as I expect a lot of change in the next weeks (after this has been merged). This initial implementation only checks for "near max" uint256 values, so we do not need to differentiate them.

crates/safe-tx/src/target_effects.rs
rmeissner: Probably makes sense, especially under the premise that we only want to check for "near max values". As mention in the other comment this is code that we will have to iterate on in the next weeks to some extend, so  need to check how to best handle this here.

PR #610, opened 2026-07-21
[Phase 5a]: Add dynamic check effect and resume

rmeissner:
To allow dynamic checks a explicit split from the current static checks is implemented. Dynamic checks are performed via the effect/ resume logic, to avoid async handling in state transitions.

To make this paradigm clear the detector has been renamed to static checker. The initial dynamic checker defers the check to a remote endpoint (if configured).

This initial remote checker will only trigger a vote if a clear approve or deny is returned from the remote endpoint (so with a reason). If there is any error/ timeout the sentinel will abstain from voting to avoid losing the bond.

---

## CLA signature


crates/sentinel/src/dynamic_checker.rs
```diff
@@ -0,0 +1,189 @@
...
+        Ok(
+            match (
+                response.approve,
+                response.rule.as_deref().map(RuleId::from_code),
+            ) {
+                (true, _) => RemoteCheckOutcome::Approved,
+                (false, Some(Some(rule))) => RemoteCheckOutcome::Denied(rule),
+                (false, _) => {
+                    tracing::error!(%safe, "remote check denied without a recognized rule code");
+                    RemoteCheckOutcome::Failed
```
remedcu, line 102: Question: Why`false` approval with an unrecognized reason failed, and not denied?

remedcu (approved):

crates/sentinel/src/dynamic_checker.rs
rmeissner: Because if we deny then there is a chance to lose our bond. We should be sure that we know why we deny or approve, otherwise we should not participate in this vote.

Edit: will add this to the PR description so that it is part of the commit message

crates/sentinel/src/dynamic_checker.rs
```diff
@@ -0,0 +1,189 @@
...
+    /// `url: None` means no remote check is configured; every call then
+    /// resolves to [`RemoteCheckOutcome::Approved`] without a request, so
+    /// the reference Sentinel works with just its local checks until an
+    /// operator opts into a remote one.
+    pub fn new(url: Option<Url>) -> Self {
+        Self {
+            url,
+            client: reqwest::Client::new(),
+        }
+    }
```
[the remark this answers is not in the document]
rmeissner: This will have to be properly setup in a future phase

PR #611, opened 2026-07-22
[Guard - Step 8] Add SafenetGuard README

remedcu:
Documents the canonical SafenetGuard, split out of the contract PR (#604) as requested.

### Context and Motivation

Adds a `SafenetGuard` section to the existing guard README: self-contained (no comparison framing against the legacy guards), documenting the epoch forest, the signature-extension attestation trailer, the nonce-free escape hatch, the bounded window, and the accepted design decisions. It also records the type-hash-framed signature extension as the first such pattern the Safe project publishes and a reusable convention for future extensions.

### Decisions and Tradeoffs

- Kept as a separate PR so README iteration never blocks the contract.
- The `SafenetGuardA` / `SafenetGuardB` sections are left untouched; removing them is future cleanup alongside deleting those contracts.
- Design rationale is phrased as decisions taken, not pros/cons.

### Testing

- Docs only.

### DIFF
--- a/contracts/src/guard/README.md
+++ b/contracts/src/guard/README.md
@@ -1,6 +1,100 @@
 # Safenet Guards
 
-This directory contains a collection of Safe Guard implementations for Safenet. Each variant explores a different design trade-off in how FROST threshold-signature attestations are delivered, verified, and stored on-chain. The goal is to evaluate approaches across dimensions such as gas cost, tooling compatibility, epoch management, and escape-hatch ergonomics before settling on a production design.
+`SafenetGuard` is the canonical Safe Guard for Safenet. `SafenetGuardA` and `SafenetGuardB` are earlier explorations, retained for comparison and slated for removal once `SafenetGuard` is adopted.
+
+---
+
+## SafenetGuard
+
+**File:** `SafenetGuard.sol`
+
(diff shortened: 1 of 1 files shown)


contracts/src/guard/README.md
```diff
@@ -1,6 +1,100 @@
...
+```
+
+Anchoring the extension at the end leaves Safe's front-to-back signature parser untouched, and the terminal type hash makes detection independent of signature suffixes — only a blob whose last word equals `TYPE_HASH` is treated as an attestation, so a valid Safe signature ending in an unrelated value (even the number 192) is never mis-parsed. The type hash doubles as an extensible format tag: it embeds the version, so a future format uses a different type hash (which this guard reads as "no trailer"). This is the first canonical guard the Safe project publishes to use signature extension, and the type-hash separator is intended as a reusable convention for future extensions.
+
+`checkTransaction` outcomes: no type hash → falls through to the announcement path; a type hash on a too-short blob → reverts `MalformedAttestationTrailer`; a recognised trailer → the `(groupKey, epoch)` pair must be trusted (else `UntrustedAttestationKey`) and the FROST signature is verified. A recognised trailer never silently falls through. Forest membership already implies a non-zero key, so no separate non-zero check is needed.
```
rmeissner, line 35: uber nit: maybe call it "no signature type hash" (do clearly differentiate it from our other type hashes)

rmeissner (approved):

PR #618, opened 2026-07-24
Safenet Guard

remedcu:
Adds SafenetGuard: a standalone Safe transaction guard that gates every owner-signed `execTransaction` behind a Safenet FROST threshold-signature attestation, with a nonce-free, time-windowed escape hatch for liveness.

### Context and Motivation

This lands the reviewed `standalone-guard` stack as a single change. The feature was built as a linear stack of small, individually reviewed PRs - TransactionAnnouncement library (#600), ISafenetGuard interface (#601), AttestationTrailer library (#602), the SafenetGuard contract (#604), its test suite (#605), the deploy script (#606), and the README (#611) - so each concern had a focused review surface.

How it works:
- **Attestation via signature extension.** The FROST attestation `(epoch, groupKey, signature)` rides as a trailer appended to Safe's `signatures`, framed by a terminal type hash (`keccak256("SafenetGuard.AttestationTrailer.v1")`) that also tags the format version. `checkTransaction` verifies it against the full nonce-bound Safe tx hash. This is the first canonical guard the Safe project publishes to use signature extension.
- **Epoch forest.** Trusted `(group key, epoch)` pairs are tracked via the shared `EpochRollover` library; rollovers are permissionless (the FROST signature is the authorization).
- **Nonce-free escape hatch.** `announceTransaction` queues a transaction by its parameters minus the nonce; after a fixed delay it executes without an attestation inside a bounded `[activeFrom, activeUntil]` window, at any nonce. Single-use, revocable immediately via `cancelAnnouncement`.

### Decisions and Tradeoffs

- **Scope is transaction-guard only** - Safe module executions bypass it. Deployments must forbid modules or treat each as an explicit bypass.
- `ISafenetGuard` extends `ITransactionGuard`, so the guard hooks are part of the published ABI.
- The forest keeps historical `(groupKey, epoch)` pairs forever: post-rollover FROST shares are destroyed, so retaining old keys is not a practical risk (a historical key cannot replay past transactions - the Safe nonce binds them).
- Announcing emits a single `TransactionAnnounced` event; the full parameters are recoverable from the announce calldata.
- The structural self-call gate for the escape-hatch functions is kept inline in the guard for explicitness (the earlier `GuardAutoAllow` library was folded in, #603).
- SafenetGuard requires `via_ir` (added to `foundry.toml`); full rationale and accepted design decisions live in `src/guard/README.md`.

### Testing

- `cd contracts && forge test` - full suite green (SafenetGuard: 65 tests with a real Safe and real FROST signing; TransactionAnnouncement: 8 tests).
- `forge build`, `forge fmt --check`, and `forge lint --deny notes` pass.

### DIFF
--- a/contracts/.env.sample
+++ b/contracts/.env.sample
@@ -4,7 +4,7 @@
 # 2 = Canonical Deterministic Deployment Factory
 FACTORY=1
 
-# SafenetCosigner Deployment Config
+# SafenetCosigner / SafenetGuard Deployment Config (shared)
 # Chain ID of the Consensus contract (100 = Gnosis Chain)
 CONSENSUS_CHAIN_ID=100
 # Address of the Consensus contract on the consensus chain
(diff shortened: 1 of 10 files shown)


rmeissner (approved):
note: I did not re-review the changes, as these should be what has been reviewed in the stack

PR #619, opened 2026-07-24
[Phase 6]: Implement initial address poisoning check

rmeissner:
The check is implemented as a dynamic check based on external information (i.e. RPC endpoint). This version only check if a previous valid transfer exists. No explicit denial yet.

---

## CLA signature


crates/sentinel/src/effect.rs
```diff
@@ -59,27 +70,46 @@ impl EffectHandler<Effect, Resume> for Handler {
...
                 transaction,
                 deadline,
-            } => Resume::DynamicCheckResult {
-                request_id,
-                deadline,
-                outcome: self.checker.check(safe, &transaction).await,
-            },
+            } => {
+                let outcome = match self.address_poisoning.check(safe, &transaction).await {
+                    Verdict::Approved => RemoteCheckOutcome::Approved,
```
remedcu, line 75: Question: What happens if the prior interaction turned bad recently and the dynamic check would have found that malicious intent, but skipped in this case?

remedcu (approved):

crates/sentinel/src/effect.rs
rmeissner: The current logic is not flexible/ good enought for this. It is something that needs to be covered by an advanced logic. Hopefully we get to a point that we can input all of this into this public reference sentinel service, as this would put some pressure on the other sentinels to provide good checks

Issue #614, opened 2026-07-23
Evaluate Parallel Execution of Effects


rmeissner:
Effects are currently executed while blocking the progression of the state transition. It could be considered to allow them to be non-blocking. This would enable longer running sentinel checks that do not block the state transition. 

In general signing requests (so related to proposed transactions) can run in parallel, but this is currently not supported by our stack.

# References

- https://github.com/safe-research/safenet/pull/613

Gemini comment:

```
1. Head-of-Line Blocking / Event-Loop Stalling
The implementation explicitly avoids creating a WaitingForDynamicCheck state by relying on the assumption that TransitionBatch::apply "fully resolves an emitted effect's resume before the next message is applied."

If this assumption is true: The state machine blocks synchronously while waiting for the remote dynamic check (HTTP request) to complete. This introduces a Head-of-Line (HoL) blocking vulnerability. An attacker can spam proposals, and if the remote checker takes 2–3 seconds per request, the node will stop processing new blocks, miss consensus events, and fall completely out of sync.

If this assumption is false (effects are async): The node will suffer a severe race condition. A NewRequest event arriving while the dynamic check is still pending will be entirely ignored because state.0.get_mut(&event.requestId) will return None. The node will never vote on the proposal.

Recommendation: You must introduce a WaitingForDynamicCheck variant to RequestState. When the dynamic check effect is emitted, transition the request into this state. This allows the event loop to remain non-blocking while securely queuing or handling events (like NewRequest) that arrive before the dynamic check HTTP request resolves.
```

nlordell:
done

PR #620, opened 2026-07-24
[Epic] Non-Blocking Effects

rmeissner:
Adds an epic to allow non-blocking background execution of tasks (in parallel to the state transition). This is important for the dynamic checks and relates to #614

---

## CLA signature

### DIFF
--- a/epics/2026_07_24_nonblocking_effects.md
+++ b/epics/2026_07_24_nonblocking_effects.md
@@ -1,0 +1,115 @@
+# Plan: Non-Blocking Effects
+
+Component: `crates/core` (state machine, driver), `crates/sentinel` (Rust).
+
+---
+
+## Overview
+
+Every effect a [`StateTransition`](../crates/core/src/state/mod.rs) emits today runs fully in-band and blocking: `TransitionBatch::apply` (`crates/core/src/state/mod.rs:313-339`) `await`s `EffectHandler::perform_effect` directly, and the whole call chain — from `StateMachine::handle_update` down — runs under `&mut self`/a single `Mutex` guard held for the duration (`mod.rs:196-286`). A slow effect stalls everything: no other log in the current batch, no later block, nothing.
+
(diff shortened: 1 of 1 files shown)


epics/2026_07_24_nonblocking_effects.md
```diff
@@ -0,0 +1,115 @@
...
+| Phase | Summary | Depends on | Own PR |
+|---|---|---|---|
+| 1 | Core: `Command::Background`/`Message::BackgroundResume`, `BackgroundEffectHandler` trait, `StateMachine` spawn/`JoinSet`/resume-fetch plumbing | — | ✅ |
+| 2 | Core: wire background resumes into `Driver::step`; `Service` trait gains `Background`; validator + sentinel opt in with a no-op handler (zero behavior change) | 1 | ✅ |
+| 3 | Sentinel: migrate `Effect::DynamicCheck` to `BackgroundEffectHandler`; add `WaitingForDynamicCheck`/`PreparingToVote`; fix the `NewRequest`-vs-check race in `handle_new_request`/`handle_dynamic_check_result`; deadline-sweep arms | 2 | ✅ |
+| 4 | Remove this plan | 3 | ✅ |
+
+---
+
+## Assumptions
```
remedcu, line 110: nit: A safe assumption we can add:

```
A background check's `(approve, reason)` is not snapshotted before its `Commit` is broadcast, so it must be reproducible for a request within its commit/reveal window: if the process crashes between broadcasting the `Commit` and the next snapshot, the only way to later reveal a value matching the on-chain commitment is to re-derive the same decision on restart (`salt` is already deterministic; `approve`/`reason` are not persisted). This holds for the deterministic/near-static built-in checks, but not for a time-varying `RemoteChecker`, where a flipped verdict would reveal a mismatch (`InvalidReveal` → forfeited bond).

Accepted as a low-probability, scoped risk.
```

remedcu (approved):

epics/2026_07_24_nonblocking_effects.md
rmeissner: But they are snapshotted in the current implementation. See https://github.com/safe-research/safenet/blob/750c5407e3212d86acfdffaeaf164ba80cb6e13c/crates/sentinel/src/state.rs#L15-L16

PR #621, opened 2026-07-24
[Phase 7a]: Checker Trait

rmeissner:
Introduce the checker trait to allow a composition of checkers and an easy introduction of new check logic. The CheckOutcome will replace the RemoteCheckOutcome and the Verdict to simplify the code.

---

## CLA signature


crates/sentinel/src/checker.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// No checker so far denied the transaction.
+    Approved,
+    /// This checker denied the transaction, citing this rule.
+    Denied(RuleId),
+    /// This checker has nothing conclusive to say — including a lookup
+    /// failure (e.g. an unreachable RPC/HTTP endpoint), which can't be
+    /// trusted as evidence either way. Move on to whichever checker runs
+    /// next; if every checker in the chain resolves to this, the request is
+    /// dropped unanswered rather than guessed at.
+    Unknown,
```
remedcu, line 33: Question: Why `Failed` and `NoOpinion` was merged together? I see the comment `can't be trusted as evidence either way`, but both those options provide different opinion on the check IMO.

remedcu (approved):

crates/sentinel/src/checker.rs
rmeissner: Because ultimately it results in the same outcome -> don't vote/ don't participate in the game.

For handling errors this should be handled before it is passed back to the "game logic".

Would you see that you perform different actions based Failed or NoOpinion?

PR #626, opened 2026-07-28
[Phase 8d]: Validate CoW api response

rmeissner:
Add a check where the order id of the calldata is compared against the order id calculated based on the api response. This will flag if invalid data is returned by the api. In this case the check will return UNKNWON, as the presinged order cannot be validate.

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/cow.rs
+++ b/crates/sentinel/src/cow.rs
@@ -58,9 +58,9 @@
 
 use crate::checker::{CheckOutcome, Checker};
 use alloy::{
-    primitives::{Address, Bytes, U256, address},
+    primitives::{Address, B256, Bytes, U256, address},
     sol,
-    sol_types::{SolCall, SolValue},
+    sol_types::{Eip712Domain, SolCall, SolStruct, SolValue, eip712_domain},
 };
 use safe_tx::{
     multi_send::decode_multi_send_call,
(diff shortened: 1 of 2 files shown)


crates/sentinel/src/cow.rs
```diff
@@ -78,6 +101,53 @@ fn order_api_base_url(chain_id: U256) -> Option<&'static str> {
...
+        name: "Gnosis Protocol",
+        version: "v2",
+        chain_id: chain_id,
+        verifying_contract: GP_V2_SETTLEMENT,
+    }
+}
+
+/// Recomputes `order`'s 56-byte `orderUid` (`orderDigest(32) ‖ owner(20) ‖
+/// validTo(4)`, exactly how `GPv2Order.sol`'s `packOrderUidParams` and
+/// EIP-712 `hash` combine) for `chain_id`'s domain — the cryptographic tie
```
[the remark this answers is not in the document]
rmeissner: for the hash calculation the null value (so address zero) should be used. The replacement with the owner address only happens later (when executing the trade).

crates/sentinel/src/cow.rs
```diff
@@ -131,6 +154,53 @@ fn order_api_base_url(chain_id: U256) -> Option<&'static str> {
     }
 }
 
+/// GPv2Settlement's EIP-712 domain. `verifyingContract` is the same address
+/// on every network it's deployed to (deterministic `CREATE2`), so only
+/// `chain_id` varies. Chain ids of the networks CoW supports (see
+/// `order_api_base_url`) are all well within `u64` range, which is all the
+/// `eip712_domain!` macro accepts.
+fn gp_v2_domain(chain_id: U256) -> Eip712Domain {
+    let chain_id = u64::try_from(chain_id).expect("chain id expected to be in u64 range");
```
remedcu, line 163: Question: Not an issue right now as we have limited chains supported, but should this return `UNKNOWN` if not within those supported chains?

remedcu (approved):

crates/sentinel/src/cow.rs
rmeissner: It does already:
- https://github.com/safe-research/safenet/blob/7e9723c3f0c0232ee92607157fd78bba95cc4491/crates/sentinel/src/cow.rs#L315-L317
- https://github.com/safe-research/safenet/blob/7e9723c3f0c0232ee92607157fd78bba95cc4491/crates/sentinel/src/cow.rs#L438-L443

PR #635, opened 2026-07-28
[Guard] Extract design decisions into DESIGN.md; flatten guard README

remedcu:
Moves the accepted design decisions into a dedicated `src/guard/DESIGN.md` (rmeissner's suggestion) and flattens the now single-guard README.

### Context and Motivation
`DESIGN.md` holds only SafenetGuard's decisions (no legacy-guard/cosigner references); the README links to it and drops the redundant collection-level heading.

### Testing
- Docs only.

### DIFF
--- a/contracts/src/guard/DESIGN.md
+++ b/contracts/src/guard/DESIGN.md
@@ -17,7 +17,7 @@ Deliberate choices with their rationale, recorded so reviewers and auditors can 
 
 - **Inline signature-extension trailer carrying `epoch + groupKey + signature` explicitly** (the forest has no epoch→key reverse lookup). The non-standard encoding (wallets must be Safenet-aware) and the larger trailer are accepted trade-offs.
 - **Type-hash trailer framing** (`keccak256("SafenetGuard.AttestationTrailer.v1")`, not length-only) so a valid Safe-signature suffix cannot be mis-parsed; a recognised trailer never falls through to the announcement path.
-- **Replay/ordering come from the Safe nonce** bound into the verified hash; there is no spent-signature registry.
+- **Replay/ordering come from the Safe nonce** bound into the FROST-verified attestation message (over the nonce-bound Safe tx hash); there is no spent-signature registry. The escape-hatch path is instead single-use and time-boxed.
 
 ## Escape hatch (announcements)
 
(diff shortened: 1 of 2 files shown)


contracts/src/guard/DESIGN.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Escape hatch (announcements)
+
+- **`announceTransaction` takes the full parameter struct** (not a bare hash) — signers see what they authorize and the guard derives the hash on-chain (so it cannot diverge from `checkTransaction`).
+- **Nonce-free announcement hash** (excludes the Safe nonce and the Safe address; scoped by storage key) — keeps the hatch usable while unrelated transactions advance the Safe nonce.
+- **Bounded, inclusive `[activeFrom, activeUntil]` window**, both durations immutable; packed into one slot (two `uint128`), with `WindowOverflow`/constructor bounds preventing absurd values.
+- **Single-use; expired entries are renewable in place**; pending/active ones cannot be overwritten; `cancelAnnouncement` is immediate.
+- **Consumption tracks the authorization path, not execution success** (a caught inner-call failure with non-zero `safeTxGas`/`gasPrice` still consumes the announcement). A lock/finalize/restore state machine was considered and deferred.
+- **A relayer holding both a valid attestation and a matured announcement can choose the path**; the attestation path takes precedence and does not consume the announcement.
```
rmeissner, line 29: question: interesting, wasn't aware of this. What is the rational that the attestation path takes precendence? Without thinking too much about it, I would have though that consuming the announcement is more secure.

rmeissner (approved):

contracts/src/guard/DESIGN.md
remedcu: The intent is that the two paths are **mutually exclusive**: an announcement is the fallback for when there's *no* attestation, so a tx that carries a valid attestation is the expected happy path and shouldn't touch the announcement store at all.

Three reasons we don't consume it:

1. **Expected vs. exception (gas).** Attestation is the common path; announcements are the exception. Reaching into the announcement store (hash + `SLOAD`/delete) on *every* attested tx just to catch the rare overlap would add gas to the hot path.
2. **The lingering announcement is bounded.** Even un-consumed, it's time-delayed (`activeFrom`), time-bound (`activeUntil`), single-use, and immediately cancellable. So a possible repeat of the same params has a limited window and the owners can revoke it at any time.
3. **Mutually exclusive by design.** An announcement is meant for when attestation isn't available; using both for one tx isn't an expected state, so we keep the two owner-controlled channels independent rather than coupling them.

PR #643, opened 2026-07-29
[Phase 3.2]: Generate Sentinel Configs

rmeissner:
Generate the toml config files for the sentinels. This requires to calculate the deployment addresses before the pod is started. The logic for this was present, but not reusable, therefore new helper methods are introduced.

---

## CLA signature

### DIFF
--- a/scripts/run_devnet.sh
+++ b/scripts/run_devnet.sh
@@ -31,6 +31,20 @@ ARBITRATOR=0x9965507D1a55bcC2695C58ba16FB37d819B0A4dc
 # Private key: 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
 DEPLOYER=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
 
+# SentinelOracleV2 economics (see the epic's Architecture Decision): a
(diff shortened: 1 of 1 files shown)


scripts/run_devnet.sh
```diff
@@ -237,6 +279,40 @@ for validator in "${VALIDATORS[@]}"; do
...
+    {
+        echo "rpc = \"http://localhost:8545\""
+        echo "signer = \"${private_key}\""
+        echo "database = \"sqlite::memory:\""
+        echo "oracle = \"${sentinel_oracle}\""
+        echo "consensus = \"${consensus}\""
+        echo
+        echo "[sentinel]"
+        echo "fee_token = \"${fee_token}\""
+        echo "voting_window = 1"
```
remedcu, line 333: nit: As per epic
```suggestion
        echo "voting_window = 10"
```

remedcu (approved):

scripts/run_devnet.sh
rmeissner: That was explicitly changes, as the epic was not using a meaningful value. The naming is not 100% correct here, as this is more a request_timeout. But it is outside of the scope of this PR to adjust this. The request timeout can be small, as in the contract this always happens in the same block. So this is more a cleanup mechanism.

PR #662, opened 2026-08-03
[Phase 2]: Add support for oracle consensus events

rmeissner:
Adjust the proposal and attestation events to also cover the events with oracle information. The current version support both proposal paths. The detailed tracking of oracle events (including sentinel votes) will happen in a separate PR.

---

## CLA signature

### DIFF
--- a/explorer/src/lib/consensus/transactions.ts
+++ b/explorer/src/lib/consensus/transactions.ts
@@ -2,6 +2,7 @@ import {
 	type Address,
 	formatLog,
 	getAbiItem,
+	getAddress,
 	type Hex,
 	numberToHex,
 	type PublicClient,
@@ -39,6 +40,7 @@ export type TransactionProposal = {
 	chainId: bigint;
 	safeTxHash: Hex;
 	epoch: bigint;
+	oracle: Address | null;
 	transaction: SafeTransaction;
 	proposedAt: ExecutionLink;
 	attestedAt: ExecutionLink | null;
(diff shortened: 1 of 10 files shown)


explorer/src/lib/consensus/transactions.ts
```diff
@@ -124,16 +128,34 @@ export const loadTransactionProposals = async ({
 		}),
 	);
 
-	const attestationKey = (log: { args: { safeTxHash: Hex; epoch: bigint } }) =>
-		`${log.args.safeTxHash}:${log.args.epoch}`;
+	// `oracle` isn't an indexed topic on either oracle event, so it can't be filtered via
+	// eth_getLogs topics; trust is resolved here, after decoding. With no explicit allow-list,
```
nlordell, line 132: This feels fragile, shouldn't the `oracle` address become indexed?

nlordell (approved):
Approving the explorer changes themselves, but I do think we should _separately_ make `oracle` and indexed event field if possible.

explorer/src/lib/consensus/transactions.ts
rmeissner: we cannot index any additional field, as we already index 3 fields. We would have to remove one of the existing

nlordell: We should consider it IMO.

rmeissner: https://github.com/safe-research/safenet/issues/669

PR #663, opened 2026-08-03
[Phase 3]: Oracle events data layer

rmeissner:
This PR only implements the data layer (hashing, abi, status loading and worker logic) for oracle based events and proposals. This PR does not include any UI adjustments.

---

## CLA signature

### DIFF
--- a/explorer/src/lib/oracle/index.ts
+++ b/explorer/src/lib/oracle/index.ts
@@ -1,0 +1,13 @@
+export * from "./abi";
+export * from "./hashing";
+export * from "./votingStatus";
+
+import { type Remote, wrap } from "comlink";
+import type { OracleWorkerApi } from "./worker";
+
+let instance: Remote<OracleWorkerApi> | undefined;
+
+export function getOracleWorker(): Remote<OracleWorkerApi> {
+	instance ??= wrap<OracleWorkerApi>(new Worker(new URL("./worker.ts", import.meta.url), { type: "module" }));
+	return instance;
+}
--- a/explorer/src/lib/oracle/votingStatus.ts
+++ b/explorer/src/lib/oracle/votingStatus.ts
@@ -1,0 +1,58 @@
+import { type Address, getAbiItem, type Hex, type PublicClient, zeroAddress } from "viem";
+import { getBlockRange, mostRecentFirst } from "@/lib/utils";
+import { oracleAbi, sentinelOracleV2Abi } from "./abi";
+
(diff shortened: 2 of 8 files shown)


explorer/src/lib/oracle/votingStatus.ts
```diff
@@ -0,0 +1,58 @@
...
+		});
+		if (request.proposer === zeroAddress) return null;
+		return {
+			kind: "sentinel",
+			state: sentinelRequestStates[request.state],
+			approveCount: request.approveSentinelCount,
+			denyCount: request.denySentinelCount,
+		};
+	} catch {
+		const { fromBlock, toBlock } = await getBlockRange(provider, maxBlockRange);
```
nlordell, line 44: nit: do we need to support the "old" oracle at all? I would avoid adding extra code that we don't need for the final version.

explorer/src/lib/oracle/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export * from "./hashing";
+export * from "./votingStatus";
+
+import { type Remote, wrap } from "comlink";
+import type { OracleWorkerApi } from "./worker";
+
+let instance: Remote<OracleWorkerApi> | undefined;
+
+export function getOracleWorker(): Remote<OracleWorkerApi> {
+	instance ??= wrap<OracleWorkerApi>(new Worker(new URL("./worker.ts", import.meta.url), { type: "module" }));
```
nlordell, line 11: request change: Huh... making this a worker does not make sense at all to me. None of this is CPU-bound code right?

nlordell (changes requested):

explorer/src/lib/oracle/votingStatus.ts
rmeissner: it not about an old oracle, but about an unknown oracle. I would support this here, as we might want to test different oracles

explorer/src/lib/oracle/index.ts
rmeissner: Same as [here](https://github.com/safe-research/safenet/pull/664#discussion_r3711138724). This is the current pattern, but something to reevaluate in more details: #670

nlordell (approved):

explorer/src/lib/oracle/votingStatus.ts
nlordell: I'm personally not a fan of the unused code - but don't feel strongly about it.

PR #664, opened 2026-08-03
[Phase 4]: Implement detailed vote fetching

rmeissner:
Implement fetching of votes based on events to display voting details (i.e. reason and submission status).

---

## CLA signature

### DIFF
--- a/explorer/src/lib/oracle/votes.ts
+++ b/explorer/src/lib/oracle/votes.ts
@@ -1,0 +1,64 @@
+import { type Address, formatLog, type Hex, numberToHex, type PublicClient, parseEventLogs } from "viem";
+import { getBlockRange } from "@/lib/utils";
+import { sentinelOracleV2Abi, sentinelVoteEventSelectors } from "./abi";
+
+export type SentinelVote =
+	| { sentinel: Address; state: "committed" }
+	| { sentinel: Address; state: "approved" | "denied"; reason: string };
+
+// Per-sentinel breakdown for a `SentinelOracleV2` request, discovered from indexed `Committed`/
+// `Revealed` logs rather than a configured roster — only sentinels who acted appear, ordered by
+// their `Committed` log position (first vote first). Callers should only invoke this once
+// `loadVotingStatus` has confirmed `kind === "sentinel"`; a generic `IOracle` has no such events.
+export const loadSentinelVotes = async ({
+	provider,
+	oracle,
+	requestId,
+	maxBlockRange,
+}: {
+	provider: PublicClient;
+	oracle: Address;
+	requestId: Hex;
+	maxBlockRange: bigint;
+}): Promise<SentinelVote[]> => {
+	const { fromBlock, toBlock } = await getBlockRange(provider, maxBlockRange);
+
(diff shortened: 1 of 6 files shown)


explorer/src/lib/oracle/votes.ts
```diff
@@ -0,0 +1,64 @@
...
+	});
+
+	const revealedBySentinel = new Map(
+		logs.filter((log) => log.eventName === "Revealed").map((log) => [log.args.sentinel, log.args]),
+	);
+
+	return logs
+		.filter((log) => log.eventName === "Committed")
+		.sort((left, right) => {
+			if (left.blockNumber !== right.blockNumber) {
```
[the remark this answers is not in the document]
rmeissner: This should never happen and in this case it is fair to not display the vote

explorer/src/lib/oracle/votes.ts
```diff
@@ -0,0 +1,64 @@
...
+import { sentinelOracleV2Abi, sentinelVoteEventSelectors } from "./abi";
+
+export type SentinelVote =
+	| { sentinel: Address; state: "committed" }
+	| { sentinel: Address; state: "approved" | "denied"; reason: string };
+
+// Per-sentinel breakdown for a `SentinelOracleV2` request, discovered from indexed `Committed`/
+// `Revealed` logs rather than a configured roster — only sentinels who acted appear, ordered by
+// their `Committed` log position (first vote first). Callers should only invoke this once
+// `loadVotingStatus` has confirmed `kind === "sentinel"`; a generic `IOracle` has no such events.
```
nlordell, line 12: comment: This makes me think that the "legacy Oracle" ABI support in #663 should be removed, given it does not contain sufficient information for the explorer anyway.

explorer/src/lib/oracle/worker.ts
```diff
@@ -1,11 +1,14 @@
...
 import type { Address, Hex } from "viem";
 import { createRpcClient } from "@/lib/rpc";
+import { loadSentinelVotes } from "./votes";
 import { loadVotingStatus } from "./votingStatus";
 
 const workerApi = {
 	loadVotingStatus: ({ rpc, ...params }: { rpc: string; oracle: Address; requestId: Hex; maxBlockRange: bigint }) =>
 		loadVotingStatus({ ...params, provider: createRpcClient(rpc) }),
+	loadSentinelVotes: ({ rpc, ...params }: { rpc: string; oracle: Address; requestId: Hex; maxBlockRange: bigint }) =>
+		loadSentinelVotes({ ...params, provider: createRpcClient(rpc) }),
```
nlordell, line 11: request change: Idem, I do not think this should be a worker without some justification. Right now, this is all IO-bound logic (i.e. waiting for a fetch on the RPC node to perform the request, with only a few events to order). I do not see a valid reason for a worker here.

nlordell (changes requested):

explorer/src/lib/oracle/worker.ts
rmeissner: We already use workers for the other parsing of requests (list of all proposals and list of attestations), but we should reevaluate this:

https://github.com/safe-research/safenet/issues/670

nlordell (approved):

PR #673, opened 2026-08-04
[Phase 5a]: Vote UI components

rmeissner:
Adding the new UI React components that will be used in the SafeTx Proposal overview. The style of the new components is the same as for the attestation status component.

---

## CLA signature

### DIFF
--- a/explorer/src/components/common/ValidatorList.tsx
+++ b/explorer/src/components/common/ValidatorList.tsx
@@ -14,11 +14,13 @@ export function ValidatorList({
 	active,
 	mapInfo,
 	completed,
+	popoverContent,
 }: {
 	all: Address[];
 	active: Address[];
 	mapInfo: (suffix: string) => (address: Address) => string;
 	completed: boolean;
+	popoverContent?: (address: Address) => React.ReactNode;
 }) {
 	const activeSet = new Set(active);
 
@@ -42,6 +44,7 @@ export function ValidatorList({
 							<span className="font-mono text-xs">{item.address}</span>
 							<CopyButton value={item.address} />
 						</div>
+						{popoverContent?.(item.address)}
 					</InfoPopover>
 					{index < allItems.length - 1 && ", "}
 				</Fragment>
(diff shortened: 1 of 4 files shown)


explorer/src/components/common/ValidatorList.tsx
```diff
@@ -42,6 +44,7 @@ export function ValidatorList({
 							<span className="font-mono text-xs">{item.address}</span>
 							<CopyButton value={item.address} />
 						</div>
+						{popoverContent?.(item.address)}
```
nlordell, line 47: ~~question: I have never seen this notation before - is it semantically the same as `popoverContent ? popoverContent.item.address : undefined` (which can also be written as `popoverContent?.item?.address` - albeit with slightly different semantics).~~

Edit: Ah - it's a function. Brain fart 🤦.

explorer/src/components/transaction/SentinelVoteList.tsx
```diff
@@ -0,0 +1,44 @@
...
+		votes
+			.filter((vote): vote is SentinelVote & { reason: string } => vote.state !== "committed")
+			.map((vote) => [vote.sentinel, vote.reason]),
+	);
+
+	return (
+		<div className="md:flex md:justify-between">
+			<p className="ml-4">Votes:</p>
+			<div>
+				<ValidatorList
```
nlordell, line 31: nit: This is no longer a validator list right? This is more of a "account/participant/actor/member/etc. list" (naming is hard), and `ValidatorList` has one materialisation of this, and `SentinelVoteList` is another.

nlordell (approved):
LGTM. One nit that we are reusing `ValidatorList` when (IMO), we should be composing on a `AccountList`: `ValidatorList(AccountList)` and `SentinelVoteList(AccountList)`.

explorer/src/components/transaction/SentinelVoteList.tsx
rmeissner: I will push this as part of Phase 6 (Cleanup)

PR #680, opened 2026-08-05
Update workflow to build docker images

rmeissner:
- Build the rust based validator
- Add build for rust based sentinel

The same image names are kept. This is important to note as the new versions come with breaking changes. The `beta` tag is removed, as the new versions are not compatible with Safenet Beta.

Building the arm64 images for the Rust based sentinel and validator takes significant time. Therefore these will only be build for tags and the main branch. Also to avoid stacking of long running actions, we create a concurrency group to cancel the running workflow when new changes are pushed. 

---

## CLA signature

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,12 +8,24 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   docker:
     runs-on: ubuntu-latest
     permissions:
       contents: read
       packages: write
+
+    strategy:
+      matrix:
+        include:
+          - image: validator
+            dockerfile: crates/validator/Dockerfile
+          - image: sentinel
+            dockerfile: crates/sentinel/Dockerfile
 
     steps:
       - uses: actions/checkout@v5
@@ -29,12 +41,7 @@ jobs:
       - id: meta
         uses: docker/metadata-action@v5
         with:
-          images: ghcr.io/safe-research/safenet-validator
-          tags: |
-            type=ref,event=branch
-            type=ref,event=tag
-            # create beta tags on new releases
-            type=raw,value=beta,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+          images: ghcr.io/safe-research/safenet-${{ matrix.image }}
           flavor: |
             latest=false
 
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -44,7 +52,7 @@ jobs:
       - uses: docker/build-push-action@v6
         with:
           context: .
-          file: validator/Dockerfile
+          file: ${{ matrix.dockerfile }}
           platforms: linux/amd64,linux/arm64
```
nlordell: Oouf, I can imagine that `arm` build will take a while to complete.

.github/workflows/docker.yml
```diff
@@ -29,12 +37,12 @@ jobs:
...
         with:
-          images: ghcr.io/safe-research/safenet-validator
+          images: ghcr.io/safe-research/safenet-${{ matrix.image }}
           tags: |
             type=ref,event=branch
             type=ref,event=tag
-            # create beta tags on new releases
-            type=raw,value=beta,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
+            # create q3 tags on new releases
+            type=raw,value=q3,enable=${{ startsWith(github.ref, 'refs/tags/v') }}
```
nlordell: I wonder if channel tagging (i.e. `beta`, `q3`, etc.) should be something that we do manually. That is a separate discussion though, and it is fine as-is for now.

.github/workflows/docker.yml
```diff
@@ -29,12 +37,12 @@ jobs:
...
           flavor: |
             latest=false
```
nlordell: Might make sense to remove the custom tagging for now, until we decide what we want to do:

```suggestion
```

nlordell (approved):

.github/workflows/docker.yml
rmeissner: I will remove it as discussed in Slack

.github/workflows/docker.yml
rmeissner: I would keep branch and tag, or would you remove that too?

nlordell: I would just use the defaults (which includes branch and tag).

.github/workflows/docker.yml
```diff
@@ -8,13 +8,25 @@ on:
     branches:
       - main
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
nlordell, line 14: request change: it isn't clear to me what this does (at least the details of it) or why it is needed or a desired change - can the rationale be included in the PR description?

rmeissner: done, let me know if the description works for you

nlordell (approved):

PR #682, opened 2026-08-05
Improve Driver Error Handling

nlordell:
Improve driver error handling so failures cannot cause an already-fetched chain update to be skipped. Watcher failures are retried, transaction queue intermittent failures are ignored, and all other errors are considered unrecoverable and stop the driver.

Supersedes #654 

### Context and Motivation

The driver previously retried most step errors by starting a new iteration. If an error occurred after the watcher produced an update but before the state machine processed it, the next iteration requested a subsequent update and could desynchronize the state machine. This could happen with an intermittent RPC error while resubmitting stale transactions for example. Error handling now depends on where the failure occurs.

### Decisions and Tradeoffs

- Watcher failures are handled before an update is consumed and retried after a short delay. This matches the existing behaviour for watcher errors.
- Transaction queue RPC errors are treated as intermittent. They are logged while state processing continues, allowing transaction housekeeping and submission to recover naturally on later blocks.
- All other errors (some of which were previously retried) are now considered fatal.
- State transitions and action enqueueing remain separate operations. The protocol's consensus and timeout mechanisms provide recovery if an action is not submitted because of an intermittent failure.

### Testing

For now, we've omitted tests (since we didn't have existing driver tests, and the happy path is already verified as part of the integration tests).

---

## CLA signature

### DIFF
--- a/crates/core/src/driver.rs
+++ b/crates/core/src/driver.rs
@@ -181,10 +181,13 @@ where
     /// Only the wait for the next update races `shutdown`; once an update
     /// arrives it is processed to completion so its state transition and queued
     /// transactions are committed before the loop can stop.
-    async fn step(&mut self, shutdown: Pin<&mut impl Future<Output = ()>>) -> Result<Loop, Error> {
+    async fn step(
+        &mut self,
+        mut shutdown: Pin<&mut impl Future<Output = ()>>,
+    ) -> Result<Loop, Error> {
         let update = tokio::select! {
             biased;
-            _ = shutdown => return Ok(Loop::Break),
+            _ = shutdown.as_mut() => return Ok(Loop::Break),
             update = self.watcher.next() => update,
         };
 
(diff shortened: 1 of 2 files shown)


crates/core/src/driver.rs
```diff
@@ -160,18 +161,14 @@ where
 
         loop {
             match self.step(shutdown.as_mut()).await {
-                Ok(Loop::Continue) => {}
+                Ok(Loop::Continue) => continue,
```
rmeissner, line 164: question: is this functionally the same as before, just more explicit?

rmeissner (approved):

crates/core/src/driver.rs
nlordell: Exactly.

crates/core/src/driver.rs
```diff
@@ -188,16 +185,41 @@ where
...
+            Ok(update) => {
+                tracing::trace!(?update, "received watcher update");
+                update
+            }
+            Err(err) => {
+                tracing::warn!(
+                    ?err,
+                    "failed to get next blockchain update; retrying after delay"
+                );
+                tokio::time::sleep(STEP_RETRY_DELAY).await;
```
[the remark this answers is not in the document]
nlordell: fixed

PR #685, opened 2026-08-06
Resume indexing from retained snapshots

### LINKED ISSUE #667: Restarting during a historical warp can fail deterministically
During a historical warp, state persistence prunes intermediate snapshots to the final block of
each processed page. The state-machine test explicitly confirms that only the latest snapshot
survives.

If the process restarts after a sufficiently advanced warp page:

1. The latest committed snapshot is block `S`.
2. Startup generates a synthetic uncle around `S - max_reorg_depth`.
3. Reorg restoration requires the uncle's parent snapshot.
4. That snapshot was deleted during warp pruning.
5. Snapshot storage returns `MissingSnapshot`, and the driver exits.

This applies both to a restart in the middle of a long warp and to the window between completing a
warp and building a full set of recent snapshots.

Required remediation:

- Retain at least `max_reorg_depth + 1` snapshots while warping, or
- Persist a warp checkpoint and resume mode that does not synthesize a reorg until a normal reorg
  window has been established.
- Add process-restart tests after every warp page and during recent-block catch-up.

nlordell:
Fixes #667 

Resume block indexing from the state snapshots retained in storage. This prevents startup from requesting a pruned rollback snapshot after restarting during or shortly after a historical warp.

### Context and Motivation

Driver startup previously derived its synthetic reorg boundary from `latest - max_reorg_depth`, assuming every snapshot in that range still existed. This is not always the case:

- The service is restarted mid-warp sync, where only the latest snapshot is kept (since all blocks in a warp are reorg safe)
- The service is restarted mid-reorg, where a snapshot was removed from storage, but the new canonical block was not yet committed to storage (for example, only the block update but not the logs were processed).

### Testing

Added some new tests for the modified behaviour.

---

## CLA signature

### DIFF
--- a/crates/core/src/index/blocks.rs
+++ b/crates/core/src/index/blocks.rs
@@ -138,15 +138,14 @@ where
 {
     /// Creates and initializes a block watcher.
     ///
-    /// When `last_indexed_block` is set, the watcher resumes from it, replaying
-    /// the last `max_reorg_depth` blocks via a deliberate "fake" reorg so any
-    /// reorg that happened while the service was down is re-indexed. Otherwise,
-    /// when `Config::start_block` is set, it back-fills from there via a warp
-    /// without a (fake) reorg.
+    /// When `indexed` is set, the watcher resumes from the persisted snapshot
+    /// range, replaying everything after its safe rollback anchor via a
+    /// synthetic reorg. Otherwise, when `Config::start_block` is set, it
+    /// back-fills from there via a warp without a fake reorg.
     pub async fn new(
         provider: P,
         config: Config,
-        last_indexed_block: Option<u64>,
+        indexed: Option<BlockStatus>,
     ) -> Result<Self, Error> {
         let block_time = resolve_block_time(&provider, config.block_time).await?;
         let mut watcher = Self {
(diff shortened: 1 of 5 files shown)


crates/core/src/index/blocks.rs
```diff
@@ -259,14 +263,24 @@ where
...
-        // for reorg detection.
+        // Queue new-block updates for the recent blocks. Earlier blocks are
+        // still retained for reorg detection, but updates before the persisted
+        // rollback boundary (or configured start block on a fresh start) have
+        // already been processed and must not be emitted.
         for block in self.recent.iter().filter(|block| {
-            self.config
-                .start_block
-                .is_none_or(|start_block| block.header.number >= start_block)
+            indexed.map_or_else(
```
rmeissner, line 271: this function is unintuitive, it is called "map or else", but the first block is the "else" and then comes the "map" ...

rmeissner (approved):

crates/core/src/index/blocks.rs
nlordell: I know, I was also confused. I blame the Rust standard library :P.

PR #689, opened 2026-08-07
[Effects 1a]: Introduce Core Effect Module

nlordell:
### Context and Motivation

This PR just introduces an `effects` module to the core and moves the existing effects-related types to that module.

It was done in order to make PR reviews easier and less noisy.

### Decisions and Tradeoffs

We additionally change the method from `&self` to `&mut self`: this is because if we want concurrent processing of effects, the handler itself cannot be mutable in order for it to be shared across `perform_effect` tasks. If parts of the handler _must_ be mutable, then it should use internal mutability (with a `Mutex` or `RWLock`) and must be up to the effect handler to implement correctly (otherwise, all effects would be handled in sequence and not support concurrency).

---

## CLA signature

### DIFF
--- a/crates/validator/src/service/effect.rs
+++ b/crates/validator/src/service/effect.rs
@@ -10,7 +10,7 @@ use crate::{
     secrets::SecretStore,
 };
 use alloy::primitives::{Address, B256};
-use safenet_core::state::EffectHandler;
+use safenet_core::effects::EffectHandler;
 use std::{
     error::Error,
     fmt::{self, Display, Formatter},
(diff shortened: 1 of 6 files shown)


crates/validator/src/service/effect.rs
```diff
@@ -108,15 +108,17 @@ pub struct Handler {
...
 impl Handler {
-    async fn try_perform_effect(&mut self, effect: Effect) -> Result<Resume, InternalError> {
+    async fn try_perform_effect(&self, effect: Effect) -> Result<Resume, InternalError> {
         match effect {
             Effect::KeyGenSetup {
                 group_id,
                 count,
                 threshold,
             } => {
-                let mut rng = rand::thread_rng();
```
rmeissner, line 118: question: why was this change done?

rmeissner (approved):

crates/validator/src/service/effect.rs
```diff
@@ -237,7 +241,7 @@ impl Handler {
 }
 
 impl EffectHandler<Effect, Resume> for Handler {
-    async fn perform_effect(&mut self, effect: Effect) -> Resume {
+    async fn perform_effect(&self, effect: Effect) -> Resume {
```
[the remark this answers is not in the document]
nlordell: This is not true - we need the `effect` reference for the tracing log.

crates/validator/src/service/effect.rs
nlordell: It was a small style thing that was done in passing by the agent: it is to indicate that the `rng` that we use here is _only_ used for the FROST operation (i.e. generating the `secrets`) and not after anymore.

IMO, it is a nice cleanup to have (but you are right that it should not be necessarily part of this PR as it is unrelated). Let me know if you want me to split this out.

rmeissner: no, totally fair to have this here, just wanted to know if this is related to the other mut change (#learningrust)

PR #695, opened 2026-08-07
[Engine Epic]: Prepare Sentinel Engine Epic

nlordell:
:point_up: 

---

## CLA signature

### DIFF
--- a/epics/2026_08_07_sentinel_engine.md
+++ b/epics/2026_08_07_sentinel_engine.md
@@ -1,0 +1,397 @@
+# Plan: Sentinel Engine
+
+Component: `crates/sentinel` (Rust), new `crates/sentinel-engine` (Rust), `crates/safe-tx` (Rust), plus `scripts/` and `docs/`.
+
+---
+
+## Overview
+
+`crates/sentinel` currently does two unrelated jobs in one process. It is a **protocol participant**: it watches `Consensus`/`SentinelOracle`, drives the commit-reveal FSM (`state.rs`, `service.rs`), signs and submits `commit`/`reveal`/`finalize`/`claim` transactions, and manages its bond. It is also a **transaction verifier**: `static_checker.rs` runs the Charter's deterministic base guarantees plus a destination blocklist, `cow.rs` (1297 lines) decodes CoW Protocol batches and calls CoW's public order API, `address_poisoning.rs` issues `eth_getLogs` lookups, and `dynamic_checker.rs` POSTs to an optional operator-configured endpoint for whatever isn't implemented locally.
+
+These two jobs have nothing in common. The protocol side is fixed by the contracts and must be identical across every sentinel operator; the verification side is exactly where operators are meant to differentiate, iterate quickly, and take on liability. Today they share a process, a crate, a config file, a release cadence, and a deployment. `dynamic_checker.rs`'s own module docs already name the seam ("`RemoteChecker`'s `Checker` impl is already the seam to split _trigger this endpoint, parse the response_ along if that ever needs to move out on its own").
+
+This epic promotes that seam into a real service boundary:
+
+- **`crates/sentinel-engine`** — a new, stateless HTTP service. It accepts a proposed Safe transaction and answers whether it is valid. It holds no request state, has no signer, no database, and submits nothing onchain. Every checker in the repo today (`static_checker.rs`, `cow.rs`, `address_poisoning.rs`) moves here, unchanged. It also holds the hand-authored **`openapi.yaml`** that is the interface's source of truth.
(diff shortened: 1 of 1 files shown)


epics/2026_08_07_sentinel_engine.md
```diff
@@ -0,0 +1,397 @@
...
+
+---
+
+## Architecture Decision
+
+**The engine is stateless, synchronous request/response, and untrusted.** One `POST` in, one verdict out. No queue, no callback, no database, no signer, no onchain access other than read-only RPC for its own checks. This keeps the trust story simple: a compromised or buggy engine can only ever make its sentinel vote wrongly (or not at all) — it can never spend the sentinel's bond, sign anything, or corrupt the sentinel's FSM. It also means an operator can restart, rewrite, or replace their engine without touching the process that holds their key and their in-flight requests.
```
rmeissner: nit: the "no database" is not really a requirement, right? It doesn't have a state machine, but it could have a database of malicious accounts or similar.

epics/2026_08_07_sentinel_engine.md
```diff
@@ -0,0 +1,397 @@
...
+      required: [verdict, rule]
+      properties:
+        verdict: { const: invalid }
+        rule: { $ref: "#/components/schemas/RuleId" }
+    RuleId:
+      type: string
+      description: >-
+        A Safenet Arbitration Charter rule citation. Grown incrementally with
+        the checks that give each rule meaning; see `safe_tx::rule::RuleId`.
+      enum: ["R-4.1", "R-4.2", "R-4.3", "R-4.4", "R-4.5", "R-4.6"]
```
rmeissner: Nit: might make sense to note that this is not complete, I.e. how would we handle common law related verdicts

epics/2026_08_07_sentinel_engine.md
```diff
@@ -0,0 +1,397 @@
...
+      enum: [CALL, DELEGATECALL]
+    Verdict:
+      oneOf:
+        - $ref: "#/components/schemas/Valid"
+        - $ref: "#/components/schemas/Invalid"
+        - $ref: "#/components/schemas/Inconclusive"
+      discriminator:
+        propertyName: verdict
+        mapping:
+          valid: "#/components/schemas/Valid"
```
rmeissner: Request change: we should call them secure and insecure as this is the internally aligned naming (also in the charter)

epics/2026_08_07_sentinel_engine.md
```diff
@@ -0,0 +1,397 @@
...
+        rule: { $ref: "#/components/schemas/RuleId" }
+    RuleId:
+      type: string
+      description: >-
+        A Safenet Arbitration Charter rule citation. Grown incrementally with
+        the checks that give each rule meaning; see `safe_tx::rule::RuleId`.
+      enum: ["R-4.1", "R-4.2", "R-4.3", "R-4.4", "R-4.5", "R-4.6"]
+    Address:
+      type: string
+      description: Lower-case, `0x`-prefixed 20-byte address.
```
rmeissner: Nit that I would prefer to have changed if not too much overhead: Not checksummed? I know that this is not a human readable endpoint, therefore it is not the most important, but would prefer to always checksum addresses.

rmeissner (approved):

epics/2026_08_07_sentinel_engine.md
nlordell: It is not a general requirement, but how the reference sentinel engine will be implemented. I didn't correct it too much (since I don't want the agent to add dependencies for `sqlx` "just in case" we need a database later). You are right that it isn't a general sentinel agent requirement.

epics/2026_08_07_sentinel_engine.md
nlordell: I will review the OpenAPI spec in detail when I make the PR. There are a few other small issues I found.

epics/2026_08_07_sentinel_engine.md
nlordell: Again, I don't want to review the OpenAPI spec in this PR - the epic is more to direct the agent. I have noted the change in the epic.md, but generally I will leave the OpenAPI spec a bit "wrong" here (as it will get added in a separate PR).

rmeissner: Generally fine with this, out of experience Agents tend to take the epic spec as the source of truth sometimes :sweat_smile: so something to pay attention to later.

epics/2026_08_07_sentinel_engine.md
nlordell: addressed.

epics/2026_08_07_sentinel_engine.md
nlordell: I addressed your OpenAPI comments so it has something a bit better to work with anyhow :)

PR #698, opened 2026-08-10
feat(guard): verify oracle transaction-proposal attestations

remedcu:
Migrates `SafenetGuard`'s attestation path from the basic `TransactionProposal` message to the oracle `OracleTransactionProposal` message.

### Context and Motivation

- Safenet now gates transaction proposals through an oracle and threshold-signs `OracleTransactionProposal(epoch, oracle, safeTxHash)`. The guard reconstructed the basic `TransactionProposal(epoch, safeTxHash)`, so it would reject every oracle-based attestation. This makes the guard verify the message the network actually produces.

### Decisions and Tradeoffs

- The `AttestationTrailer` payload is re-laid out to carry the gating `oracle` (192 → 224 bytes). `TYPE_HASH` (`SafenetGuard.AttestationTrailer.v1`) is kept - a re-layout, not a version bump, since the format is unreleased.
- The oracle is trailer-supplied; the guard does not pin or allow-list it, it trusts the group key and rebuilds the exact signed message, leaving the "which oracle" policy to the validators (off-chain). The basic path is removed (full migration, no back-compat).

### Testing

- Updated the trailer codec and guard tests to the 224-byte oracle payload and the `OracleTransactionProposal` message; added guard-level fail-closed coverage for the `payloadLength != 224` shape.
- `forge test` - 246 passed, 0 failed.

### DIFF
--- a/contracts/src/guard/SafenetGuard.sol
+++ b/contracts/src/guard/SafenetGuard.sol
@@ -28,8 +28,8 @@ import {ISafe} from "@safe/interfaces/ISafe.sol";
  *      - **Epochs:** a forest of trusted `(group key, epoch)` pairs kept forever (Consensus lives only on
  *        Gnosis Chain; the guard holds a local copy). Any recorded pair may attest; a compromised
  *        historical key can attest future transactions (but not replay past ones — the Safe nonce binds).
- *      - **Attestation:** an inline trailer on `signatures` carrying `(epoch, groupKey, signature)`,
- *        verified against the full nonce-bound Safe tx hash.
+ *      - **Attestation:** an inline trailer on `signatures` carrying `(epoch, oracle, groupKey, signature)`,
+ *        verified against the oracle-transaction-proposal message over the full nonce-bound Safe tx hash.
  *      - **Escape hatch:** `announceTransaction` queues a transaction by its parameters (nonce excluded);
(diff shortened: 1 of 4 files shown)


rmeissner (approved):

contracts/src/guard/SafenetGuard.sol
```diff
@@ -220,7 +220,11 @@ contract SafenetGuard is ISafenetGuard, BaseTransactionGuard {
                     nonce: nonce
                 })
             );
-            bytes32 message = ConsensusMessages.transactionProposal(_CONSENSUS_DOMAIN_SEPARATOR, epoch, safeTxHash);
+            // The attestation binds the oracle it was gated by; the guard trusts the group key and rebuilds
```
nlordell, line 223: nit: The guard doesn't trust the group key since it is verified as known above (at least, this sentence sounds to me "the guard trusts the group key _value_").

nlordell (approved):

contracts/src/guard/SafenetGuard.sol
nlordell: @remedcu - in case you missed this

remedcu: https://github.com/safe-research/safenet/pull/705

PR #703, opened 2026-08-10
[Phase 2]: Run validator integration tests in CI

rmeissner:
Uses the existing validator port integration tests as a replacement for the original npm based integration tests. The integration tests uses an AlwaysApprove oracle. Only the happy path is tested. This integration test is run alongside the sentinel integration test in CI.

Further improvements (i.e. more extensive test cases like timeout flows) should be covered in a separate project.

For stability the integration tests have been slowed down and siplified:
- Block time is increased to 2 seconds
- Only propose one transaction for checking, after epoch rotation

---

## CLA signature


.github/workflows/integration.yml
```diff
@@ -40,3 +40,37 @@ jobs:
...
+
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          submodules: recursive
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: "lts/*"
+          cache: "npm"
```
nlordell, line 55: question: Hmm, we only need this in order to run the NPM scripts right? (like _start_ the `run_validator_integration_test.sh` script right?).

nlordell (approved):
LGTM - installing Node JS and NPM with all the dependencies needed for the explorer feels like a bit of a code-smell to me, but otherwise change looks good.

.github/workflows/integration.yml
```diff
@@ -40,3 +40,37 @@ jobs:
...
+
+      - uses: foundry-rs/foundry-toolchain@v1
+        with:
+          # We fix the foundry version because of a formatting bug introduced
+          # in version 1.7.0.
+          # See <https://github.com/foundry-rs/foundry/issues/13362>
+          version: v1.5.1
+
+      - run: npm ci
```
nlordell, line 64: comment: This _might_ not be needed - we don't actually execute any NodeJS code.

rmeissner: That is true, I will invoke the scripts directly and remove the npm related stuff (some work for the last phase already that removed the whole npm harness ;)

rmeissner: Actually I cannot clean this up here, as our integration test scripts rely on npm usage. This is something I will refactor in Phase 7 (where we deprecate the npm harness)

PR #705, opened 2026-08-10
Fill test gaps found during FV work and a small nit

remedcu:
## What it solves

While developing the SafenetGuard formal-verification specs, a few behaviours turned out to have no direct unit coverage. This adds those tests and, separately, clarifies the attestation trust-model wording that read as if the guard trusts the group key value (it verifies the key is a known forest member first).

## How this PR fixes it

Tests:

- **SafenetGuard**: the constructor derives the consensus domain separator; an announcement cannot authorize a delegatecall; cancelling an announcement works after its validity window expires.
- **FROST**: a non-canonical scalar (`z == N`) is rejected by the range check before any curve arithmetic.
- **TransactionAnnouncement**: a golden vector pins the full `hash` field layout (all nine fields, their order, and the pre-hashing of `data`).

Wording:

- Reword the `checkTransaction` comment and the DESIGN note so they state that the guard verifies the `(groupKey, epoch)` is a known forest member before rebuilding the signed message, rather than "trusting the group key".

## How to test it

`cd contracts && forge test` (251 tests pass).


rmeissner (approved):

rmeissner:
Should the test code coverage of the guard be 100%? The current reported coverage is `SafenetGuard.sol                               |84.2%    57|78.6%  14|    -    0`

PR #709, opened 2026-08-10
explorer: aim proposal lookup at the transaction timestamp

Fbartoli:
Proposals older than the head-relative `maxBlockRange` window (default 10k blocks, ~14h on Gnosis) show as "No proposals found" on the transaction page.

Fix: when the Safe tx-service reports a `submissionDate`, estimate the consensus-chain block at that timestamp and read a second log window aimed there, in addition to the head window. The head window is still read, so a proposal submitted now for an old transaction keeps showing up.

The estimate is secant refinement on observed block time: the first step uses the probe-to-head average, later steps use the cadence between the last two probes, so it converges across historical cadence shifts. Cost is 2-4 `eth_getBlockByNumber` plus one extra `eth_getLogs`, and only when the target is older than the head window. With no timestamp, a non-converging estimate, or a recent target, behaviour is unchanged.

Verified against `rpc.gnosischain.com` with an attestation ~210k blocks back (epoch 32947, proposed at block 47444993): main finds nothing, this branch shows it ATTESTED.

### DIFF
--- a/explorer/src/lib/utils.ts
+++ b/explorer/src/lib/utils.ts
@@ -38,6 +38,83 @@ export const getBlockRange = async (
 	return { fromBlock, toBlock };
 };
 
+/** Nominal seed for the block-at-timestamp estimate; refinement corrects for the real cadence. */
+const BLOCK_TIME_SEED_SECONDS = 5;
+/** Stop refining once a probe lands within this many seconds of the target. */
+const BLOCK_ESTIMATE_TOLERANCE_SECONDS = 600;
+/** Max refinement round-trips after the first probe (one cheap `eth_getBlockByNumber` each). */
+const BLOCK_ESTIMATE_MAX_REFINEMENTS = 8;
+/** How far behind the estimated block a targeted window starts, absorbing estimate error. */
+const TARGETED_WINDOW_BACK_BLOCKS = 1000n;
+
+/**
+ * Estimate the block nearest a unix timestamp via secant refinement: the first step
+ * uses the block time observed between the probe and the head, later steps the local
+ * cadence between the last two probes, so the estimate converges even when the real
+ * cadence drifts from the nominal seed or shifted over the chain's history.
+ * `driftSeconds` is measured at the returned block and is `Infinity` when no probe
+ * succeeded.
+ */
+export const estimateBlockAt = async (
+	provider: PublicClient,
+	targetSeconds: number,
+	head: { number: bigint; timestamp: number },
+): Promise<{ block: bigint; driftSeconds: number }> => {
+	const headNumber = Number(head.number);
+	const clamp = (block: number) => Math.min(headNumber, Math.max(0, Math.round(block)));
+	let guess = clamp(headNumber - (head.timestamp - targetSeconds) / BLOCK_TIME_SEED_SECONDS);
+	let best = { block: guess, driftSeconds: Number.POSITIVE_INFINITY };
+	let prev: { block: number; timestamp: number } | null = null;
+
+	for (let probe = 0; probe <= BLOCK_ESTIMATE_MAX_REFINEMENTS; probe++) {
+		// A failed probe (e.g. a pruned header on public RPC) degrades to the caller's
+		// head-relative scan rather than failing the whole read.
+		const block = await provider.getBlock({ blockNumber: BigInt(guess) }).catch(() => null);
+		if (!block) break;
+		const timestamp = Number(block.timestamp);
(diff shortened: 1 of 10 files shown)


explorer/src/lib/utils.ts
```diff
@@ -38,6 +38,83 @@ export const getBlockRange = async (
...
+		// head-relative scan rather than failing the whole read.
+		const block = await provider.getBlock({ blockNumber: BigInt(guess) }).catch(() => null);
+		if (!block) break;
+		const timestamp = Number(block.timestamp);
+		const driftSeconds = timestamp - targetSeconds;
+		if (Math.abs(driftSeconds) < Math.abs(best.driftSeconds)) best = { block: guess, driftSeconds };
+		if (Math.abs(driftSeconds) <= BLOCK_ESTIMATE_TOLERANCE_SECONDS) break;
+		// Consecutive probes bracket the target's cadence regime, so their secant slope
+		// beats the whole-chain average; the first probe only has the head to lean on.
+		const span = prev === null ? headNumber - guess : guess - prev.block;
```
rmeissner, line 80: nit: I find it confusing that one time guess is the subtrahend and the other it is the minuend. While logic wise this is ok, I was thinking about if order matters (as far as I can tell it does not).

explorer/src/lib/utils.ts
```diff
@@ -38,6 +38,83 @@ export const getBlockRange = async (
...
+	targetSeconds: number,
+	head: { number: bigint; timestamp: number },
+): Promise<{ block: bigint; driftSeconds: number }> => {
+	const headNumber = Number(head.number);
+	const clamp = (block: number) => Math.min(headNumber, Math.max(0, Math.round(block)));
+	let guess = clamp(headNumber - (head.timestamp - targetSeconds) / BLOCK_TIME_SEED_SECONDS);
+	let best = { block: guess, driftSeconds: Number.POSITIVE_INFINITY };
+	let prev: { block: number; timestamp: number } | null = null;
+
+	for (let probe = 0; probe <= BLOCK_ESTIMATE_MAX_REFINEMENTS; probe++) {
```
rmeissner, line 69: question: how many loops do we need in average?

rmeissner (commented):
If I see the code correctly then it would not find recent proposals for old transactions anymore, right? 

So if I take a transaction that was create a couple days ago and open it in the explorer, the explorer will look only on that day (within the block range), if it doesn't find anything and I propose it, then it will still stick to that old block range therefore not display the new proposal, correct?

Fbartoli:
> If I see the code correctly then it would not find recent proposals for old transactions anymore, right?
> 
> So if I take a transaction that was create a couple days ago and open it in the explorer, the explorer will look only on that day (within the block range) if it doesn't find anything and I propose it, then it will still stick to that old block range therefore not display the new proposal, correct?

No, the targeted window range is added to the head range.

There is still a gap in the approach, that would be fixed with a proper indexing solution. While not perfect it is still better than not displaying transaction older than 14 hours.

```ts                                           
const ranges = targeted !== null ? [targeted,
headRange] : [headRange];
```      

now where we would have problem is if we have something like that 

 ```                                                                                                           
range:      −4          −3          −2          −1        current
        ●───────────┬───────────┬───────────┬───────────┬──────────▶ head
        ▲ Safe tx               ▲ Safenet 
          (submissionDate)        proposal & attestation
        [targeted window]                               [head window]
        est−1k ────── est+10k                           head−10k ── head
        └── covers −4 → −3 ──┘                          └─ covers 0 ─┘
 ```
 
 in wallet, tx proposal and safenet proposal will be close to each other.
 Then attestations should happen soon after Safenet proposal so in theory should be fine

explorer/src/lib/utils.ts
Fbartoli: 2-3, worst case scenario is 9 if the chain doesn't respect 5 sec per block.

It depends how much the average block time deviates.

explorer/src/lib/utils.ts
Fbartoli: The only order than matter is that span and rise stay consistent with each other.

since we do rise / span later on, both needs to be the same sign, and the eventual negative signs cancel each others out in the division
```
const span = prev === null ? headNumber - guess : guess - prev.block;
const rise = prev === null ? head.timestamp - timestamp : timestamp - prev.timestamp;
const observedBlockTime = span !== 0 ? Math.max(rise / span, 0.1) : BLOCK_TIME_SEED_SECONDS;
```

PR #696, opened 2026-08-09
[Epic] Remove legacy code

rimeissner-ai:
Creates the epic for https://app.notion.com/p/safefoundation/Remove-legacy-code-3b4a8a34f3b880d69d1bcb046de2ed03

Concretely this will remove the typescript based validator and sentinel, adjust integration tests and remove non-oracle flows in the validator and explorer. This way only active code has to be maintained.

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://docs.safefoundation.org/cla).

PR #712, opened 2026-08-10
[Sentinel Engine 1a]: Prefer Local Binding Types

nlordell:
This PR refactors the `sentinel` and `validator` crates to use their internal `binding::SafeTransaction` types until using the common ones are necessary.

### Context and Motivation

With the Sentinel Engine epic, we will be defining strict `SafeTransaction` wire types for the sentinel engine API. In particular, these types **are not compatible** with the `sol!` generated types (since they add an `__Invalid` variant to the enum). On the way to implementing the sentinel engine, we want to make the `SafeTransaction` types more sound.

Unfortunately, just refactoring the Safe transaction type _everywhere_ would end up in a 1000+ line PR which is too much to review at once. Instead, we first modify both the `sentinel` and `validator` crates to use their internal `binding::SafeTransaction` representations everywhere, and only convert to the `safe_tx` variants where it is necessary. This also paves the way for removing the `safe_tx` dependency completely from the `validator` once the #696 lands in full, and move its contents completely to the `sentinel-engine` crate in this epic (also enabling the removal of the dependency in the `sentinel` crate by the end of this epic).

Additionally, we have a bit of type-level coupling between the `sentinel` and `safe_tx` (some check and types are implemented in `safe_tx`, others in `sentinel` with the separation appearing seemingly arbitrary). I will move everything over time to the `sentinel-engine` crate.

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/cow.rs
+++ b/crates/sentinel/src/cow.rs
@@ -56,17 +56,16 @@
 //! under-approval and an over-approval are denied here. An unreachable or
 //! malformed API response is `Unknown`, not guessed at either way.
 
-use crate::checker::{CheckOutcome, Checker};
+use crate::{
+    bindings::consensus::{Operation, SafeTransaction},
+    checker::{CheckOutcome, Checker},
+};
 use alloy::{
     primitives::{Address, B256, Bytes, U256, address},
     sol,
     sol_types::{Eip712Domain, SolCall, SolStruct, SolValue, eip712_domain},
 };
-use safe_tx::{
-    multi_send::decode_multi_send_call,
-    rule::RuleId,
-    types::{Operation, SafeTransaction, erc20::approveCall},
-};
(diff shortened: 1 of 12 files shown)


crates/sentinel/src/cow.rs
```diff
@@ -466,9 +465,10 @@ impl Checker for CowChecker {
 
 /// `tx` itself, or, if it's a MultiSend batch, each of its sub-calls.
 fn sub_transactions(tx: &SafeTransaction) -> Vec<SafeTransaction> {
-    decode_multi_send_call(tx)
-        .map(|(sub_txs, _)| sub_txs)
-        .unwrap_or_else(|| vec![tx.clone()])
+    let tx = tx.clone().into();
+    decode_multi_send_call(&tx)
+        .map(|(sub_txs, _)| sub_txs.into_iter().map(|sub_tx| sub_tx.into()).collect())
```
rmeissner, line 470: yikes, but makes sense for the current state.

rmeissner (approved):

PR #713, opened 2026-08-10
[Sentinel Engine 1b]: Static Checker Returns Outcome

nlordell:
`sol!` generated types are generated with an additional `__Invalid` variant on their enums. In order to correctly handle this case with the static checker (in a follow-up PR), we need to be able to respond with `Unknown` in case of an unexpected value. It should be prevented at the contract level, so this is mostly to be extra defensive (and avoid `panic!`s in face of a misbehaving node).

---

## CLA signature

### DIFF
--- a/crates/sentinel/src/service.rs
+++ b/crates/sentinel/src/service.rs
@@ -131,32 +131,36 @@ impl SentinelTransition {
             return (state, Vec::new());
         }
         let deadline = block.saturating_add(self.voting_window);
-        let decision = self.static_checker.check(&event.transaction);
-        if !decision.approve {
-            state.0.insert(
-                request_id,
+        let outcome = self.static_checker.check(&event.transaction);
+        let (request, commands) = match outcome {
+            CheckOutcome::Approved => (
+                RequestState::WaitingForDynamicCheck {
+                    deadline,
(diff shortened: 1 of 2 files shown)


crates/sentinel/src/service.rs
```diff
@@ -131,32 +131,36 @@ impl SentinelTransition {
             return (state, Vec::new());
         }
         let deadline = block.saturating_add(self.voting_window);
-        let decision = self.static_checker.check(&event.transaction);
-        if !decision.approve {
-            state.0.insert(
-                request_id,
+        let outcome = self.static_checker.check(&event.transaction);
```
rmeissner, line 134: question: from the epic it sounds like that the logic of the static checks is kept (so checks that are run before invoking the external checks). Do we want to keep this, or should we fully remove the static_checker?

Disclaimer: I only searched the epic for "static"

rmeissner (approved):

crates/sentinel/src/service.rs
nlordell: I plan on fully removing the static checks (i.e. moving all checks to the engine). I don't think this is explicit in the epic.

PR #714, opened 2026-08-10
[Sentinel Engine 1c]: Exact Safe Transaction Type

nlordell:
This PR moves the `SafeTransaction` type that was previously generated in a `sol!` binding to be a manual definition. In fact, there was no reason to use `sol!` bindings in the first place, and it allows us to be more exact with the `Operation` enum type (which _previously_ included a hidden `__Invalid` variant).

### Decisions and Tradeoffs

- In order to reduce the amount of code-change noise, we keep the field names in `camelCase` and added a `#[allow(nonstandard_style)]` to appease the linter. This will be addressed in a followup.
- We needed to deal with the fallibility of converting a `bindings::Operation` to a `safe_tx::Operation` (i.e. handle the `__Invalid` case) in a few places, but it was mostly mechanical. The most notable change is that the static checker can now return an `Unknown` outcome in case this conversion fails (although, this should be prevented at a contract level and can only happen in face of a misbehaving node).

---

## CLA signature


crates/safe-tx/src/bindings.rs
```diff
@@ -0,0 +1,34 @@
+//! `sol!`-generated bindings for the contract interfaces the checks decode
+//! calldata against.
+
+pub mod erc20 {
```
rmeissner, line 4: nit: I am missing that these are removed from the types.rs, will this happen later?

rmeissner (approved):

crates/safe-tx/src/bindings.rs
nlordell: no... it should be gone. Must have forgotten to `git rm` the file. Will fix.

PR #719, opened 2026-08-11
docs: epic for binding oracleData into the oracle attestation

remedcu:
Plan for making `oracleData` a first-class, authenticated field of the SafenetGuard oracle attestation. This is an epic-only PR.

### Context and Motivation

- Today `oracleData` is passed to `Consensus.proposeOracleTransaction` and forwarded to the oracle via `postRequest`, but is deliberately excluded from the FROST-signed `OracleTransactionProposal` message, so the guard never sees it and nothing authenticates it.
- The team wants it to become a first-class, authenticated part of the attestation so a future guard or verifier can rely on it. The plan commits `keccak256(oracleData)` into the signed message and carries the raw bytes in the guard trailer, with the guard rebuilding and FROST-verifying the commitment. It is empty (`keccak256("")`) today, so there is no semantic change yet, only the authenticated capacity to carry one.

### Decisions and Tradeoffs

- **Bind a hash, not the raw bytes, into the message.** The message gains one fixed-width `bytes32 oracleDataHash`; the raw bytes travel only in the guard trailer (the one place that must recompute the commitment). Keeps the EIP-712 struct static and the new event field 32 bytes.
- **Protocol-wide, not guard-local.** The signed message hash is also the oracle `requestId` and the `$attestations` key, so `ConsensusMessages`/`Consensus`/`IConsensus` (with a new `oracleDataHash` event field), the Rust validator hasher, the Rust sentinel hasher, the explorer mirror, and the guard all change together.
- **Phase 1 couples Solidity core plus the Rust validator and sentinel into one PR.** The cross-language integration jobs exercise the propose to attest round-trip and would go red on any partial landing, so the coupled loop moves together; the guard, docs, explorer, and example are separate downstream phases.
- **Open items flagged for the team:** redefine trailer `v1` vs bump to `v2`; the `requestId` now depends on `oracleData` (same tx with different `oracleData` no longer collides on `AlreadyAttested`); whether to preserve `MalformedAttestationTrailer` given the now variable-length payload.


epics/2026_08_11_bind_oracle_data.md
```diff
@@ -0,0 +1,151 @@
...
+7. **The reference example** - `examples/attest-safe-tx.ts` builds the trailer and the signed message by hand (`:54,183,271`).
+
+None of this is deployed (the guard is unreleased; the protocol runs on devnet), so there is no runtime backward-compatibility window to preserve - only per-PR compile/test consistency and cross-language integration-test consistency to keep green.
+
+---
+
+## Architecture Decision
+
+**Bind a `keccak256(oracleData)` commitment, not the raw bytes, into the signed message.** The message gains one field: `OracleTransactionProposal(uint64 epoch,address oracle,bytes32 safeTxHash,bytes32 oracleDataHash)`. A fixed-width commitment keeps the EIP-712 struct static and cheap, keeps the event field 32 bytes, and is all any hash-reconstructing consumer (validator, sentinel, explorer) needs. The raw `oracleData` bytes are only needed by parties that must *recompute* the commitment (the proposer already has them; the oracle already receives them via `postRequest`; the guard receives them in the trailer). So the raw bytes travel in exactly one new place - the guard trailer - and everywhere else carries the 32-byte hash.
```
rmeissner: > not the raw bytes, into the signed message.

request change: The typed data should include the raw bytes in my opinion. Hashing will happen in any case as part of this. I would also emit the whole oracleData in a event.

Happy to discuss this

rmeissner: Also `A fixed-width commitment keeps the EIP-712 struct static and cheap, keeps the event field 32 bytes, and is all any hash-reconstructing consumer (validator, sentinel, explorer) needs.` is kind of wrong, as EIP-712 is always static, and costs actually don't change at all, in both cases the data needs to be hashed once. Only change is potentially what is emitted with the event.

rmeissner (changes requested):

epics/2026_08_11_bind_oracle_data.md
remedcu: Updated in the latest commit 👍🏾

rmeissner (approved):

Issue #669, opened 2026-08-04
Adjust Proposal Events


rmeissner:
The following fields should be indexed:
- SafeTxHash
- Safe Id (Combination of chainId and Safe address)
- Oracle

Reference:
- https://github.com/safe-research/safenet/blob/a331badd5421ecb9467da6ee5678fa5eea1f0d4f/contracts/src/interfaces/IConsensus.sol#L101-L117

PR #722, opened 2026-08-11
[Phase 6] Remove non oracle proposal related explorer code

rmeissner:
To reduce the amount of unused explorer code the code paths not related to oracle proposals are removed.

### Testing

- Feature preview version should work with the current testnet

---

## CLA signature


explorer/src/lib/consensus/transactions.ts
nlordell, line 121: question: I don't remember what we decided about oracle indexing... Do you remember?

nlordell (approved):

explorer/src/lib/consensus/transactions.ts
rmeissner: ahhh yeah, there is a github issue for this https://github.com/safe-research/safenet/issues/669

I would handle this after this PR, hopefully tomorrow. In the best case we could relaunch testnet later this week.

PR #723, opened 2026-08-11
[Phase 7a] Remove Oracle prefix from contract code naming

rmeissner:
As there is only one transaction type it is not necessary to use the oracle prefix to distuingish. This PR adjusts this across the whole code base. Only renaming should be done, no logic change.

### Testing

- Integration tests should pass

---

## CLA signature

### DIFF
--- a/crates/validator/src/state/mod.rs
+++ b/crates/validator/src/state/mod.rs
@@ -424,10 +424,10 @@ impl StateTransition<State> for Transition {
                 Event::Consensus(Consensus::ConsensusEvents::EpochStaged(event)) => {
                     self.handle_epoch_staged(state, &event)
                 }
-                Event::Consensus(Consensus::ConsensusEvents::OracleTransactionProposed(event)) => {
+                Event::Consensus(Consensus::ConsensusEvents::TransactionProposed(event)) => {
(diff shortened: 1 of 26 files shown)


crates/validator/src/state/mod.rs
```diff
@@ -425,10 +425,10 @@ impl StateTransition<State> for Transition {
                 Event::Consensus(Consensus::ConsensusEvents::EpochStaged(event)) => {
                     self.handle_epoch_staged(state, &event)
                 }
-                Event::Consensus(Consensus::ConsensusEvents::OracleTransactionProposed(event)) => {
+                Event::Consensus(Consensus::ConsensusEvents::TransactionProposed(event)) => {
                     self.handle_oracle_transaction_proposed(state, log.block, &event)
```
nlordell, line 428: nit: here too:
```suggestion
                    self.handle_transaction_proposed(state, log.block, &event)
```

nlordell: *Edit*: Ah, the PR title says it all: **"FROM CONTRACT CODE NAMING"**. Rust and TS method renaming will come in a followup I assume.

nlordell (approved):

PR #726, opened 2026-08-11
feat(consensus): bind oracleData into the oracle transaction proposal

remedcu:
Bind `oracleData` into the FROST-signed transaction-proposal message across consensus, the validator, and the sentinel.

### Context and Motivation

- The guard attestation needs to authenticate the `oracleData` a proposal was gated with. This adds `oracleData` as a `bytes` member of the `TransactionProposal` EIP-712 message, so the signed message (and the oracle `requestId`) commits to it.

### Decisions and Tradeoffs

- `ConsensusMessages.transactionProposal` takes the `oracleData` as its `keccak256` directly (EIP-712 encodes a `bytes` member as its hash), placed next to `oracle`. Callers holding the full data hash at the callsite; the guard supplies the hash. This keeps a single builder (no separate hash variant).
- The proposed event emits the full `oracleData` (and it is passed to the oracle); the attested event, `attestTransaction`, the getter, and the `onSignCompleted` context carry only the `bytes32` hash, so the attestation and validator callback stay a constant size.
- Validator: the proposed/signing path uses alloy's `SolStruct` hashing over the full bytes; the attest path (which only has the hash on-chain) rebuilds the same digest by hand, pinned against the `SolStruct` and the cross-language vectors in tests.

### Testing

- `forge test` (253), `forge fmt --check`, `forge lint`.
- `cargo fmt --check`, `cargo clippy --workspace --all-targets`, `cargo test --workspace`.
- `scripts/run_validator_integration_test.sh` and `scripts/run_sentinel_integration_test.sh`: both SUCCESS end-to-end.

### DIFF
--- a/contracts/src/Consensus.sol
+++ b/contracts/src/Consensus.sol
@@ -256,10 +256,12 @@ contract Consensus is IConsensus, IERC165, IFROSTCoordinatorCallback {
     {
         Epochs memory epochs = _processRollover();
         safeTxHash = transaction.hash();
-        bytes32 message = domainSeparator().transactionProposal(epochs.active, oracle, safeTxHash);
+        // forge-lint: disable-next-line(asm-keccak256)
+        bytes32 message =
+            domainSeparator().transactionProposal(epochs.active, oracle, keccak256(oracleData), safeTxHash);
         require($attestations[message].isZero(), AlreadyAttested());
         SafeId.T safeId = SafeId.create(transaction.chainId, transaction.safe);
-        emit TransactionProposed(safeTxHash, safeId, oracle, epochs.active, transaction);
+        emit TransactionProposed(safeTxHash, safeId, oracle, epochs.active, oracleData, transaction);
         _COORDINATOR.sign($groups[epochs.active], message);
         IOracle(oracle).postRequest(message, msg.sender, oracleData);
     }
@@ -270,29 +272,30 @@ contract Consensus is IConsensus, IERC165, IFROSTCoordinatorCallback {
     function attestTransaction(
         uint64 epoch,
         address oracle,
+        bytes32 oracleDataHash,
         uint256 chainId,
         address safe,
         bytes32 safeTxStructHash,
         FROSTSignatureId.T signatureId
     ) public {
         bytes32 safeTxHash = SafeTransaction.partialHash(chainId, safe, safeTxStructHash);
-        bytes32 message = domainSeparator().transactionProposal(epoch, oracle, safeTxHash);
+        bytes32 message = domainSeparator().transactionProposal(epoch, oracle, oracleDataHash, safeTxHash);
         require($attestations[message].isZero(), AlreadyAttested());
         FROST.Signature memory attestation = _COORDINATOR.signatureVerify(signatureId, $groups[epoch], message);
         $attestations[message] = signatureId;
         SafeId.T safeId = SafeId.create(chainId, safe);
-        emit TransactionAttested(safeTxHash, safeId, oracle, epoch, signatureId, attestation);
+        emit TransactionAttested(safeTxHash, safeId, oracle, epoch, oracleDataHash, signatureId, attestation);
     }
 
     /**
      * @inheritdoc IConsensus
      */
-    function getTransactionAttestationByHash(uint64 epoch, address oracle, bytes32 safeTxHash)
+    function getTransactionAttestationByHash(uint64 epoch, address oracle, bytes32 oracleDataHash, bytes32 safeTxHash)
         public
         view
         returns (FROST.Signature memory signature)
     {
-        bytes32 message = domainSeparator().transactionProposal(epoch, oracle, safeTxHash);
+        bytes32 message = domainSeparator().transactionProposal(epoch, oracle, oracleDataHash, safeTxHash);
         return _COORDINATOR.signatureValue($attestations[message]);
     }
 
@@ -331,9 +334,15 @@ contract Consensus is IConsensus, IERC165, IFROSTCoordinatorCallback {
                 abi.decode(context[4:], (uint64, uint64, FROSTGroupId.T));
             stageEpoch(proposedEpoch, rolloverBlock, groupId, signatureId);
         } else if (selector == this.attestTransaction.selector) {
-            (uint64 epoch, address oracle, uint256 chainId, address safe, bytes32 safeTxStructHash) =
-                abi.decode(context[4:], (uint64, address, uint256, address, bytes32));
-            attestTransaction(epoch, oracle, chainId, safe, safeTxStructHash, signatureId);
+            (
+                uint64 epoch,
+                address oracle,
+                bytes32 oracleDataHash,
+                uint256 chainId,
+                address safe,
+                bytes32 safeTxStructHash
+            ) = abi.decode(context[4:], (uint64, address, bytes32, uint256, address, bytes32));
+            attestTransaction(epoch, oracle, oracleDataHash, chainId, safe, safeTxStructHash, signatureId);
         } else {
             revert UnknownSignatureSelector();
         }
--- a/contracts/src/interfaces/IConsensus.sol
+++ b/contracts/src/interfaces/IConsensus.sol
@@ -71,6 +71,7 @@ interface IConsensus {
      * @param safeId The identifier of the Safe account, combining its chain ID and address.
      * @param oracle The address of the oracle contract used for evaluation.
      * @param epoch The epoch in which the transaction is proposed.
+     * @param oracleData Arbitrary oracle-specific data, bound into the signed message hash.
      * @param transaction The proposed Safe transaction.
      */
     event TransactionProposed(
@@ -78,6 +79,7 @@ interface IConsensus {
         SafeId.T indexed safeId,
         address indexed oracle,
         uint64 epoch,
+        bytes oracleData,
         SafeTransaction.T transaction
     );
 
@@ -87,6 +89,9 @@ interface IConsensus {
      * @param safeId The identifier of the Safe account, combining its chain ID and address.
      * @param oracle The address of the oracle contract used for evaluation.
      * @param epoch The epoch in which the attested transaction was proposed.
+     * @param oracleDataHash The `keccak256` of the oracle-specific data bound into the signed message hash. Only the
+     *        hash is echoed here (the full data is emitted by `TransactionProposed`), keeping the attestation a fixed
+     *        size so the validator callback has a constant calldata cost.
      * @param signatureId The FROST signature identifier corresponding to the attestation.
      * @param attestation The attestation to the oracle-checked Safe transaction.
      */
@@ -95,6 +100,7 @@ interface IConsensus {
         SafeId.T indexed safeId,
         address indexed oracle,
         uint64 epoch,
+        bytes32 oracleDataHash,
         FROSTSignatureId.T signatureId,
         FROST.Signature attestation
     );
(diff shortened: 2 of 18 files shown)


nlordell:
I will hold off on reviewing the stack because there will be a lot of conflicts to resolve (to the point that the code will change dramatically).

rmeissner (approved):

rmeissner (commented):
Integration tests fail and it would be very nice (close to a request change) that these would be fixed before merging, or that we explicitly mention the later phase.

contracts/src/interfaces/IConsensus.sol
```diff
@@ -71,13 +71,15 @@ interface IConsensus {
...
+     * @param oracleData Arbitrary oracle-specific data, bound into the signed message hash.
      * @param transaction The proposed Safe transaction.
      */
     event TransactionProposed(
         bytes32 indexed safeTxHash,
         SafeId.T indexed safeId,
         address indexed oracle,
         uint64 epoch,
+        bytes oracleData,
         SafeTransaction.T transaction
```
nlordell, line 83: ubernit: I would keep ordering consistent with the typed data, i.e.:


```suggestion
        SafeTransaction.T transaction,
        bytes oracleData
```

contracts/src/interfaces/IConsensus.sol
```diff
@@ -95,6 +98,7 @@ interface IConsensus {
         SafeId.T indexed safeId,
         address indexed oracle,
         uint64 epoch,
+        bytes oracleData,
```
nlordell: request change: We do not need to replay the full oracle data here (just like we don't replay the full transaction data), just enough to compute the packet:


```suggestion
        bytes32 oracleDataHash,
```

This matters in order to make the validator callback have a constant size (otherwise, we would need dynamic gas pricing in the validator logic).

contracts/src/interfaces/IConsensus.sol
```diff
@@ -173,7 +177,7 @@ interface IConsensus {
     /**
      * @notice Proposes a transaction for oracle-checked validator approval.
      * @param oracle Address of the oracle contract to use for evaluation.
-     * @param oracleData Arbitrary oracle-specific data passed to the oracle; not part of the signed message hash.
+     * @param oracleData Arbitrary oracle-specific data passed to the oracle; bound into the signed message hash (as the `bytes oracleData` EIP-712 member).
```
nlordell: nit: maximum column size is not respected here.

contracts/src/interfaces/IConsensus.sol
nlordell: Oops, got it wrong, it is in the correct order already.

contracts/src/interfaces/IConsensus.sol
```diff
@@ -194,6 +199,7 @@ interface IConsensus {
     function attestTransaction(
         uint64 epoch,
         address oracle,
+        bytes memory oracleData,
```
nlordell: nit:

```suggestion
        bytes calldata oracleData,
```

nlordell: I think for `external` functions, it doesn't make a difference, and keeps it consistent with the other attestation method.

contracts/src/libraries/ConsensusMessages.sol
```diff
@@ -85,20 +85,58 @@ library ConsensusMessages {
...
+            mstore(add(ptr, 0x20), epoch)
+            mstore(add(ptr, 0x40), oracle)
+            mstore(add(ptr, 0x60), safeTxHash)
+            mstore(add(ptr, 0x80), keccak256(add(oracleData, 0x20), mload(oracleData)))
+            mstore(add(ptr, 0x22), keccak256(ptr, 0xa0))
+            mstore(ptr, hex"1901")
+            mstore(add(ptr, 0x02), domainSeparator)
+            result := keccak256(ptr, 0x42)
+        }
+    }
```
nlordell: request change: This function is completely unnecessary IMO, you just replace all callsites with `keccak256(oracleData)` which is just as easy to type. This just adds a lot of additional code to the contract without a clear benefit.

```suggestion
```

contracts/src/Consensus.sol
```diff
@@ -256,10 +256,10 @@ contract Consensus is IConsensus, IERC165, IFROSTCoordinatorCallback {
     {
         Epochs memory epochs = _processRollover();
         safeTxHash = transaction.hash();
-        bytes32 message = domainSeparator().transactionProposal(epochs.active, oracle, safeTxHash);
+        bytes32 message = domainSeparator().transactionProposal(epochs.active, oracle, safeTxHash, oracleData);
```
nlordell: comment: regarding the comment above (about `transactionProposal` being unnecessary), this is trivially rewritten as:


```suggestion
        bytes32 message = domainSeparator().transactionProposal(epochs.active, oracle, safeTxHash, keccak256(oracleData));
```

Without duplicating a large chunk of hashing code in the final contract.

contracts/src/Consensus.sol
```diff
@@ -270,29 +270,31 @@ contract Consensus is IConsensus, IERC165, IFROSTCoordinatorCallback {
     function attestTransaction(
         uint64 epoch,
         address oracle,
+        bytes memory oracleData,
```
nlordell: request change: attestations do not need to echo the oracle data:


```suggestion
        bytes32 oracleDataHash,
```

contracts/src/Consensus.sol
```diff
@@ -270,29 +270,31 @@ contract Consensus is IConsensus, IERC165, IFROSTCoordinatorCallback {
...
-    function getTransactionAttestationByHash(uint64 epoch, address oracle, bytes32 safeTxHash)
-        public
-        view
-        returns (FROST.Signature memory signature)
-    {
-        bytes32 message = domainSeparator().transactionProposal(epoch, oracle, safeTxHash);
+    function getTransactionAttestationByHash(
+        uint64 epoch,
+        address oracle,
+        bytes calldata oracleData,
```
nlordell: nit: Same here, it is inconsistent with the `safeTxHash` handling too.

contracts/src/Consensus.sol
```diff
@@ -331,9 +333,15 @@ contract Consensus is IConsensus, IERC165, IFROSTCoordinatorCallback {
                 abi.decode(context[4:], (uint64, uint64, FROSTGroupId.T));
             stageEpoch(proposedEpoch, rolloverBlock, groupId, signatureId);
         } else if (selector == this.attestTransaction.selector) {
-            (uint64 epoch, address oracle, uint256 chainId, address safe, bytes32 safeTxStructHash) =
-                abi.decode(context[4:], (uint64, address, uint256, address, bytes32));
-            attestTransaction(epoch, oracle, chainId, safe, safeTxStructHash, signatureId);
+            (
+                uint64 epoch,
+                address oracle,
+                bytes memory oracleData,
```
nlordell: nit: Especially here, we would make the validators pay with way more calldata than needed if we just include the hash (keeping the size fixed).

crates/sentinel/src/bindings.rs
```diff
@@ -84,13 +84,14 @@ pub mod consensus {
 
         // EIP-712 struct for the oracle requestId.
         // Field order and types must exactly match the onchain typehash in ConsensusMessages.sol:
-        // keccak256("TransactionProposal(uint64 epoch,address oracle,bytes32 safeTxHash)")
+        // keccak256("TransactionProposal(uint64 epoch,address oracle,bytes32 safeTxHash,bytes oracleData)")
```
nlordell: request change: One thing I am not a fan of, is that ordering of the `oracleData` parameter is not consistent across the different functions and structs, for example:

```solidity
    function getTransactionAttestationByHash(
        uint64 epoch,
        address oracle,
        bytes calldata oracleData,
        bytes32 safeTxHash
    ) public view returns (FROST.Signature memory signature) {
```

I would just pick one (in particular, I would personally prefer it next to the `oracle` field):


```suggestion
        // keccak256("TransactionProposal(uint64 epoch,address oracle,bytes oracleData,bytes32 safeTxHash)")
```

---

Events are obviously different because indexed field ordering matters.

crates/validator/src/bindings.rs
```diff
@@ -5,6 +5,10 @@
 //! values are transcribed only where a return is actually decoded onchain
 //! (`getValidatorStaker`); the mutating calls omit them.
 
+// `sol!` generates constructors whose arity matches the onchain event/function
+// field count; the oracle attestation carries enough fields to trip the lint.
+#![allow(clippy::too_many_arguments)]
```
nlordell, line 13: question: is it possible you add this to the specific thing that has an arity issue (I figure it is `function attestTransaction`)?

nlordell (changes requested):
Lots of small nits, and some of them should be addressed IMO.

contracts/test/libraries/ConsensusMessages.t.sol
```diff
@@ -28,15 +28,17 @@ contract ConsensusMessagesTest is Test {
...
             ConsensusMessages.TRANSACTION_PROPOSAL_TYPEHASH,
-            keccak256("TransactionProposal(uint64 epoch,address oracle,bytes32 safeTxHash)")
+            keccak256("TransactionProposal(uint64 epoch,address oracle,bytes oracleData,bytes32 safeTxHash)")
         );
     }
 
     function test_TransactionProposal() public pure {
         bytes32 message = ConsensusMessages.domain(23, 0x4838B106FCe9647Bdf1E7877BF73cE8B0BAD5f97)
-            .transactionProposal(1, 0x1234567890123456789012345678901234567890, bytes32(uint256(42)));
+            .transactionProposal(
```
rmeissner, line 37: request: can we add a test that this corresponds to the eip712 message with "cafe"?

Maybe we could calculate the hash `973d5f5f4f873250bac60f83283dcff0f1bfbfa8ef19c2120830189a4ffa7084` in one of the other tests using standard 712 libs

rmeissner: Nvm, the test below does exactly that :)

crates/validator/src/consensus/hashing.rs
```diff
@@ -103,30 +112,66 @@ impl ConsensusDomain {
...
     /// The consensus-domain hash of an oracle-backed Safe transaction proposal,
     /// embedding the already-computed [`safe_tx_hash`] as its `safeTxHash` field.
+    ///
+    /// This is the proposal/signing path, which always holds the full `oracleData`, so it uses the
+    /// canonical `SolStruct` EIP-712 hashing (alloy hashes the `bytes` member itself).
     pub fn transaction_proposal_hash(
         &self,
         epoch: EpochId,
         oracle: Address,
+        oracle_data: Bytes,
```
nlordell, line 122: nit: can we also implement this here with

```suggestion
        oracle_data_hash: B256,
```

So we only have one hashing function implementation? You can trivially `keccak256(oracle_data)` at the callsite as well.

crates/validator/src/consensus/hashing.rs
```diff
@@ -103,30 +112,66 @@ impl ConsensusDomain {
...
+    /// value holding the full bytes, which this path does not have.
+    pub fn transaction_proposal_hash_from_data_hash(
+        &self,
+        epoch: EpochId,
+        oracle: Address,
+        oracle_data_hash: B256,
+        safe_tx_hash: B256,
+    ) -> B256 {
+        let mut data = [0u8; 160];
+        data[0..32].copy_from_slice(TRANSACTION_PROPOSAL_TYPE_HASH.as_slice());
```
nlordell, line 150: nit: I think `sol!` generates the typehash value already:

```rust
TransactionProposal::eip712_type_hash
```

crates/validator/src/bindings.rs
nlordell: nit: I feel this comment is overly verbose.

nlordell (approved):
LGTM, just some smaller nits.

crates/validator/src/consensus/hashing.rs
rmeissner: hmm I would prefer to use the "official" erc712 implementation if possible. The least I would like is that we use it in the tests to ensure our hashing functions return the correct erc712 hash

PR #727, opened 2026-08-11
[Phase 8]: Remove npm workspaces

rmeissner:
The npm workspace boilerplate is removed, as only the explorer and example use typescript now.

As a coordination layer Just was introduced. All exisiting commands have been migrated and documentation was updated.

Biome was added to the examples npm project, as the project wide biome was removed.

### Testing

- All tests (including integration tests) need to pass successfully
- Explorer preview should be deployed
- All `npm` commands should be replaced by `just` commands

---

## CLA signature


.github/workflows/coverage.yml
```diff
@@ -29,10 +32,14 @@ jobs:
           # See <https://github.com/foundry-rs/foundry/issues/13362>
           version: v1.5.1
 
+      - name: Install just
```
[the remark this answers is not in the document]
rmeissner: this runner explicitly uses ubuntu

Justfile
```diff
@@ -0,0 +1,109 @@
...
+set shell := ["bash", "-euo", "pipefail", "-c"]
+
+# List available recipes.
+default:
+    @just --list
+
+# Build every buildable package (contracts, explorer).
+build:
+    (cd contracts && forge build --force)
+    npm --prefix explorer run build
```
nlordell, line 16: why not `cargo build`?

nlordell:
```suggestion
    npm --prefix explorer run build
    cargo build
```

Justfile
```diff
@@ -0,0 +1,109 @@
...
+
+set shell := ["bash", "-euo", "pipefail", "-c"]
+
+# List available recipes.
+default:
+    @just --list
+
+# Build every buildable package (contracts, explorer).
+build:
+    (cd contracts && forge build --force)
```
nlordell, line 15: nit: Why not use `--root` like we do in the scripts?

nlordell (approved):

PR #728, opened 2026-08-11
feat(guard): carry and verify oracleData in the attestation trailer

remedcu:
Carry and verify the `oracleData` hash in the guard's attestation trailer.

### Context and Motivation

- With `oracleData` now bound into the consensus message (#726), the guard must rebuild that message to verify the FROST signature. The trailer gains a `bytes32 oracleDataHash` field so `checkTransaction` can reconstruct the exact signed message.

### Decisions and Tradeoffs

- Only the `keccak256` of `oracleData` is carried (a fixed 256-byte payload), placed next to `oracle`; the guard does not introspect the data and the raw bytes stay with the oracle contract. The message is rebuilt via `ConsensusMessages.transactionProposal`, which takes the hash directly.

### Testing

- `forge test` (253, incl. non-empty `oracleData` success + tampered-hash revert + fixed-size codec round-trip), `forge fmt --check`, `forge lint`.

### DIFF
--- a/contracts/src/guard/SafenetGuard.sol
+++ b/contracts/src/guard/SafenetGuard.sol
@@ -198,8 +198,13 @@ contract SafenetGuard is ISafenetGuard, BaseTransactionGuard {
         if (_isAutoAllowed(to, value, data, operation)) return;
 
         if (AttestationTrailer.hasTrailer(signatures)) {
-            (uint64 epoch, address oracle, Secp256k1.Point memory groupKey, FROST.Signature memory signature) =
-                AttestationTrailer.decode(signatures);
+            (
+                uint64 epoch,
+                address oracle,
+                bytes32 oracleDataHash,
+                Secp256k1.Point memory groupKey,
+                FROST.Signature memory signature
+            ) = AttestationTrailer.decode(signatures);
             // Cheap forest-membership check first (also implies a non-zero key, enforced on record), so an
             // untrusted key short-circuits before the more expensive Safe tx hash is computed.
             require($epochs.isKnown(groupKey, epoch), UntrustedAttestationKey());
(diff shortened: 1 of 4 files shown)


contracts/src/guard/SafenetGuard.sol
```diff
@@ -198,8 +198,13 @@ contract SafenetGuard is ISafenetGuard, BaseTransactionGuard {
...
 
         if (AttestationTrailer.hasTrailer(signatures)) {
-            (uint64 epoch, address oracle, Secp256k1.Point memory groupKey, FROST.Signature memory signature) =
-                AttestationTrailer.decode(signatures);
+            (
+                uint64 epoch,
+                address oracle,
+                Secp256k1.Point memory groupKey,
+                FROST.Signature memory signature,
+                bytes memory oracleData
```
rmeissner: nit: the oracle feels off for me, would expect the oracleData to be next to the oracle address.

rmeissner (dismissed):

contracts/src/libraries/AttestationTrailer.sol
```diff
@@ -58,22 +61,30 @@ library AttestationTrailer {
...
+            bytes memory oracleData
+        )
     {
         bytes calldata payloadData = SignatureExtension.payload(signatures, TYPE_HASH);
-        require(payloadData.length == _PAYLOAD_LENGTH, MalformedAttestationTrailer());
-        (epoch, oracle, groupKey, signature) =
-            abi.decode(payloadData, (uint64, address, Secp256k1.Point, FROST.Signature));
+        require(payloadData.length >= _MIN_PAYLOAD_LENGTH, MalformedAttestationTrailer());
+        (epoch, oracle, groupKey, signature, oracleData) =
+            abi.decode(payloadData, (uint64, address, Secp256k1.Point, FROST.Signature, bytes));
```
rmeissner: discussion/ request change: it might make sense to only pass the oracleData hash in the attestation trailer. This would make it a constant bytes32 and therefore reduce the size. We are currently not introspecting the data, so it is not required and could be adjusted in the future when we need it.

This would require to add a method to calculate the eip712 hash with the oracleData hash (so skip the one hashing) instead of using the full data (which is still required in the oracle contract).

contracts/src/libraries/AttestationTrailer.sol
```diff
@@ -9,10 +9,12 @@ import {SignatureExtension} from "@/libraries/SignatureExtension.sol";
  * @title AttestationTrailer
  * @notice Recognises and decodes a Safenet attestation carried as a {SignatureExtension} on Safe's
  *         `signatures` bytes.
- * @dev The attestation is a typed signature extension: its payload is the fixed 224-byte
- *      `abi.encode(uint64 epoch, address oracle, Secp256k1.Point groupKey, FROST.Signature signature)`,
+ * @dev The attestation is a typed signature extension: its payload is the fixed 256-byte
+ *      `abi.encode(uint64 epoch, address oracle, bytes32 oracleDataHash, Secp256k1.Point groupKey, FROST.Signature signature)`,
```
nlordell, line 13: nit: maximum column length not respected here.

nlordell (approved):

PR #736, opened 2026-08-11
[Phase 9] Docs Adjustment

rmeissner:
After the npm workspace cleanup this adjust and extends the docs.

Example configurations for the validator and sentinel have been added. The validator has been adjusted to include the apply to the new validator implementation. Furthermore a first version of the sentinel handbook has been added. It is currently incomplete and should be extended over time.

The configuration has not been migrated, as it describes the Beta state which is not compatible with the current implementation. Once the new configuration is set it should be extended.

---

## CLA signature

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,7 @@ lcov.info
 node_modules/
 target/
 validator*.toml
+!/crates/validator/validator.sample.toml
 sentinel*.toml
+!/crates/sentinel/sentinel.sample.toml
 *_logs.txt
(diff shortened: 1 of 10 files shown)


nlordell (approved):

.gitignore
```diff
@@ -14,5 +14,6 @@ node_modules/
 sentinel_a_logs.txt
 sentinel_b_logs.txt
 target/
-validator*.toml
+/validator*.toml
+/sentinel*.toml
```
nlordell: nit: I would prefer:

```suggestion
validator*.toml
!/crates/validator/validator.sample.toml
sentinel*.toml
!/crates/sentinel/sentinel.sample.toml
```

As I keep test configurations outside of the root as well.

PR #742, opened 2026-08-12
[Sentinel Engine 2d]: Add OpenAPI Specification

nlordell:
This PR adds an OpenAPI specification for the sentinel engine HTTP API.

### Decisions and Tradeoffs

We added another `just` recipe for validating the OpenAPI spec is valid. I decided to:

- Use docker to run the `redocly` CLI to avoid any additional NPM dependencies
- Only check against the specification is valid instead of general linting, since many of the rules do not apply to us (we are using the spec to define what should be implemented by a sentinel engine and not as a general public API specification).

Additionally, I updated the Cargo crate versions to 0.2.0 to match our "this is Safenet Q3" versioning everywhere.

### Testing

OpenAPI specification is now checked in CI.

---

## CLA signature

### DIFF
--- a/crates/sentinel-engine/Cargo.toml
+++ b/crates/sentinel-engine/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sentinel-engine"
-version = "0.1.0"
+version = "0.2.0"
 edition = "2024"
 publish = false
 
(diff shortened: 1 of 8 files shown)


crates/sentinel-engine/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "sentinel-engine"
-version = "0.1.0"
+version = "0.2.0"
```
rmeissner, line 3: uber nit: why increase this version? We never published it before, right?

rmeissner (approved):

PR #743, opened 2026-08-12
Use SafeId and oracleas indeces on proposal event

### LINKED ISSUE #669: Adjust Proposal Events

rmeissner:
Fixes #669 

To allow indexing for the oracle address on the proposal event the chain id and Safe address are combined into a Safe id. The assumption is made that currently common chainIds do not exceed uint96.

The new indexed oracle is used to limit the events loaded in the explorer.

### Testing

- Integration tests should pass

---

## CLA signature

### DIFF
--- a/contracts/src/interfaces/IConsensus.sol
+++ b/contracts/src/interfaces/IConsensus.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.8.30;
 import {FROST} from "@/libraries/FROST.sol";
 import {FROSTGroupId} from "@/libraries/FROSTGroupId.sol";
 import {FROSTSignatureId} from "@/libraries/FROSTSignatureId.sol";
+import {SafeId} from "@/libraries/SafeId.sol";
 import {SafeTransaction} from "@/libraries/SafeTransaction.sol";
 import {Secp256k1} from "@/libraries/Secp256k1.sol";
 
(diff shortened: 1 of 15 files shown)


contracts/src/interfaces/IConsensus.sol
```diff
@@ -67,37 +68,37 @@ interface IConsensus {
...
     event TransactionAttested(
         bytes32 indexed safeTxHash,
-        uint256 indexed chainId,
-        address indexed safe,
+        SafeId.T indexed safeId,
+        address indexed oracle,
         uint64 epoch,
-        address oracle,
+        uint256 chainId,
+        address safe,
```
nlordell: request change: This feels like a lot of redundant data that I don't think should be there. If we do want to include it, then I think it should be justified in the PR description (under decisions and tradeoffs), but right now it doesn't make sense to me why we are paying extra log data gas for information to be included twice.

nlordell (changes requested):

contracts/src/interfaces/IConsensus.sol
rmeissner: will remove this, I don't think we rely on this

crates/sentinel/src/service.rs
```diff
@@ -835,10 +836,9 @@ mod tests {
         SentinelEvents::Consensus(Consensus::ConsensusEvents::TransactionProposed(
             Consensus::TransactionProposed {
                 safeTxHash: safe_tx_hash,
-                chainId: U256::from(CHAIN_ID),
-                safe: SAFE,
-                epoch: 7,
+                safeId: keccak256((U256::from(CHAIN_ID), SAFE).abi_encode()),
```
nlordell: request change: this is wrong right?

scripts/run_validator_integration_test.sh
```diff
@@ -232,7 +232,7 @@ while [ "$SECONDS" -lt "$DEADLINE" ]; do
             --from-block 0 \
             --to-block latest \
             --address "$CONSENSUS_ADDR" \
-            'TransactionAttested(bytes32,uint256,address,uint64,address,bytes32,((uint256,uint256),uint256))')
+            'TransactionAttested(bytes32,bytes32,address,uint64,uint256,address,bytes32,((uint256,uint256),uint256))')
```
nlordell: request change:

```suggestion
            'TransactionAttested(bytes32,bytes32,address,uint64,bytes32,((uint256,uint256),uint256))')
```

nlordell (approved):

PR #748, opened 2026-08-12
[Phase 3] Allow to configure the fee

rmeissner:
Uses the same delayed update pattern for the fee as for other configuration values.

Currently each config type will have its own delayed type library. This can be reevaluated once all config parameters are set and implemented.

---

## CLA signature

### DIFF
--- a/contracts/src/SentinelOracle.sol
+++ b/contracts/src/SentinelOracle.sol
@@ -6,6 +6,7 @@ import {SafeERC20} from "@oz/token/ERC20/utils/SafeERC20.sol";
 import {IOracle} from "@/interfaces/IOracle.sol";
 import {BondConfig} from "@/libraries/BondConfig.sol";
 import {DelayedAddress} from "@/libraries/DelayedAddress.sol";
+import {DelayedUint256} from "@/libraries/DelayedUint256.sol";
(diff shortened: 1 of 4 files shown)


contracts/src/SentinelOracle.sol
```diff
@@ -301,6 +314,18 @@ contract SentinelOracle is IOracle {
         return $protocolFundsReceiverConfig.pendingActiveAt;
     }
 
+    function fee() external view returns (uint256) {
+        return $feeConfig.current();
+    }
+
+    function pendingFee() external view returns (uint256) {
+        return $feeConfig.pendingValue;
```
nlordell, line 322: question: I think these are slightly wrong - in case `pendingValue` is already current, this will not return 0. Is this intended? I would expect `pendingValue` to be applied (like with `current`) before returning.

For `pendingFeeActiveAt` it can be a bit more jarring where you get a block that is in the past.

rmeissner: I will handle this in a future PR. Should be part of the library (i.e. a `pending()` function that returns a block number and value). Added this as an explicit phase

nlordell (approved):

PR #753, opened 2026-08-13
[Nonces 4a]: Refactor Nonce Generator Synchronization

nlordell:
### Context and Motivation

This PR refactors how nonce generator synchronization works in a couple meaningful ways that are required for correctness between the state machine and the effect handler:

- We no longer implement synchronization on the `NonceGenerator` itself and instead wrap it in a `Mutex` at the effect handler level. This is to ensure that we have strict ordering of `NonceGenerator` operations at the effect handler level (otherwise, effects arriving simultaneously for stopping and starting the nonce generation stream - in case we are catching up after a reorg for example - can race and cause the last effect to arrive to be processed in a way where its effect is discarded by a previous one)
- We now cancel ongoing nonce chunk futures correctly (which was not the case before - see the added test cases).

### Decisions and Tradeoffs

- There is a narrow case where `NonceStream` is closed _as_ the `NonceChunk` is being delivered over the `oneshot` channel which will resolve the `next` future even if nonce stream was stopped. This is acceptable in our case (since additional nonce chunk generation just means some additional CPU work, and it should almost never happen practice).
- We always completely drain the `requests` queue - see the inline comment that explains why.

---

## CLA signature

### DIFF
--- a/crates/validator/src/secrets/nonces.rs
+++ b/crates/validator/src/secrets/nonces.rs
@@ -2,10 +2,10 @@
 
 use crate::frost::{keygen::KeyShare, preprocess::NonceChunk};
 use alloy::primitives::B256;
-use rand::{CryptoRng, RngCore};
+use rand::rngs::ThreadRng;
 use std::{
     collections::{BTreeMap, btree_map},
-    sync::{Arc, Mutex, PoisonError, mpsc},
+    sync::{Arc, mpsc},
     thread,
     time::Instant,
 };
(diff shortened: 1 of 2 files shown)


crates/validator/src/secrets/nonces.rs
```diff
@@ -52,48 +51,39 @@ impl NonceGenerator {
...
-    size: Option<u64>,
+enum Sampler {
+    /// Generate a full chunk of nonces using the specified key share.
+    Full(Arc<KeyShare>),
+    /// Custom function for generating a nonces chunk.
+    ///
+    /// Used only for unit testing.
+    #[allow(clippy::type_complexity)]
+    #[cfg_attr(not(test), expect(dead_code))]
+    Custom(Box<dyn Fn(&mut ThreadRng) -> Result<NonceChunk, rand::Error> + Send + Sync + 'static>),
```
rmeissner, line 79: nit: might make sense to only make this available in tests

crates/validator/src/secrets/nonces.rs
```diff
@@ -138,17 +128,49 @@ impl NonceStream {
...
-                    Ok(()) => continue 'outer,
-                    Err(value) => {
-                        tracing::debug!("nonces future was canceled; waiting for next request");
-                        nonces = value;
+                match requests.try_recv() {
+                    Ok(new_request) => {
+                        tracing::debug!("nonces future was canceled; using next request");
+                        request = new_request;
                     }
+                    Err(mpsc::TryRecvError::Empty) => break,
```
rmeissner: nit: would be nice to have a comment here, as it is confusing that we continue to the next part of the logic (the send) on an error (which indicates that we are done draining).

rmeissner (approved):

crates/validator/src/secrets/nonces.rs
nlordell: fixed

crates/validator/src/secrets/nonces.rs
nlordell: fixed

PR #754, opened 2026-08-13
[Nonces 4b]: Retain Groups in Secret Storage

nlordell:
### Context and Motivation

Previously, we would require _carefully_ located pruning effects within the state machine to ensure that the secret store gets correctly garbage collected, and that no dangling values are kept over time (i.e. no storage leaks).

This was a bit brittle (especially in the face of reorgs, and asynchronous effects, where reasoning about the order in which effects happen is impossibly hard). In order to address this, we want to define an idempotent "reconcile groups" effect that will keep only the relevant groups around.

This PR adds the required `SecretStore` database methods for retaining group secrets.

### Decisions and Tradeoffs

We split retaining keygen secrets and nonces into separate functions, allowing us to keep a different subset of secrets or different groups (i.e. we can cleanup the keygen secrets without removing nonces and vice-versa).

---

## CLA signature

### DIFF
--- a/crates/validator/src/secrets/store.rs
+++ b/crates/validator/src/secrets/store.rs
@@ -31,7 +31,7 @@ use alloy::{
     hex::ToHexExt,
     primitives::{Address, B256},
 };
-use sqlx::sqlite::SqlitePool;
+use sqlx::{QueryBuilder, Sqlite, sqlite::SqlitePool};
 use std::num::TryFromIntError;
 
 /// Error produced by the [`SecretStore`].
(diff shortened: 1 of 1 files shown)


crates/validator/src/secrets/store.rs
```diff
@@ -422,6 +439,44 @@ impl SecretStore {
...
+
+    /// Deletes every row in `table` whose `group_id` is not one of `groups`.
+    async fn retain_groups(
+        &self,
+        table: &'static str,
+        groups: impl IntoIterator<Item = B256>,
+    ) -> Result<(), Error> {
+        let mut groups = groups.into_iter().peekable();
+        let mut query = if groups.peek().is_none() {
+            QueryBuilder::<Sqlite>::new(format!("DELETE FROM {table}"))
```
rmeissner, line 472: nit: maybe worth making it explicit in teh function comment that it is ok to delete all rows if no group is defined

rmeissner (approved):

crates/validator/src/secrets/store.rs
nlordell: Added

PR #756, opened 2026-08-13
[Nonces 4c]: Reconcile Groups with State Machine

nlordell:
### Context and Motivation

A followup to #754, this PR adds a new effect to reconcile groups between the state machine and the effect handler.

### Decisions and Tradeoffs

- We needed to work around a borrow checker implementation where passing in an iterator without collecting it would cause the future to require different lifetime bounds that it could not guarantee (it would cause the reference to `groups` to be held across await boundaries). I don't believe it is an actual issue (rather a borrow checker limitation), but we work around it by collecting the groups needed for DKG into a `Vec`.
- We do not use the new effect yet, this will happen in the following PR.

---

## CLA signature

### DIFF
--- a/crates/validator/src/service/effect.rs
+++ b/crates/validator/src/service/effect.rs
@@ -12,6 +12,7 @@ use crate::{
 use alloy::primitives::{Address, B256};
 use safenet_core::effects::EffectHandler;
 use std::{
+    collections::BTreeMap,
     error::Error,
     fmt::{self, Display, Formatter},
     sync::Arc,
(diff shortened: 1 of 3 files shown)


crates/validator/src/service/effect.rs
```diff
@@ -303,6 +315,36 @@ impl Handler {
...
+                            .filter_map(|(group_id, key_share)| {
+                                key_share.is_none().then_some(*group_id)
+                            })
+                            // Collecting here is unfortunately required for the
+                            // compiler to correctly resolve lifetimes of our
+                            // future; in theory it should not be needed.
+                            .collect::<Vec<_>>(),
+                    )
+                    .await?;
+                self.secrets.retain_nonces(groups.keys().copied()).await?;
```
rmeissner: request change: this retains all groups, while the nonce generator only retains when we still have a key share.

rmeissner (changes requested):

crates/validator/src/service/effect.rs
nlordell: Nice catch! Fixed.

rmeissner (approved):

PR #761, opened 2026-08-13
[Phase 9c] Pack governance-config and Request storage

rmeissner:
To optimize storage access all parameters and variables will use only the necessary width.

Also to simplify the code the applied event are removed. Event for scheduling configuration changes are still emitted. The changes will be autoapplied as soon as they are matured.

## Design Decisions
- Block numbers as points in time use `uint64` as a unit
- Block numbers as a range (i.e. timeouts) use `uint32` as a unit

---

## CLA signature


contracts/script/DeploySentinelOracle.s.sol
```diff
@@ -32,6 +32,10 @@ contract DeploySentinelOracleScript is Script {
 
         vm.startBroadcast();
 
+        // Casts are safe: env vars are deployer-controlled and reviewed before broadcasting, and
+        // the constructor's own require()s (ZeroFee/InvalidFeeShare/...) still validate the
+        // decoded values' business-logic bounds regardless of width.
+        // forge-lint: disable-start(unsafe-typecast)
```
nlordell: request change: Honestly, I would just use [`SafeCast`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/SafeCast.sol) (we already use OpenZeppelin libraries) so that we get reverts when values are not correct. I don't think silent failures here are a good idea. For example:

```solidity
uint96 requestFee = vm.envUint("SENTINEL_REQUEST_FEE").toUint96();
```

In the end, it will be about the same amount of code change lines, and you won't need a comment to justify your bad behaviour.

contracts/src/libraries/BondConfig.sol
```diff
@@ -68,20 +76,31 @@ library BondConfig {
...
+    // Returns the scheduled pending (bondMultiplier, slashingMultiplier) and the block they
+    // activate at, or `(0, 0, 0)` if nothing is currently scheduled -- regardless of whether a
+    // scheduled change is already due.
+    function pending(T storage self)
+        internal
+        view
+        returns (uint32 bondMultiplier, uint32 slashingMultiplier, uint256 activeAt)
+    {
+        if (self.pendingActiveAt == 0) return (0, 0, 0);
+        return (self.pendingBondMultiplier, self.pendingSlashingMultiplier, self.pendingActiveAt);
```
nlordell: question: Wait - is this actually a single storage read? Don't you need the following to guarantee it:


```suggestion
        T memory config = self;
        if (config.pendingActiveAt == 0) return (0, 0, 0);
        return (config.pendingBondMultiplier, config.pendingSlashingMultiplier, config.pendingActiveAt);
```

rmeissner: do we want to optimize for this? then I would have to recheck the whole code for this.

contracts/src/libraries/DelayedAddress.sol
```diff
@@ -27,25 +32,24 @@ library DelayedAddress {
...
         applyPending(self);
 
-        uint256 activeAt = block.number + governanceDelay;
+        activeAt = block.number + governanceDelay;
+        require(activeAt <= type(uint64).max, ActiveAtOutOfRange());
         self.pendingValue = newValue;
-        self.pendingActiveAt = activeAt;
-        emit ValueScheduled(newValue, activeAt);
+        // forge-lint: disable-next-line(unsafe-typecast)
+        self.pendingActiveAt = uint64(activeAt);
```
nlordell: question: Same here, are you sure that this actually compiles to a single storage read/write? I am not sure (in fact, we use `memory` explicitly in the `FROSTCoordinator` contract because I had, at the time, determined that the compiler was not optimizing storage access).

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -89,16 +102,24 @@ library SentinelOracleRequest {
         if (approveMet || denyMet) {
             if (approveMet && denyMet) {
                 newState = State.FROZEN;
-                self.arbitrationDeadline = block.number + arbitrationTimeout;
+                uint256 arbitrationDeadline = block.number + arbitrationTimeout;
+                require(arbitrationDeadline <= type(uint48).max, DeadlineOutOfRange());
```
nlordell: question: Why not use the OpenZeppelin `SafeCast` library?

contracts/src/libraries/DelayedAddress.sol
rmeissner: yes, that is probably I did focus on storage layout optimizations only. Do you expect this to be handled in this PR?

crates/sentinel/src/service.rs
```diff
@@ -273,7 +273,7 @@ impl SentinelTransition {
     ) -> (State, Commands<State, Self>) {
         let request_id = event.requestId;
         let request = Request {
-            bond_target: event.bondTarget,
+            bond_target: U256::from(event.bondTarget),
```
nlordell: question: Why not change `Request::bond_target: u128`?

crates/sentinel/src/service.rs
```diff
@@ -273,7 +273,7 @@ impl SentinelTransition {
...
             commit_deadline: event.commitDeadline.saturating_to::<u64>(),
```
nlordell: question: Request can also use `Uint<48>` to have more correct typing right?

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -20,20 +20,32 @@ library SentinelOracleRequest {
...
-        uint256 slashAmount;
-        uint256 commitDeadline;
-        uint256 revealDeadline;
-        uint256 arbitrationDeadline;
         State state;
-        uint256 committedCount;
-        uint256 revealedCount;
-        uint256 approveSentinelCount;
-        uint256 denySentinelCount;
+        uint48 commitDeadline;
```
nlordell: request change: we use `uint64` for deadlines in some places and `uint48` for deadlines in others - I would request that we stay consistent with how we represent deadlines in the codebase (I would err on the side of `uint64`, since that is how it is represented in clients).

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -20,20 +20,32 @@ library SentinelOracleRequest {
...
+        uint16 committedCount;
+        uint16 revealedCount;
+        // slot 2
+        uint48 revealDeadline;
+        uint48 arbitrationDeadline;
+        uint16 approveSentinelCount;
+        uint16 denySentinelCount;
+        uint24 daoFeeShare;
```
nlordell, line 38: ~~request change: doesn't this need to be a `uint32` to match the fee multiplier we use in other places?~~

rmeissner: no, as it is relative to the denominator which is 100000

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -89,16 +102,24 @@ library SentinelOracleRequest {
...
             // An established side exists, so a non-revealer's silence can only be griefing (stalling
             // a request whose outcome their own commit already contributed to) -- slash the governed
-            // slash amount from their bond to the protocol funds receiver.
-            unrevealedBond = (self.committedCount - self.revealedCount) * self.slashAmount;
+            // slash amount from their bond to the protocol funds receiver. Widen the (uint16) count
+            // to uint128 *before* multiplying: `count * self.slashAmount` would otherwise compute in
+            // `uint96` (the wider of the two operand types) and could overflow-revert even though
+            // the true product fits comfortably in `uint128` (`type(uint16).max * type(uint96).max`
+            // is always < `type(uint128).max`, with room to spare).
+            uint128 nonRevealerCount = self.committedCount - self.revealedCount;
```
nlordell: nit: unless we wrap these blocks in `unchecked`, this widening makes no sense to me (we will likely never get close to `type(uint96).max` for the token that we chose, but we are still generating code that does overflow checks).

contracts/src/libraries/SentinelOracleRequests.sol
nlordell: **Edit**: THis is something different and not a fee multiplier.

nit: Change this to a `uint16` and `FEE_SHARE_DENOMINATOR` to 10_000. It is enough precision, and 10_000 are basis points, which are very commonly used in finance.

nlordell: > no, as it is relative to the denominator which is 100000

Realized this also 🙈

nlordell (changes requested):
Requesting changes:

- The scripts should use stricter casting on the env-vars (we already import `SafeCast`, why not use it in the scripts since it is just as ergonomic IMO?)
- I would request we use strict Rust types (I think `alloy` provides types for representing `uint48` for example, and I would not cast to a `u64`).
- I feel like there are some inconsistencies here with the types:
  - block deadlines are 64-bit for [configuration deadlines](https://github.com/safe-research/safenet/blob/711ab4cbc278d9967253caf0e2ef834641b7abfb/contracts/src/libraries/DelayedAddress.sol#L17)
  - block deadlines are 48-bit for [oracle process](https://github.com/safe-research/safenet/blob/711ab4cbc278d9967253caf0e2ef834641b7abfb/contracts/src/libraries/SentinelOracleRequests.sol#L40)
- We use `SafeCast` in some places, and manual overflow checks in others, I would just use `SafeCast` (its very ergonomic).

I had some other nits, but ☝️ were the main ones that I would like to see changed.

contracts/src/libraries/DelayedAddress.sol
nlordell: It can be in a followup, but just wanted to clarify that the storage optimizations are not actually optimizing storage access (just size).

contracts/src/libraries/BondConfig.sol
nlordell: see below.

contracts/src/libraries/SentinelOracleRequests.sol
rmeissner: we explicitly decided on 100_000 and uint24, therefore I would stick to this for now.

contracts/src/libraries/BondConfig.sol
```diff
@@ -42,46 +48,71 @@ library BondConfig {
...
+        // Discards the return value: already persisted by `applyPending` itself if a change was
+        // due, so re-writing `bondMultiplier`/`slashingMultiplier` here again would just be a
+        // redundant extra write to the same slot.
         applyPending(self);
 
-        uint256 activeAt = block.number + governanceDelay;
+        activeAt = (block.number + governanceDelay).toUint64();
         self.pendingBondMultiplier = newBondMultiplier;
         self.pendingSlashingMultiplier = newSlashingMultiplier;
         self.pendingActiveAt = activeAt;
```
nlordell, line 67: nit: The optimization fixes are not applied consistently. For example, here we are potentially writing to storage 3 separate times (I don't know for sure, I don't know what guarantees the Solidity compiler provides in these cases).

contracts/src/libraries/SentinelOracleRequests.sol
```diff
@@ -20,20 +23,30 @@ library SentinelOracleRequest {
...
     // ============================================================
 
+    // Fields are ordered and sized to pack into 3 storage slots instead of 12. `fee`,
+    // `bondTarget`, and `slashAmount` are ERC20 amounts sized `uint96` (the width Uniswap uses
+    // for token amounts -- vastly more than any realistic fee/bond needs). Window/deadline block
+    // numbers are `uint64` -- matching the offchain sentinel's native block-number type exactly,
+    // rather than a narrower onchain-only width (slot 3 is left with 32 idle bits as a result, but
+    // still fits the same 3 slots). Sentinel counts fit in `uint16` (we do not expect anywhere near
+    // that many sentinels); `daoFeeShare` fits in `uint24` (`FEE_SHARE_DENOMINATOR` is 100_000).
     struct Request {
```
nlordell, line 33: comment: These massive structs are a bit unweidly, and make it hard to optimize "I only need slot 1 and 3". **Not for this PR**, but we should consider restructuring this and potentially reordering fields to improve access.

nlordell (approved):
LGTM, it is hard to review all of the uint size changes though, so I may have missed something.

PR #767, opened 2026-08-14
[Nonces 5]: Nonce Topup in State Machine

nlordell:
### Context and Motivation

This PR moves the decision to topup nonces to the state machine. This is done as part of a larger change to track nonces in the state machine (since nonce chunk linking must be reorg aware). Note that nonce linking still happens at the storage level, in order to keep this PR shorter and more reviewable.

### Decisions and Tradeoffs

Instead of running every signing request, we decide if a topup is needed once per block.

---

## CLA signature

### DIFF
--- a/crates/validator/src/state/preprocess.rs
+++ b/crates/validator/src/state/preprocess.rs
@@ -1,12 +1,19 @@
-use super::{KeyGenConfirmation, KeyGenParticipation, RolloverState, State, Transition};
+use super::{
+    KeyGenConfirmation, KeyGenParticipation, NonceState, RolloverState, State, Transition,
+};
 use crate::{
     bindings::Coordinator,
     consensus::epoch::EpochId,
+    frost::preprocess::{self, SEQUENCE_CHUNK_SIZE},
     service::{Action, Effect},
 };
 use alloy::primitives::B256;
 use safenet_core::state::{Command, Commands};
 use std::collections::BTreeMap;
+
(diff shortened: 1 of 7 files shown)


crates/validator/src/state/preprocess.rs
```diff
@@ -130,3 +177,54 @@ impl Transition {
...
+
+    /// Records an onchain assignment.
+    pub(super) fn link(&mut self, chunk: u64, root: B256) {
+        self.chunks.insert(chunk, Some(root));
+    }
+
+    /// Returns the chunk the onchain contract is expected to assign to the
+    /// next commitment.
+    ///
+    /// Returns `None` in case we've reached the very last chunk.
```
rmeissner, line 207: question: what is the very last chunk?

crates/validator/src/state/preprocess.rs
```diff
@@ -130,3 +177,54 @@ impl Transition {
         )
     }
 }
+
+impl NonceState {
+    /// Advances the observed sequence and forgets past roots for nonce chunks
+    /// that can no longer used for signing.
+    pub(super) fn observe(&mut self, sequence: u64) {
```
rmeissner, line 184: uber nit: `observer` does not sound state changing and was surprised to read that it "forgets" something for good :D

rmeissner: after reading the code maybe something like "advance" would be a better name

rmeissner (approved):

PR #768, opened 2026-08-14
refactor(validator): single oracleData-hash proposal hasher

remedcu:
Single validator proposal hasher taking the oracleData hash, addressing the #726 re-review.

### Context and Motivation

- Follow-up to the merged oracleData-binding stack. On #726, nlordell asked for one hashing implementation (`crates/validator/src/consensus/hashing.rs`) and rmeissner asked that our hashing be verified against the official EIP-712 implementation.

### Decisions and Tradeoffs

- Collapsed the two proposal hashers into one `transaction_proposal_hash(…, oracle_data_hash: B256, …)`; callers `keccak256` the full data at the callsite (mirroring the on-chain `ConsensusMessages.transactionProposal`). The `Transaction` packet now carries the hash rather than the full bytes.
- The type hash comes from the generated `TransactionProposal` type instead of a hardcoded constant, and a test asserts the hand-rolled encoding equals alloy's `SolStruct::eip712_signing_hash` over the full-data message (the attest path can't call it directly, since it only has the hash).
- Trimmed the `too_many_arguments` allow comment in `bindings.rs`.

### Testing

- `cargo fmt --check`, `cargo clippy -p validator --all-targets`, `cargo test -p validator` (40).
- `scripts/run_validator_integration_test.sh`: SUCCESS end-to-end.


crates/validator/src/consensus/hashing.rs
```diff
@@ -110,44 +102,27 @@ impl ConsensusDomain {
...
+    /// already-computed [`safe_tx_hash`] and the `keccak256` of `oracleData`.
     ///
-    /// This is the proposal/signing path, which always holds the full `oracleData`, so it uses the
-    /// canonical `SolStruct` EIP-712 hashing (alloy hashes the `bytes` member itself).
+    /// Mirrors `ConsensusMessages.transactionProposal` onchain: it takes the `oracleData` already
+    /// reduced to its `keccak256` (EIP-712 encodes the message's `bytes oracleData` member as its hash),
+    /// so a caller holding the full data `keccak256`s it at the callsite while the attest path passes the
+    /// hash straight from the `TransactionAttested` event (which only carries the hash, keeping the
+    /// validator callback a constant size). The encoding is spelled out by hand rather than via
+    /// `SolStruct::eip712_signing_hash`, whose signing hash requires a value holding the full bytes; the
```
nlordell: nit: I would use the term `oracle data preimage` here to be more precise.

crates/validator/src/state/mod.rs
```diff
@@ -236,9 +236,9 @@ enum Packet {
         epoch: EpochId,
         /// The oracle vouching for the transaction.
         oracle: Address,
-        /// Arbitrary oracle-specific data, bound into the signed message. Retained as bytes and hashed
-        /// only when building the (constant-size) attestation call.
-        oracle_data: Bytes,
+        /// The `keccak256` of the oracle-specific data bound into the signed message. Only the hash is
+        /// retained (it is all the proposal digest and the constant-size attestation call need).
+        oracle_data_hash: B256,
```
nlordell: request change: The packet itself uses `oracle_data: Bytes` as it contains the preimage (see `safe_tx` which also contains the preimage). It should be mirrored here.

crates/validator/src/consensus/hashing.rs
```diff
@@ -168,10 +143,10 @@ impl ConsensusDomain {
         &self,
         epoch: EpochId,
         oracle: Address,
-        oracle_data: Bytes,
+        oracle_data_hash: B256,
```
nlordell: nit: THis is for the packet, where you _do_ have the preimage:

```suggestion
        oracle_data: Bytes,
```

nlordell:
I don't know if the integration test is failing because of flakiness or legitimate. It seems to fail every time for you (while it would pass pretty often for previous PRs). You can try an rebase your changes on top of the nonces branch, or temporarily raise the block time to 2 in order to check whether or not the failure is legitimate.

